PR #9, opened 2021-03-02
Reconfigure scope hoisting, add core folder, reset files

wibjorn:
minireset.css

task-378377

- Created the /css/src/core folder.
- I had previously misconfigured the scope hoisting in the monorepo's package.json. Packages were still being hoisted. Fixed this here. This was necessary because ...
- I pulled in normalize and minireset via npm. These needed to be in the /css folder for importing.


package.json
```diff
@@ -13,16 +13,24 @@
 		"pretty-quick": "pretty-quick --staged --pattern \"**/*.{scss,ts,js,json,yml}\""
 	},
 	"nohoist": [
-		"./@microsoft/atlas-css"
+		"**/@microsoft/atlas-css/"
 	],
```
jdanyow, line 17: what's the nohoist for?

wibjorn: added nohoist to ensure that during dev those files would exist in the same place they would after npm installing atlas-css

jdanyow (approved):

css/src/tokens/palette.scss
```diff
@@ -8,7 +8,7 @@ Notes:
 
 // Monochrome
 
-$palette-grey-10: hsla(0, 0%, 98, 1) !default; // #F2F2F2
+$palette-grey-10: hsla(0, 0%, 98%, 1) !default; // #F2F2F2
 $palette-grey-20: hsla(0, 0%, 95%, 1) !default; // #F2F2F2
```
squalrus, line 12: Do these two values _**really**_ result in the same HEX?

PR #12, opened 2021-03-03
Adding shadow tokens

ollips:
task-377246

Creating shadow.scss, within which we store all shadow values.

### DIFF
--- a/.changeset/warm-jars-battle.md
+++ b/.changeset/warm-jars-battle.md
@@ -2,4 +2,4 @@
 '@microsoft/atlas-css': minor
 ---
 
-Adding shadow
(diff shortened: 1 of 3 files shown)


.changeset/warm-jars-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-css': minor
+---
+
+Adding shadow
```
wibjorn, line 5: Adding tokens for box shadow.

ollips: Updated

wibjorn (approved):

PR #22, opened 2021-03-11
Adding Typography Token

asrivastava0308:
task-377244

Creating Typography token for Atlas

### DIFF
--- a/.changeset/forty-trees-burn.md
+++ b/.changeset/forty-trees-burn.md
@@ -1,0 +1,5 @@
+---
+'@microsoft/atlas-css': patch
+---
+
+Adding Typography Token
--- a/.changeset/forty-trees-burn.md
+++ b/.changeset/forty-trees-burn.md
@@ -1,5 +1,5 @@
 ---
-'@microsoft/atlas-css': patch
+'@microsoft/atlas-css': minor
 ---
 
-Adding Typography Token
+Add typography tokens
--- a/css/src/tokens/typography.scss
+++ b/css/src/tokens/typography.scss
@@ -9,6 +9,3 @@ $font-size-3: 1.75rem !default; // 28px
 $font-size-2: 2.125rem !default; // 34px
 $font-size-1: 2.5rem !default; // 40px
 $font-size-0: 3.375rem !default; // 54px
-
-// Heading Sizes
-$font-heading-sizes: $font-size-1 $font-size-2 $font-size-3 $font-size-4 $font-size-5 $font-size-6 !default;
--- a/css/src/tokens/typography.scss
+++ b/css/src/tokens/typography.scss
@@ -10,12 +10,5 @@ $font-size-2: 2.125rem !default; // 34px
 $font-size-1: 2.5rem !default; // 40px
 $font-size-0: 3.375rem !default; // 54px
 
-// Base
-$font-size-small: $font-size-8 !default;
-$font-size-medium: $font-size-7 !default;
-$font-size-large: $font-size-4 !default;
-$font-size-extra-large: $font-size-1 !default;
-$font-size-extra-extra-large: $font-size-0 !default;
-
 // Heading Sizes
 $font-heading-sizes: $font-size-1 $font-size-2 $font-size-3 $font-size-4 $font-size-5 $font-size-6 !default;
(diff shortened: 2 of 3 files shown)


.changeset/forty-trees-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-css': patch
+---
+
+Adding Typography Token
```
wibjorn, line 5: Add typography tokens.

css/src/tokens/typography.scss
```diff
@@ -0,0 +1,21 @@
...
+$font-size-7: 1rem !default; // 16px
+$font-size-6: 1.125rem !default; // 18px
+$font-size-5: 1.25rem !default; // 20px
+$font-size-4: 1.5rem !default; // 24px
+$font-size-3: 1.75rem !default; // 28px
+$font-size-2: 2.125rem !default; // 34px
+$font-size-1: 2.5rem !default; // 40px
+$font-size-0: 3.375rem !default; // 54px
+
+// Base
```
wibjorn, line 13: Let's use the following variable names:

```scss
$font-size-jumbo
$font-size-large
$font-size-body
$font-size-small
$font-size-extra-small
$font-size-tiny
```

css/src/tokens/typography.scss
```diff
@@ -0,0 +1,21 @@
...
+$font-size-small: $font-size-8 !default;
+$font-size-medium: $font-size-7 !default;
+$font-size-large: $font-size-4 !default;
+$font-size-extra-large: $font-size-1 !default;
+$font-size-extra-extra-large: $font-size-0 !default;
+
+// Heading Sizes
```
wibjorn, line 20: For now, let's not add an array with the sizes

.changeset/forty-trees-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-css': patch
```
v-mingho, line 2: Just wanted to clarify my own understanding. Wouldn't this be a minor update instead of a patch, since it's just an addition rather than a fix? I don't think this needs to be changed though as it'll probably get lumped into another minor when it's published.

asrivastava0308: changed

css/src/tokens/typography.scss
asrivastava0308: removed

css/src/tokens/typography.scss
asrivastava0308: lets add them , once we have definitive sizes for them .

squalrus: On ACOM we used to use words like "small", "extra-small", and "tiny" but it became difficult for people to understand if "extra-small" was smaller than "tiny", or vice-versa, on the name alone.

We did switch to numeric sizing "size-1", "size-2", "size-3", etc. which removed the ambiguity and has worked for us.

Alternatively, if a non-numeric value is desired, following the t-shirt sizing model could help: "small", "xsmall", and "xxsmall".

jdanyow: agreed- numeric scale is as meaningful as t-shirt sizes and familiar (heading elements use numeric scale).

.changeset/forty-trees-burn.md
wibjorn: minor👍

css/src/tokens/typography.scss
jdanyow: ah - @wibjorn reminded me this is the existing scale/names. So we'll need to track some work to make this consistent with whatever the new pattern is coming out of the spacing work. We definitely want to account for @squalrus's comment about ambiguity.

wibjorn: It's worth noting there is actual a fair bit in the air regarding how we'll structure our atomics. Next week we should be able to solidify the team's opinions regarding that. The semantic sizes I've put above (while needed to make docs-ui truly inherit from Atlas) don't need to be included until we've figured that structure.

We did just make in this pattern Docs for font scale. (Aside: I think we could remove `tiny`, so it's not competing with extra-small), and it includes a mixture of two categories of atomic class for font-size: 1) those pointed directly headings and 2) those for more general non-heading use cases (large, body, small, etc). We can revisit this next week though.

css/src/tokens/typography.scss
wibjorn: Not seeing the changes

asrivastava0308: Apologies , it should be visible now !

wibjorn (approved):

PR #23, opened 2021-03-11
consolidate builds

jdanyow:
no need for separate builds *yet* if they only differ by the last step


.github/workflows/main.yml
```diff
@@ -21,11 +25,7 @@ jobs:
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v2
-
-      # Runs a set of commands using the runners shell
-      - name: Go to workspace
```
jdanyow, line 27: ought to already be in the workspace dir

.github/workflows/main.yml
```diff
@@ -47,4 +47,5 @@ jobs:
 
       # If the version has been incremented (with `yarn changeset version`), publish package
       - name: Publish on version increment
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
```
jdanyow, line 50: the ref check for main is to future proof this line if we expand the pr branch matching  on line 11/12/13

wibjorn: I think I'd prefer to keep the build separate

wibjorn: I think I'd prefer to keep the build separate

wibjorn:
Works! https://github.com/microsoft/atlas-design/runs/2117039762?check_suite_focus=true

wibjorn (approved):

PR #27, opened 2021-03-18
Add misc tokens

v-mingho:
task-[378374]

This PR migrates tokens in initial-variables.scss and derived-variables.scss that seems useful to Atlas. 

Initial-variables: https://ceapex.visualstudio.com/Engineering/_git/docs-ui?path=%2Fpackages%2Fstyles%2Fsrc%2Futilities%2F_initial-variables.scss

Derived-variables: https://ceapex.visualstudio.com/Engineering/_git/docs-ui?path=%2Fpackages%2Fstyles%2Fsrc%2Futilities%2F_derived-variables.scss

Tokens not being migrated in this PR:
- uhf related
- font stack
- loop indices
- $footer-desktop-height: 128px;
- $content-header-height: 42px;
- $universal-top-margin: 1.5rem;
- $rfs-factor: 6 !default;

## Testing

1. Review changes.


css/src/tokens/misc.scss
```diff
@@ -0,0 +1,33 @@
...
+
+// Focus related variables
+$focus-width: 0.125rem !default;
+$focus-style: dashed !default;
+$focus-outline-offset: 0.125rem !default;
+
+// Layout
+$column-gap: 0.75rem !default;
+$large-column-gap: 1.5rem !default;
+$reading-max-width: 75ch;
```
ollips, line 33: What do you think about moving `$reading-max-width: 75ch;` to the `typography.scss` file? 🤔🤷‍♀️

ollips (commented):
Have a small comment, the rest looks good to me 👍

css/src/tokens/misc.scss
```diff
@@ -0,0 +1,33 @@
...
+$radius-rounded: 290486px !default;
+
+// Direction
+$user-text-direction: ltr !default;
+$content-text-direction: ltr !default;
+
+$user-left: if($user-text-direction == rtl, right, left);
+$user-right: if($user-text-direction == rtl, left, right);
+
+$content-left: if($content-text-direction == rtl, right, left);
```
wibjorn, line 14: This wasn't in any of the tasks, so apologies about the churn: Atlas isn't going to support content directions. We can the content- related vars. The rest is looking good.

css/src/tokens/misc.scss
wibjorn: Not a bad idea, but I think, since its a container, this is probably the right spot for it

wibjorn: I think it might be good if we separate some of these into smaller files (although we don't _need_ to do this now, I think it'll keep things nice and organized. Thoughts?

layout.scss - $column-gap, $large-, $reading-max-width 
focus.scss
animations.scss
direction.scss
radius.scss

wibjorn:
`changeset` as well?

css/src/tokens/misc.scss
v-mingho: Removed content- tokens.

css/src/tokens/misc.scss
v-mingho: I had the same thought after I made this PR, updated.

css/src/tokens/typography.scss
```diff
@@ -9,3 +9,11 @@ $font-size-3: 1.75rem !default; // 28px
 $font-size-2: 2.125rem !default; // 34px
 $font-size-1: 2.5rem !default; // 40px
 $font-size-0: 3.375rem !default; // 54px
+
+// Font Weight
+$weight-normal: 400;
```
wibjorn, line 14: @v-mingho We can add the variables for all the weight here. (We no longer need to use the $dont-use prefix either)

wibjorn (approved):

PR #31, opened 2021-03-22
Add Basic Typography Classes 

asrivastava0308:
task-[378400]

Adding Typography Classes 

## Testing

Should Include basic Typography classes :

- Heading Classes
- Non-Heading Classes
- weight Classes
- Case Classes
- Letter spacing Classes

### DIFF
--- a/css/src/atomics/typography.scss
+++ b/css/src/atomics/typography.scss
@@ -1,58 +1,57 @@
 // Heading sizes
-.heading-h1 {
+
+.font-size-h1 {
 	font-size: $font-size-1 !important;
 }
 
-.heading-h2 {
+.font-size-h2 {
 	font-size: $font-size-2 !important;
 }
 
-.heading-h3 {
+.font-size-h3 {
 	font-size: $font-size-3 !important;
 }
 
-.heading-h4 {
+.font-size-h4 {
 	font-size: $font-size-4 !important;
 }
 
-.heading-h5 {
+.font-size-h5 {
 	font-size: $font-size-5 !important;
 }
 
-.heading-h6 {
+.font-size-h6 {
 	font-size: $font-size-6 !important;
 }
 
 // Non-heading sizes
+
 .font-size-xxl {
 	font-size: $font-size-0 !important;
 }
 
 .font-size-xl {
-	font-size: $font-size-1 !important;
-}
-
-.font-size-l {
 	font-size: $font-size-2 !important;
 }
 
+.font-size-l {
+	font-size: $font-size-4 !important;
+}
+
 .font-size-m {
-	font-size: $font-size-3 !important;
+	font-size: $font-size-6 !important;
 }
 
 .font-size-s {
-	font-size: $font-size-5 !important;
+	font-size: $font-size-8 !important;
 }
 
 .font-size-xs {
-	font-size: $font-size-7 !important;
-}
-
-.font-size-xxs {
 	font-size: $font-size-9 !important;
 }
 
 // Weight
+
 .font-weight-light {
 	font-weight: $weight-light !important;
 }
@@ -74,39 +73,17 @@
 }
 
 // Case
-.capitalized {
-	text-transform: capitalize !important;
-}
 
-.lowercase {
-	text-transform: lowercase !important;
-}
-
-.uppercase {
-	text-transform: uppercase !important;
-}
-
-.italic {
+.font-style-italic {
 	font-style: italic !important;
 }
 
-.underlined {
+.text-decoration-underline {
 	text-decoration: underline;
 }
 
-.underlined-dotted {
-	border-bottom: 1px dotted;
-}
-
-.undecorated {
-	text-decoration: none;
-}
-
-.normal {
-	font-style: normal;
-}
-
 //Letter Spacing
+
 .letter-spacing-wide {
 	letter-spacing: $letter-spacing-wide;
 }
--- a/css/src/atomics/typography.scss
+++ b/css/src/atomics/typography.scss
@@ -1,5 +1,4 @@
 // Heading sizes
-
 .heading-h1 {
 	font-size: $font-size-1 !important;
 }
@@ -25,7 +24,6 @@
 }
 
 // Non-heading sizes
-
 .font-size-xxl {
 	font-size: $font-size-0 !important;
 }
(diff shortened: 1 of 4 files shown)


css/src/atomics/typography.scss
```diff
@@ -0,0 +1,112 @@
...
+.font-weight-semibold {
+	font-weight: $weight-semibold !important;
+}
+
+.font-weight-bold {
+	font-weight: $weight-bold !important;
+}
+
+// Case
+.capitalized {
```
wibjorn, line 77: text-transform: capitalize -> `text-transform-capitalize`, along with others

css/src/atomics/typography.scss
```diff
@@ -0,0 +1,112 @@
+// Heading sizes
+.heading-h1 {
```
wibjorn, line 2: `font-size-h1`

css/src/atomics/typography.scss
```diff
@@ -0,0 +1,112 @@
...
+.lowercase {
+	text-transform: lowercase !important;
+}
+
+.uppercase {
+	text-transform: uppercase !important;
+}
+
+.italic {
+	font-style: italic !important;
```
wibjorn, line 90: font-style-italic

css/src/atomics/typography.scss
```diff
@@ -0,0 +1,112 @@
...
+}
+
+.underlined {
+	text-decoration: underline;
```
wibjorn, line 94: text-decoration-underline

css/src/atomics/typography.scss
```diff
@@ -0,0 +1,112 @@
...
+}
+
+.underlined-dotted {
```
wibjorn, line 97: lets remove this

css/src/atomics/typography.scss
```diff
@@ -0,0 +1,112 @@
...
+	border-bottom: 1px dotted;
+}
+
+.undecorated {
```
wibjorn, line 101: let's remove this

css/src/atomics/typography.scss
```diff
@@ -0,0 +1,112 @@
...
+	text-decoration: none;
+}
+
+.normal {
```
wibjorn, line 105: let's remove this

css/src/atomics/typography.scss
```diff
@@ -0,0 +1,112 @@
...
+	font-style: normal;
+}
+
+//Letter Spacing
```
wibjorn, line 109: is this used by us often?

css/src/atomics/typography.scss
wibjorn: on second thought I think we can remove these for now

css/src/atomics/typography.scss
```diff
@@ -0,0 +1,112 @@
...
+
+.heading-h5 {
+	font-size: $font-size-5 !important;
+}
+
+.heading-h6 {
+	font-size: $font-size-6 !important;
+}
+
+// Non-heading sizes
```
wibjorn, line 26: Super huge nitpick: space before and after comments?

css/src/atomics/typography.scss
```diff
@@ -0,0 +1,112 @@
...
+.font-size-m {
+	font-size: $font-size-3 !important;
+}
+
+.font-size-s {
+	font-size: $font-size-5 !important;
+}
+
+.font-size-xs {
+	font-size: $font-size-7 !important;
```
wibjorn, line 48: seems like we're missing some values while including others. how did we come around to deciding which were included and which weren't?

css/src/atomics/typography.scss
asrivastava0308: We have used the letter spacing class in like 12-13 template file . since ming already declared the variable in typography i think we can include the class .

css/src/atomics/typography.scss
asrivastava0308: 👍

css/src/atomics/typography.scss
asrivastava0308: Don't we need them ? we use them a lot in docs-ui and i think we should include them . What do you think ?

css/src/atomics/typography.scss
asrivastava0308: So `font-size-1` to `font-size-6` is already included as part of the heading classes . the only one left are `font-size-0` ,`font-size-7`, `font-size-8` and `font-size-9` . out of which i dont think we need a class for `font-size-7:1rem` . So while creating this i made sure to alternatively give class to the one which are already included as a part of heading sizes . 
$font-size-9: 0.75rem !default; // 12px XS
$font-size-8: 0.875rem !default; // 14px SMALL
$font-size-7: 1rem !default; // 16px 
$font-size-6: 1.125rem !default; // 18px MEDIUM
$font-size-5: 1.25rem !default; // 20px 
$font-size-4: 1.5rem !default; // 24px LARGE
$font-size-3: 1.75rem !default; // 28px 
$font-size-2: 2.125rem !default; // 34px XL
$font-size-1: 2.5rem !default; // 40px 
$font-size-0: 3.375rem !default; // 54px XXL

Or the other option is to include all the font-size but then we will have 10 classes and it can go tilll 4XL . Let me know what you think about this .

css/src/atomics/typography.scss
wibjorn: Sounds good

css/src/atomics/typography.scss
wibjorn: Here's what I'd like to suggestion for now

```
$font-size-9: 0.75rem !default; // 12px XS
$font-size-8: 0.875rem !default; // 14px S
$font-size-7: 1rem !default; // 16px M
$font-size-6: 1.125rem !default; // 18px L
$font-size-5: 1.25rem !default; // 20px XL

$font-size-0: 3.375rem !default; // h0 - let's add this to the heading atomics above.
```

We cover most larger sizes within the `h*` ramp. The `h*` sizes encourage (but don't require) that developers maintain some visual hierarchy with their headings.

The smaller sizes (here listed as `xs - xl`) should be used for other cases (i.e. not headings). For this reason I don't think we need to extend those classes to cover all the values in the typo ramp (however, you'll note that we _could_ do this if for some reason this decision was a mistake).

However, that does leave one value (`$font-size-0`) out in the cold. Since the primary use case for this one is _also a heading_, I think it makes sense to add it to the heading-related atomics instead of the primary oft-used tshirt atomics.

css/src/atomics/typography.scss
asrivastava0308: As discussed offline , we are not gonna use capitalize , lowercase and other classes .

wibjorn (approved):

PR #32, opened 2021-03-22
Add font stack

v-mingho:
task-[378382]

This PR adds the default, monospace, and quote font stack. The font stack replicates the one used by Bulma, which is a comprehensive font stack that covers different type of devices. I also moved two emoji fonts from existing font face rule into the stack. 
- https://github.com/jgthms/bulma/blob/master/sass/utilities/initial-variables.sass
  - -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", "Helvetica", "Arial", sans-serif

For reference, this page includes a table that lists out the devices targeted for each font. 
- https://www.wpmediamastery.com/why-and-how-to-use-a-system-font-stack/
  - How To Implement A System Font Stack

The page describes the thought process for Github's font stack, which is a more simple font stack. The post mentions some previous issues with 'Roboto' and other fonts. However, the post was written a few years ago and I haven't found much evidence that it's still a problem today. 
- https://markdotto.com/2018/02/07/github-system-fonts/
  - -apple-system, BlinkMacSystemFont, Segoe UI, Helvetica, Arial,sans-serif, Apple Color Emoji, Segoe UI Emoji

stylelint is requiring non-keywords to be lower case, so I added quotes to the font names. I think this is okay as it makes a clear distinction between font family names and generic font families (monospace, etc). We could add a list of exceptions to the "value-keyword-case" rule, but it seems hard to maintain.

## Testing

1. Review changes.

### DIFF
--- a/css/src/core/font-stack.scss
+++ b/css/src/core/font-stack.scss
@@ -12,7 +12,6 @@ samp {
 	font-family: $monospace-font-stack !important;
 	-moz-osx-font-smoothing: auto;
 	-webkit-font-smoothing: auto;
-	line-height: 1.1875;
 }
 
 code {
--- a/css/src/core/font-stack.scss
+++ b/css/src/core/font-stack.scss
@@ -12,7 +12,7 @@ samp {
 	font-family: $monospace-font-stack !important;
 	-moz-osx-font-smoothing: auto;
 	-webkit-font-smoothing: auto;
-	line-height: 19px;
+	line-height: 1.1875;
 }
 
 code {
--- a/css/src/core/font-stack.scss
+++ b/css/src/core/font-stack.scss
@@ -1,0 +1,25 @@
+html,
+body {
+	font-family: $normal-font-stack;
+	-webkit-font-smoothing: antialiased;
+	font-weight: $weight-normal;
+	text-rendering: optimizeLegibility;
+}
+
+kbd,
+pre,
+samp {
+	font-family: $monospace-font-stack !important;
+	-moz-osx-font-smoothing: auto;
+	-webkit-font-smoothing: auto;
+	line-height: 19px;
+}
+
+code {
+	direction: ltr;
+	font-family: $monospace-font-stack;
+}
+
+a > code {
+	font-family: $normal-font-stack;
+}
--- a/css/src/tokens/font-stack.scss
+++ b/css/src/tokens/font-stack.scss
@@ -1,6 +1,6 @@
 $monospace-font-stack: 'SFMono-Regular', 'Consolas', 'Liberation Mono', 'Menlo', 'Courier',
-	monospace;
+	monospace !default;
 $normal-font-stack: -apple-system, 'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu',
 	'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue', 'Helvetica', 'Arial', sans-serif,
-	'Apple Color Emoji', 'Segoe UI Emoji';
+	'Apple Color Emoji', 'Segoe UI Emoji' !default;
 $quote-font-stack: 'Arial', 'Helvetica Neue', 'Helvetica', sans-serif !default;
--- a/css/src/tokens/font-stack.scss
+++ b/css/src/tokens/font-stack.scss
@@ -1,0 +1,6 @@
+$monospace-font-stack: 'SFMono-Regular', 'Consolas', 'Liberation Mono', 'Menlo', 'Courier',
+	monospace;
+$normal-font-stack: -apple-system, 'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu',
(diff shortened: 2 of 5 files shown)


css/src/tokens/font-stack.scss
```diff
@@ -0,0 +1,6 @@
+$monospace-font-stack: 'SFMono-Regular', 'Consolas', 'Liberation Mono', 'Menlo', 'Courier',
+	monospace;
+$normal-font-stack: -apple-system, 'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu',
```
jdanyow, line 3: each var should have `!default`

css/src/core/font-stack.scss
```diff
@@ -0,0 +1,25 @@
+html,
+body {
+	font-family: $normal-font-stack;
+	-webkit-font-smoothing: antialiased;
+	font-weight: $weight-normal;
+	text-rendering: optimizeLegibility;
```
jdanyow, line 6: @wibjorn heads up for later (not needed for this PR)- turns out `optimizeLegibility` is often terrible for perf, at some point we should compare with `optimizeSpeed` and reserve `optimizeLegibility` for things like headings.

css/src/tokens/font-stack.scss
v-mingho: Added

css/src/core/font-stack.scss
v-mingho: Thanks for the heads up. Perhaps I can do some comparison now and see if there's a clear winner.

css/src/core/font-stack.scss
```diff
@@ -0,0 +1,25 @@
...
+}
+
+kbd,
+pre,
+samp {
+	font-family: $monospace-font-stack !important;
+	-moz-osx-font-smoothing: auto;
+	-webkit-font-smoothing: auto;
+	line-height: 19px;
```
wibjorn, line 15: testing required, but instead of a pix value, should we do a no-unit value? If we include `1.1875` does it have the same effect?

css/src/core/font-stack.scss
jdanyow: @v-mingho can be a separate work item. Some articles on the subject:
https://tympanus.net/codrops/css_reference/text-rendering/
https://bocoup.com/blog/text-rendering

css/src/core/font-stack.scss
v-mingho: It doesn't have the exact same effect because the unitless value is a factor of the element's font-size (currently 14px in docs-ui). The correct value would be ~1.3571 if we keep the current <pre> font size. Perhaps this line-height should be moved to wherever we'll declare the font-size for <pre> due to this relationship.

wibjorn: @v-mingho agreed, moving it to the rule where we set the font size makes sense.

css/src/core/font-stack.scss
v-mingho: New work item: https://dev.azure.com/ceapex/Engineering/_workitems/edit/397963

css/src/core/font-stack.scss
v-mingho: @wibjorn I merged the PR for Docs-ui. Updated to 1.1875 here with the assumption that the pre will inherit a font size of 1rem/16 px. As there won't be any default element styles, I think this line height remain here for now.

wibjorn: Not seeing any changes. Could you help me understand why the unitless value would need to be different in the two places? My understanding is the the unitless values change in reference to the base font-size. Wouldn't we want the height to scale between font size changes?

v-mingho: Github isn't doing a good job at reflecting the latest changes. Click on the Files Changed tab for the most recent version.

The unitless value is multiplied by the element's own font size rather than the base font size, so it will scale upon font size changes of the <pre> element. It seems like Normalize.css sets the <pre> font size to 1em and <html> line height to 1.15. 

Without any default styling, then presumably the <pre> element will inherit the base font size of 16px (?). To set an unitless value that's equivalent to 19px, the factor will need to be 19/16 = 1.1875. 

In Docs-ui, we're setting the <pre> default font size to 0.875rem (14px) so the unitless value equivalent to 19px would be 19/14 = 1.3571, hence the two different values. 

If we pair a line height 1.3571 with 16px font size, it will look like this:
image

1.1871 with 16px font size:
image

1.1871 with 14px:
image

1.3571 with 14px
image

v-mingho: I compared a 16px font with line height 1.1871 vs the Normalize default 1.15. The visual difference is actually rather minimal. Shall we just remove the line height and take Normalize's default for now?

image

v-mingho: Line height removed

wibjorn (approved):

PR #39, opened 2021-03-30
Adding color-scheme CSS property

ollips:
task-391082

Making the same update in the atlas as in docs-ui to support scrollbar theming.

### DIFF
--- a/css/src/core/themes.scss
+++ b/css/src/core/themes.scss
@@ -76,5 +76,11 @@ $theme-names: map.keys($themes);
 @each $key, $value in $themes {
 	.theme-#{$key} {
 		@extend %theme-base-styles;
+
+		@if $key == $root-theme {
+			color-scheme: light;
+		} @else {
+			color-scheme: dark;
+		}
 	}
 }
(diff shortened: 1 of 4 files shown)


css/src/core/themes.scss
```diff
@@ -76,5 +76,11 @@ $theme-names: map.keys($themes);
 @each $key, $value in $themes {
 	.theme-#{$key} {
 		@extend %theme-base-styles;
+
+		@if $key == $root-theme {
+			color-scheme: light;
+		} @else {
```
wibjorn, line 82: Would this work if we added a new theme? How might be do this in a way that could allow for flexibility in the future?

ollips: Update approach

css/src/tokens/schemes.scss
```diff
@@ -0,0 +1,5 @@
+$schemes: (
```
wibjorn, line 1: I think an explanatory comment here would be helpful here. One saying that this object's keys should match themes and that it's value will set the color-scheme css property for that theme.

wibjorn:
Needs changeset?

css/src/tokens/schemes.scss
ollips: Updated

wibjorn (approved):

PR #2, opened 2021-02-17
🛑 [never merge] plan document for /css folder

wibjorn:
# Welcome to the plan for the /css folder

The main point of this pull request is to provide a place for use look at the plan document together. This PR will never merge. The plan document is here: 

https://github.com/microsoft/atlas-design/pull/2/files#diff-7b7df4436312ddd4bf006ef1f51da2644ab69a5a58406df81a77d7321bb21e33

task-376343

Note: there are several changes in the files changed tab. These can be ignored, I've been using the PR to proof of concept certain ideas well, including changesets.

PR #40, opened 2021-03-31
Adding Component Folder

asrivastava0308:
task-378405

Add Component Folder And readme 

Note : Can't add index.scss file under components as the current lint process don't let the build to pass .


css/src/components/index.scss
```diff
@@ -0,0 +1 @@
+@import '';
```
wibjorn, line 1: Can the file remain empty while satisfying lint? If not, we can remove it and add a README.md

.changeset/honest-queens-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-css': minor
```
wibjorn, line 2: `patch`

css/src/components/index.scss
wibjorn: Readme could be populated with some of details from the plan: 
https://github.com/microsoft/atlas-design/pull/2

asrivastava0308: Current lint process not allowing the file to remain empty . So removed it

.changeset/honest-queens-listen.md
asrivastava0308: changed it to patch , as no changeset create an warning

css/src/components/component.md
```diff
@@ -0,0 +1,8 @@
+#### Components
+
+Create new component when you have a complex layout that requires many helper classes to recreate. Keep following in mind while creating a New component .
```
v-mingho, line 3: "Create a new component ......... recreate. Keep the following in mind while creating a new component."

css/src/components/component.md
```diff
@@ -0,0 +1,8 @@
...
+
+- Component must be reusable.
+- Should be output producing.
```
v-mingho, line 6: I'm not entirely sure what this line is supposed to mean. Can it be reworded to something more simple?

wibjorn: We can probably remove 
```
- Component must be reusable.
- Should be output producing.

File: css/src/components/component.md [hunk 5]
```diff
@@ -0,0 +1,8 @@
...
+- File name should match the class of the component.
+- Top level classes per file: 1
```
wibjorn, line 8: There should only be one top level class per file.

wibjorn (approved):

PR #52, opened 2021-04-07
Adding !default to focus-visible-use-polyfill

ollips:
task-404427

Adding `!default` to `$focus-visible-use-polyfill`

### DIFF
--- a/.changeset/moody-seas-promise.md
+++ b/.changeset/moody-seas-promise.md
@@ -1,0 +1,5 @@
+---
+'@microsoft/atlas-css': minor
+---
+
+Adding !default flag to focus-visible-use-polyfill.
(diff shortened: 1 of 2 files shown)


v-mingho (approved):

v-mingho (changes requested):
Changeset?

.changeset/moody-seas-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-css': minor
```
wibjorn, line 2: I think this is patch, but it's not super important, since another patch is coming.

wibjorn (approved):

v-mingho (dismissed):

.changeset/moody-seas-promise.md
ollips: Updated to patch!

v-mingho (approved):

PR #94, opened 2021-04-16
remove -dark and -light element of theme map, color page

wibjorn:
Task: task-392645

Link: preview-94

After some intregation with existing docs styling, we've determined we don't need three background variants. Removing -dark and -light versions for body and alternate backgrounds.


css/src/atomics/colors.scss
```diff
@@ -138,12 +129,12 @@
...
 	&:nth-of-type(even) {
 		outline-color: $text;
-		background-color: $body-background-medium;
+		background-color: $body-background;
 	}
 
 	&:nth-of-type(odd) {
 		outline-color: $text;
-		background-color: $body-background-dark;
+		background-color: $alternate-background;
```
v-mingho, line 137: Should we use `$body-background-medium;` here?

Currently `background-color-alternating-grey` looks like this:

image

while `background-color-alternating-grey-reverse` looks like this:
image

wibjorn: Good save! Updated.

v-mingho (approved):

PR #98, opened 2021-04-20
markdown example, code block and markdown example styles.

wibjorn:
This PR attempts, perhaps to do a few too many things, but they were all somewhat linked together. Here's what's going on:

- A `.markdown` class was created to house otherwise bare element styles ...
- With the exception of `.code-block`. This element was created with mostly specific classes that are designed to display code-header and the code body together.
- Note: consumers of Atlas may also specify `$include-bare-code-styles: true` and then the bare styles (`<pre><code>${code}</code></pre>` will be added within the `.markdown class. This is off by default in favor of more explicit classes.
- Creates a page to showcase `.markdown`. 
- Creates a page to showcase `.code-block`.
- Decouples code block backgrounds from `body-` background properties in themes.
- From here on, there were a few additions simply here to make the site a bit nicer.
- Display helpers were necessary for this, and were on the agenda anyway, so folks can now use atomics like `.display-flex`.
- Updates minorly updates the html for the table of contents.
- Adds custom renderers to our `marked` setup. To begin with, I've only extended `paragraph`, `code`, and `headings`. Each is wrapped in markdown. Why include it so granularly? Because we don't want example markup to be affected, or otherwise messed up by the bare element styling within markdown.
- Small update to the lint rule, which needed a valid word added.
- Changes how we generate example elements (which wasn't working quite right). Extend this so markdown will also render within example elements. (Only `html` and `markdown` are example generating languages).
- Installs `rfs`

Task: task-412890
Task: task-412892

Link: preview-98

## Testing

Markdown should render properly.

[markdown](https://design.docs.microsoft.com/pulls/98/components/markdown.html)

Code blocks should render properly.

[code block](https://design.docs.microsoft.com/pulls/98/components/code-block.html)


css/src/atomics/display.scss
```diff
@@ -0,0 +1,27 @@
...
+	}
+
+	.display-#{$item}-widescreen {
+		@include widescreen {
+			display: #{$item} !important;
+		}
+	}
+}
+
+[hidden] {
```
v-mingho, line 25: Are there plans to also add the `is-hidden` is-invisible` classes?

ollips:
In the code block body, the text is almost invisible to me:

image

css/src/atomics/display.scss
wibjorn: I could see `.display-none` being added to match with our atomic class structure.

wibjorn: Added 'none' as well

wibjorn:
@ollips We neglected to add a theme class to our html element! That will do the trick for now. We'll address theme switching later.

css/src/components/code-block.scss
```diff
@@ -0,0 +1,68 @@
...
+		margin: 0;
+		padding: $code-block-spacing;
+		border: 1px solid $border;
+		background-color: $code-block;
+		font-size: $code-font-size;
+		line-height: 1.3571;
+		white-space: pre;
+		word-spacing: normal;
+		overflow: auto;
+		font-smooth: auto;
```
v-mingho, line 21: I noticed that `font-smooth` and `-webkit-overflow-scrolling` doesn't seem to be working (would need to be `webkit-font-smoothing`/`-moz-osx-font-smoothing`). It's the same in Docs-UI. Should we remove them as they're non-standard? 

image

https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth
https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-overflow-scrolling

wibjorn: Re: overflow scrolling: My understanding is that its just an additive on iOS safari, so it makes sense that it would never apply on a desktop machine.

wibjorn: Re: font-smoothing - great catch! I would have expected auto-prefixer to handle this one, but it seems it does not. I've added the vendor prefixes.

site/src/components/markdown.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+## Ordered list
+
+```markdown
+1. First list item
+2. Second list item
+   1. First level of nested list
+      1. Second level of nested list
+      1. third level of nested list
```
v-mingho, line 52: Wouldn't this need another indentation in order to be third level?

site/src/components/markdown.md
```diff
@@ -0,0 +1,74 @@
...
+   1. Second level again
```
v-mingho, line 53: Same as above. Right now it looks to me that it's actually at the same level as `First level of nested list`

wibjorn: 👍Looks like the text should be updated.

site/src/components/markdown.md
wibjorn: 👍Looks like the text should be updated.

wibjorn: done

v-mingho (approved):

ollips (approved):

PR #101, opened 2021-04-20
update with very basic site layout

wibjorn:
task-414294 

This PR puts up a site layout. It is likely that this isn't going to be the layout forever. But it's simple, and does the trick. It also provides a little extra room looking forward. It includes all the major elements that most info-based sites would want.

- header
- main - houses the content
- aside (empty for now)
- nav - left hand nav
- footer (empty for now)

Testing:

Pop open the preview link.

[standard page](https://design.docs.microsoft.com/pulls/101/components/markdown.html)
[tokens page](https://design.docs.microsoft.com/pulls/101/tokens/colors.html)

preview-101

Test the layout with some labels and borders

```js
document.documentElement.classList.add('debug');
```

Test with some content in the `<aside>`

```js
document.getElementById('aside').innerHTML = document.getElementById('nav').innerHTML;
```

### DIFF
--- a/css/src/atomics/typography.scss
+++ b/css/src/atomics/typography.scss
@@ -82,6 +82,10 @@
 	text-decoration: underline;
 }
 
+.text-decoration-none {
+	text-decoration: none;
+}
+
 //Letter Spacing
 
 .letter-spacing-wide {
--- a/css/src/atomics/typography.scss
+++ b/css/src/atomics/typography.scss
@@ -79,15 +79,15 @@
 }
 
 .text-decoration-underline {
-	text-decoration: underline;
+	text-decoration: underline !important;
 }
 
 .text-decoration-none {
-	text-decoration: none;
+	text-decoration: none !important;
 }
 
 //Letter Spacing
 
 .letter-spacing-wide {
-	letter-spacing: $letter-spacing-wide;
+	letter-spacing: $letter-spacing-wide !important;
 }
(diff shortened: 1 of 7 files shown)


css/src/atomics/typography.scss
```diff
@@ -82,6 +82,10 @@
 	text-decoration: underline;
 }
 
+.text-decoration-none {
+	text-decoration: none;
```
ollips, line 86: Does the `!important` need to be added here?

site/src/scaffold/token.html
```diff
@@ -10,37 +10,50 @@
...
-			</div>
-			<pre><code>{{#tokens}}{{name}}: {{value}}
+				
+				<div class="padding-block-m">
+					<a href="https://github.com/microsoft/atlas-design/tree/main{{cssTokenSource}}">View on Github</a>
+				</div>
+				<div class="code-block">
+					<div class="code-block-body">
+					<pre><code>{{#tokens}}{{name}}: {{value}}
 {{/tokens}}</code></pre>
```
ollips, line 27: Can you fix the indent here, please? Looks off a bit.

css/src/atomics/typography.scss
wibjorn: Yes! In fact, it seems like we forgot to add !important to the two above.

Important is used with atomic classes to ensure that single class selectors will be effective even against more specific component classes.

site/src/scaffold/token.html
wibjorn: I would like to , but this is a `<pre>` tag, so we need like this for now to ensure the code renders correctly.

ollips: Oh, right! ok, resolving

css/src/atomics/typography.scss
wibjorn: (I've added it to the other here)

ollips (approved):

PR #118, opened 2021-04-27
component readme

wibjorn:


v-mingho (approved):

ollips (approved):

v-mingho (commented):
I still have mixed thoughts about whether adding documentation would require a changeset. Given that the change log is for alerting users about changes, wouldn't we want to let users know that there's new documentation?

PR #124, opened 2021-04-27
Add Flex Responsive Classes 

asrivastava0308:
Task: task-414263

Link: preview-124

Creating Responsive Classes For Flex

## Testing

1. Step one
2. Step two
   Expected result:

## Additional information

[Optional]

### DIFF
--- a/css/src/atomics/flex.scss
+++ b/css/src/atomics/flex.scss
@@ -1,5 +1,5 @@
 @use 'sass:map';
-
+$include-flex-desktop: true;
 $flex-atomics: (
 	'justify-content': (
 		flex-start,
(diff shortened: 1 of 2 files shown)


css/src/atomics/flex.scss
```diff
@@ -1,5 +1,5 @@
 @use 'sass:map';
-
+$include-flex-desktop: true;
```
wibjorn, line 2: the default would be false, to reduce output wouldn't it?

css/src/atomics/flex.scss
```diff
@@ -57,3 +57,28 @@ $flex-atomics: (
...
+	@each $key in map.keys($flex-atomics) {
+		@each $value in map.get($flex-atomics, $key) {
+			.#{$key}-#{$value}-tablet {
+				#{$key}: $value !important;
+			}
+		}
+	}
+}
+
+@if $include-flex-desktop == true {
```
wibjorn, line 71: I think this is a cool strategy we talked about, but for now, perhaps we should just remove this. We can circle back later (😆 if/when -desktop) becomes necessary.

css/src/atomics/flex.scss
wibjorn: second thought, let's remove it for now

wibjorn (approved):

PR #135, opened 2021-05-03
Adding font-stack helpers

ollips:
Task: task-419077

Link: preview-135

Creating helper classes `font-family-monospace` and `font-family-quote`.

## Testing

1. Open any page. Inspect any text paragraph/title. Try to add either of the newly created classes. You should see how the font family changes accordingly to the applied class.

### DIFF
--- a/css/src/atomics/typography.scss
+++ b/css/src/atomics/typography.scss
@@ -1,3 +1,15 @@
+// Font stack
+
+.is-monospace {
+	font-family: $monospace-font-stack !important;
+	-moz-osx-font-smoothing: auto;
+	-webkit-font-smoothing: auto;
+}
+
+.is-quote {
+	font-family: $quote-font-stack !important;
+}
+
 // Heading sizes
 
 .font-size-h0 {
--- a/css/src/atomics/typography.scss
+++ b/css/src/atomics/typography.scss
@@ -1,12 +1,12 @@
-// Font stack
+// Font family
 
-.is-monospace {
+.font-family-monospace {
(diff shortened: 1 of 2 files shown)


css/src/atomics/typography.scss
```diff
@@ -1,3 +1,15 @@
+// Font stack
+
+.is-monospace {
```
wibjorn, line 3: These should match the same pattern as other atomics.

```
.font-family-monospace
.font-family-quote
```

css/src/atomics/typography.scss
```diff
@@ -1,3 +1,15 @@
...
+	font-family: $monospace-font-stack !important;
+	-moz-osx-font-smoothing: auto;
+	-webkit-font-smoothing: auto;
+}
+
+.is-quote {
```
wibjorn, line 9: This item, at present at least, has very focused usefulness. I think we can remove it from the framework.

wibjorn, line 9: `.font-family-quote`

css/src/atomics/typography.scss
ollips: updated

css/src/atomics/typography.scss
ollips: updated

wibjorn:
Does this need documentation?

wibjorn (approved):
Looks good, but do we need documentation for this?

ollips:
> Looks good, but do we need documentation for this?

Gonna be covered in a separate task ;)

PR #202, opened 2021-05-19
[bug] colors atomics order

ollips:
Task: task-429358

Link: preview-202

Colors atomics output is in the incorrect order. `background-color-primary-tablet background-color-danger` did not work as expected.

## Testing

1. Visit https://design.docs.microsoft.com/pulls/202/atomics/position.html
2. In the example section, using dev tools, add `background-color-primary-tablet background-color-danger` classes to the example container. Nothing visually should change. You should see the `background-color-primary-tablet` takes higher priority over `background-color-danger`. Resize the browser screen, you should see how on mobile example section background color changes.

image

### DIFF
--- a/css/src/atomics/colors.scss
+++ b/css/src/atomics/colors.scss
@@ -60,30 +60,6 @@
 	}
 }
 
-@include desktop {
-	@each $name, $color-set in $colors {
-		$base: nth($color-set, $color-index-base);
-		$invert: nth($color-set, $color-index-invert);
-
-		.background-color-#{$name}-desktop {
-			outline-color: $invert;
-			background-color: $base !important;
-		}
-	}
-}
-
-@include widescreen {
-	@each $name, $color-set in $colors {
-		$base: nth($color-set, $color-index-base);
-		$invert: nth($color-set, $color-index-invert);
-
-		.background-color-#{$name}-widescreen {
-			outline-color: $invert;
-			background-color: $base !important;
-		}
-	}
-}
-
 // Text color
 
 .color-text {
--- a/css/src/atomics/colors.scss
+++ b/css/src/atomics/colors.scss
@@ -51,10 +51,6 @@
 @include tablet {
 	@each $name, $color-set in $colors {
 		$base: nth($color-set, $color-index-base);
-		$background: nth($color-set, $color-index-background);
-		$dark: nth($color-set, $color-index-dark);
-		$hover: nth($color-set, $color-index-hover);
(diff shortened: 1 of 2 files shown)


v-mingho (dismissed):

css/src/atomics/colors.scss
```diff
@@ -69,6 +48,54 @@
 	}
 }
 
+@include tablet {
```
wibjorn, line 51: This is a minor change, but I think that we should keep control flow on the outermost indent whenever possible. Media queries can follow inside. Thoughts? I think it would be good to standardize this, since we'll be writing loops in other places too.

```suggestion
@each [...] {
   $vars: value;
   
   @tablet {
   
   }
}


```

css/src/atomics/colors.scss
```diff
@@ -69,6 +48,54 @@
...
+		.background-color-#{$name}-desktop {
+			outline-color: $invert;
+			background-color: $base !important;
+		}
+	}
+}
+
+@include widescreen {
+	@each $name, $color-set in $colors {
+		$base: nth($color-set, $color-index-base);
```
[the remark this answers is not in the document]
ollips: Good catch, updated!

wibjorn (dismissed):

v-mingho (dismissed):

css/src/atomics/colors.scss
```diff
@@ -69,6 +48,18 @@
...
+		$invert: nth($color-set, $color-index-invert);
+
+		.background-color-#{$name}-tablet {
+			outline-color: $invert;
+			background-color: $base !important;
+		}
+	}
+}
```
ollips, line 61: After an additional investigation, it turned out there were no `desktop`/`widescreen` rules previously defined so removed those two loops.

But the `mobile` existed, so keeping the `tablet` instead to match the mobile-first approach. Not sure if I should though. Do we need to provide that much flexibility here? 🤔

@wibjorn @v-mingho @asrivastava0308

wibjorn: I think providing tablet only is fine for now, given that we've done that with some other helpers (@asrivastava0308 can chime in on that). Tablet is the place where major layout shifts happen.

wibjorn (dismissed):

v-mingho (approved):

wibjorn (commented):
Changeset needed "patch"

v-mingho (approved):

wibjorn (approved):

wibjorn (approved):

PR #205, opened 2021-05-21
a code block built for displaying available atomic classes

wibjorn:
As we continue through the Atomics documentation process, we're finding it difficult to structure the pages, particularly because in some case the amount of classes don't fit neatly on a page.

Take a look at this PR, does it help solve this?

- create a code block with the language `atomics-filter`, if this is there, the code block's height is constrained individual lines (which should correspond to classes) are filterable.
- creates a spacing documentation page to test it out. Does not do the full page, what I'd suggest we place under an "Available classes" section.

Task: task-421762

Link: preview-205

https://design.docs.microsoft.com/pulls/205/atomics/spacing.html


site/src/scaffold/scripts/code-filter.ts
```diff
@@ -0,0 +1,38 @@
...
+			const id = target.dataset.codeFilterInput;
+			const codeElt = document.querySelector(`[data-code-filter-code="${id}"]`);
+
+			if (!id || !codeElt) {
+				return;
+			}
+
+			// If we haven't save the whole contents of the block, do so.
+			memory[id] = memory[id] || codeElt.textContent.split('\n');
+			// If set text content to things that match the given term
```
v-mingho, line 25: I think the "if" here can be removed. "Set text content to code that match the given term"?

site/src/scaffold/code-block.scss
```diff
@@ -3,6 +3,25 @@ $code-header-color: $text !default;
...
+	.code-block-header-filter {
+		max-width: 12ch;
+		flex-shrink: 1;
+		border: none;
+		font-size: $font-size-9;
+		background: $body-background-medium;
+		margin-right: $spacer-3;
+		padding-inline: $spacer-2;
+		padding-block: $spacer-1;
+		// font-family: $monospace-font-stack;
```
v-mingho, line 17: My vote here would be to not use monospace here.

plugins/parcel-transformer-markdown-html/lib/index.js
```diff
@@ -15,9 +15,32 @@ const languageDisplayNames = {
...
 };
 
+let filterIds = 0;
+
+function createFilterableCodeBlock(code, language, displayName) {
+	filterIds++;
+	return `
+	<div class="code-block margin-top-xs" style="min-height: 20vh">
+		
+		<div class="code-block-header">
```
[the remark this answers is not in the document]
asrivastava0308: image

site/src/scaffold/code-block.scss
wibjorn: removed comment

plugins/parcel-transformer-markdown-html/lib/index.js
```diff
@@ -15,9 +15,32 @@ const languageDisplayNames = {
...
+			<span class="code-block-header-language" data-hljs-language="${language}">${displayName}</span>
+			<input
+			class="code-block-header-filter align-self-center"
+			data-code-filter-input="${filterIds}"
+			placeholder="Filter ..."
+			type="search" />
+		</div>
+		<div class="code-block-body" style="max-height: 20vh; overflow-y: scroll;">
```
asrivastava0308, line 38: I think we have to change max-height as 30vh in order to achieve the desired result !

plugins/parcel-transformer-markdown-html/lib/index.js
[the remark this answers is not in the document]
ollips: +1, just use `padding-`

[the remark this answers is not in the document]
wibjorn: These are tied to Docs for the moment, keeping as is. We can updated them as part of a real design effort.

plugins/parcel-transformer-markdown-html/lib/index.js
wibjorn: Updated!

plugins/parcel-transformer-markdown-html/lib/index.js
```diff
@@ -15,9 +15,32 @@ const languageDisplayNames = {
...
+	'atomics-filter': 'Atomics',
+	atomics: 'Atomics'
 };
 
+let filterIds = 0;
+
+function createFilterableCodeBlock(code, language, displayName) {
+	filterIds++;
+	return `
+	<div class="code-block margin-top-xs" style="min-height: 20vh">
```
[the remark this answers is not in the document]
wibjorn: Updated

ollips (approved):

v-mingho (approved):

PR #207, opened 2021-05-25
atomics documentation template example

ollips:
Task: task-430464

Proposing a template structure for the atomics documentation page.

## Testing

1. Review the file in the site's root folder - the `site/src/atomics.template`
2. Check out the proposed structure, share your thoughts :)


ollips (commented):
Any thoughts on how to include info about the `{screensize}` available? I was thinking about including it into the description paragraph at the top, and if needed, that sentence can include a link to the future "Atomics overview" page, where it is explained what the `{screensize}` is 🤔 😵

@wibjorn @v-mingho @asrivastava0308

v-mingho:
If responsiveness is going to have its own section, should it go towards the bottom (it may make more sense for a reader after seeing the Usage section)?

Courtny:
@ollips I wonder if we could spin up a Conditional Classes section of documentation that would include details on responsiveness? Something like this: [Conditional Classes](https://stackoverflow.design/product/guidelines/conditional-classes/)

ollips:
> @ollips I wonder if we could spin up a Conditional Classes section of documentation that would include details on responsiveness? Something like this: [Conditional Classes](https://stackoverflow.design/product/guidelines/conditional-classes/)

@Courtny  Sure, I'll think about how to integrate it into the template :)

site/src/template.md
```diff
@@ -0,0 +1,93 @@
...
+
+# Page title (H1)
+
+The first thing is the general description.
+
+Then the list of some general classes can be listed, if there are not many of them, or css style specific details
+
+_Example_
+
+The following spacing sizes are supported:
```
wibjorn, line 15: The following spacing sizes are supported:- `s`- `m`- `l`- `none`

Seems like the above could be removed. Isn't it fine to reference this in the table?

site/src/template.md
```diff
@@ -0,0 +1,93 @@
+---
+title: Title
```
wibjorn, line 2: Should this file be called `atomics.template`, so that it isn't built into the site?

site/src/template.md
```diff
@@ -0,0 +1,93 @@
...
+- `l`
+- `none`
+
+Otherwise consider using tables to organize information.
+
+_Example. Table 1_
+
+| cssproperty | value                                                   | screensize |
+| ----------- | ------------------------------------------------------- | ---------- |
+| `margin`    | `xxs`, `xs`, `s`, `m`, `l`, `xl`, `xxl`, `xxxl`, `none` | `tablet`   |
```
wibjorn, line 28:
```md
| `<name>` | `<value>`, `<value>`, `<value>` | `<size>
```

site/src/template.md
```diff
@@ -0,0 +1,93 @@
...
+| `padding`   | `xxs`, `xs`, `s`, `m`, `l`, `xl`, `xxl`, `xxxl`, `none` | `tablet`   |
+
+_Example. Table 2_
```
wibjorn, line 31: Remove large table example?

site/src/template.md
```diff
@@ -0,0 +1,93 @@
...
+| `xs`   | 1rem / 16px   | .margin-xs     | .padding-xs     |
+| `s`    | 1.5rem / 24px | .margin-s      | .padding-s      |
+| `m`    | 2rem / 32px   | .margin-m      | .padding-m      |
+| `l`    | 3rem / 48px   | .margin-l      | .padding-l      |
+| `xl`   | 4rem / 64px   | .margin-xl     | .padding-xl     |
+| `xxl`  | 6rem / 96px   | .margin-xxl    | .padding-xxl    |
+| `xxxl` | 8rem / 128px  | .margin-xxxl   | .padding-xxxl   |
+| `None` | 0             | .margin-none   | .padding-none   |
+
+## Usage (H2)
```
wibjorn, line 45: I don't think we need `(H2)` here. We should all know what ## means by now 👍

wibjorn: (same with the below)

site/src/template.md
```diff
@@ -0,0 +1,93 @@
...
+
+Next block is a general usage example.
```
wibjorn, line 47: > Provide a focused example that displays this atomic's usage.

site/src/template.md
```diff
@@ -0,0 +1,93 @@
...
+
+```html
+<div class="border padding-s">
+	<p>Text block</p>
```
wibjorn, line 51: Nit: `Code block`?

site/src/template.md
```diff
@@ -0,0 +1,93 @@
...
+</div>
+```
+
+if there are tokens variations available such as font weight in typography, or radius rules in border, the next section should be about them.
```
wibjorn, line 55: We should describe the next section _in_ the next section, not in the paragraph before it. Move text down?

site/src/template.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Modifier heading (H3)
+
+Currently two classes are available to set border radius:
```
wibjorn, line 59: Let's not use any specific examples here; keep everything general.

site/src/template.md
```diff
@@ -0,0 +1,93 @@
...
+
+- `.border-radius`
+- `.border-radius-l`
+
+_Example_
```
wibjorn, line 64: Delete?

site/src/template.md
```diff
@@ -0,0 +1,93 @@
...
+	<p>Default radius</p>
+</div>
+```
+
+## Available classes
+
+List of all available classes:
+
+```atomics-filter
+.margin-xs
```
wibjorn, line 77: Something general, like this?

```
.<name>-<value-1>
.<name>-<value-2>
.<name>-<value-3>
.<name>-<value-1>-<screensize>
.<name>-<value-2>-<screensize>
.<name>-<value-3>-<screensize>
[...etc]

<!--StartFragment-->```atomics-filter.margin-xs<!--EndFragment-->

File: site/src/template.md (hunk 5 above)
Comment (ollips) [reply site/src/template.md:45]: I wanted to highlight specifically what heading levels we want to use depending on where are we in the file. Of course, we all know the difference between `#`, `##`, `###`. etc but when you land on a page you might now catch it right away.

So here is a question - will that file be public? Cause I thought we're creating it mostly for ourselves 🤔

File: site/src/template.md (hunk 1 above)
Comment (ollips) [reply site/src/template.md:15]: I think it really depends on a specific case, if you need to list just 3 possible values - does it worth it to make the whole table and think about the proper titles for row/columns in these kinda cases?

File: site/src/template.md (hunk 2 above)
Comment (ollips) [reply site/src/template.md]: Updated

File: site/src/template.md (hunk 3 above)
Comment (ollips) [reply site/src/template.md]: Updated

File: site/src/template.md (hunk 4 above)
Comment (ollips) [reply site/src/template.md]: Updated

File: site/src/template.md (hunk 5 above)
Comment (ollips) [reply site/src/template.md]: Updated

File: site/src/template.md (hunk 6 above)
Comment (ollips) [reply site/src/template.md]: Updated

File: site/src/template.md (hunk 8 above)
Comment (ollips) [reply site/src/template.md]: Updated

File: site/src/template.md (hunk 9 above)
Comment (ollips) [reply site/src/template.md]: Updated

File: site/src/template.md (hunk 10 above)
Comment (ollips) [reply site/src/template.md]: Updated

File: site/src/template.md (hunk 11 above)
Comment (ollips) [reply site/src/template.md]: Updated

File: site/src/template.md (hunk 1 above)
Comment (wibjorn) [reply site/src/template.md]: Yes, I think we should try to present the information in a mostly uniform way, and I don't think a list is the right one.

File: site/src/template.md (hunk 3 above)
Comment (wibjorn) [reply site/src/template.md]: Doesn't look like this one got addressed

File: site/src/template.md (hunk 4 above)
Comment (wibjorn) [reply site/src/template.md]: Does

File: site/src/template.md (hunk 5 above)
Comment (wibjorn) [reply site/src/template.md]: Still seeing a number of `(H*)` throughout the file

File: site/src/template.md (hunk 4 above)
Comment (wibjorn) [reply site/src/template.md]: Looks like it's still here

**Review by wibjorn — CHANGES_REQUESTED** [review]
Unaddressed comments.

File: plugins/parcel-transformer-markdown-html/lib/index.js [hunk 12]
```diff
@@ -24,8 +24,7 @@ let filterIds = 0;
 function createFilterableCodeBlock(code, language, displayName) {
 	filterIds++;
 	return `
-	<div class="code-block margin-top-xs" style="min-height: 30vh">
```
wibjorn, line 27: revert

site/src/atomics.template
```diff
@@ -0,0 +1,73 @@
...
+<div class="border border-radius padding-s">
+	<p>Default radius</p>
+</div>
+```
+
+### Responsive rules
+
+If there are screen-specific class variations are available they can be described in this section.
+
+_Example_
```
wibjorn, line 41: remove `_Example_`?

site/src/atomics.template
```diff
@@ -0,0 +1,73 @@
...
+
+# Page title
+
+The first thing is the general description.
+
+Then the list of some general classes can be listed, if there are not many of them, or css style specific details
+
+_Example_
+
+The following spacing sizes are supported: `-s`, `-m`, `-l`, `-none`.
```
wibjorn, line 15: > The following spacing sizes are supported: `-s`, `-m`, `-l`, `-none`.

Remove?

site/src/atomics.template
```diff
@@ -0,0 +1,73 @@
...
+```
+
+### Modifier heading
+
+if there are tokens variations available such as font weight in typography, or radius rules in border, the next section(s) should be about them.
+
+_Example_
+
+```html
+<div class="border border-radius padding-s">
```
wibjorn, line 56: I think placing `border` in here make it appear like we require border in our example.

site/src/atomics.template
```diff
@@ -0,0 +1,73 @@
...
+
+```html
+<div class="border-tablet padding-s">
+	<p>Border on tablet and up</p>
+</div>
+```
+
+### Modifier heading
+
+if there are tokens variations available such as font weight in typography, or radius rules in border, the next section(s) should be about them.
```
wibjorn, line 51: Capitalize

plugins/parcel-transformer-markdown-html/lib/index.js
ollips: reverted

site/src/atomics.template
ollips: What exactly should be capitalized?

site/src/atomics.template
ollips: Updated

site/src/atomics.template
ollips: Updated

site/src/atomics.template
ollips: Removed.

site/src/atomics.template
wibjorn: `if` -> `If`

ollips: Updated

site/src/atomics.template
```diff
@@ -0,0 +1,59 @@
...
+
+```html
+<div>
+  <button class="button">Click me</button>
+</div>
+```
+
+### Responsive rules
+
+If there are screen-specific class variations are available they can be described in this section.
```
v-mingho, line 29: remove the second `are`?

site/src/atomics.template
```diff
@@ -0,0 +1,59 @@
...
+| ----------- | ------------------------------- | ---------- |
+| `<name>`    | `<value>`, `<value>`, `<value>` | `<size>`   |
+
+## Usage
+
+Provide a focused example that displays this atomic's usage.
+
+```html
+<div>
+  <button class="button">Click me</button>
```
v-mingho, line 23: Should the example here also be `border` to match the next two examples? 
- border
- border responsive
- border modifiers

ollips: The previous example that was here before was updated per Will's request, so it does not seem like we **require** to have `border` in examples.

site/src/atomics.template
ollips: ugh, that's a shame. Updated :D

v-mingho (approved):

wibjorn (approved):

PR #209, opened 2021-06-01
Adding basic table tokens/atomics

ollips:
Task: task-[work-item-number]

Link:
  - [border-page](https://design.docs.microsoft.com/pulls/209/atomics/border.html) - Applied tables as an example of how they will turn out in markdown 
  - [table-page](https://design.docs.microsoft.com/pulls/209/components/table.html)

## Testing

1. Visit pages above.
2. Share your feedback :)


css/src/atomics/table.scss
```diff
@@ -0,0 +1,108 @@
...
+		tbody tr:nth-of-type(odd) {
+			background-color: $box-shadow-color-light;
+		}
+	}
+
+	&.min {
+		width: unset;
+	}
+
+	&.has-inner-borders {
```
wibjorn, line 67: Let's hold on this until tomorrow's sync and thing about what we want to do with `has-*` modifiers and `is-*` modifiers.

css/src/atomics/table.scss
```diff
@@ -0,0 +1,108 @@
+/* stylelint-disable selector-no-qualifying-type */
```
wibjorn, line 1: instead of disabling, we should fix the rules.

css/src/atomics/table.scss
ollips: Oh, you're right, I was thinking about changing it but forgot 🤦🏻‍♀️ Let's sync tomorrow!

css/src/atomics/table.scss
ollips: I need some help with it cause I don't understand why I'm getting this particular error on these places:

```
.table {
  td,
  th {
    &.nowrap {
      white-space: nowrap;
    }

    &.is-narrow {
      width: $table-cell-narrow-width;
    }
  }
}
```

site/src/components/table.md
```diff
@@ -0,0 +1,256 @@
...
+</table>
+```
+
+## Modifiers
+
+These classes are available to style tables in different ways, feel free to mix and match them!
+
+| Class                      | Usage                                               |
+| -------------------------- | --------------------------------------------------- |
+| `table-striped`            | Adds background color to every odd row              |
```
v-mingho, line 60: Should this modifier follow the `is-*` pattern as well? If not, how should we determine which modifiers do/do not? With the removal of `has-*`, I think we'll have to accept the fact that sometimes it won't read as well.

ollips: Good catch! Renamed :)

css/src/mixins/media-queries.scss
```diff
@@ -40,3 +40,14 @@
 		@content;
 	}
 }
+
+/* 
+ * EXCEPTIONS
+ * Please be careful with using these classes. They don't follow the mobile-first approach but might be extremely useful in some use-cases.
+ */
+
+@mixin extremely-needed-mobile {
```
v-mingho, line 49: Trying to think what other name would make sense for these exception cases. What do you think about using a `exception` prefix?

css/src/components/table.scss
```diff
@@ -0,0 +1,128 @@
...
+		}
+	}
+
+	&.is-striped {
+		tbody tr:nth-of-type(odd) {
+			background-color: $box-shadow-color-light;
+		}
+	}
+
+	&.is-min {
```
wibjorn, line 63: To continue this work, but ensure that we don't add anything were the names don't fix well with our adjective patterns, perhaps we should leave `is-min`, `is-inner-borders`, `is-header-background` out of this PR, and leave them in docs-ui. None of the names quite fit in terms of their language or their clarity.

css/src/mixins/media-queries.scss
wibjorn: How about `discouraged-mobile`? Open to other ideas too. I just want folks to know they need to thing twice before they use this mixin (and that they can't easily find it unless they look in here). I think we could also use a comment explaining why it's named the way it is.

css/src/components/table.scss
```diff
@@ -0,0 +1,128 @@
...
+		padding: $table-cell-padding-xxs $table-cell-padding-xs;
+		border-top: $table-border-width solid $border;
+		border-right: 0;
+		border-bottom: 0;
+		border-left: 0;
+		border-style: solid;
+		line-height: 1.5;
+		vertical-align: top;
+
+		&:not([align]) {
```
wibjorn, line 37: seems like this still contains a lot of legacy classes in here

site/src/components/table.md
```diff
@@ -0,0 +1,256 @@
...
+
+These classes are available to style tables in different ways, feel free to mix and match them!
+
+| Class                      | Usage                                               |
+| -------------------------- | --------------------------------------------------- |
+| `is-striped`               | Adds background color to every odd row              |
+| `is-inner-borders`         | Adds inner borders to the table                     |
+| `is-header-background`     | Adds background color to the header row             |
+| `is-row-header-background` | Adds background color to the first cell of each row |
+| `is-stacked-mobile`        | Stacks table cells on mobile screen sizes.          |
```
wibjorn, line 64: Let's reduce this list to only `is-stacked-mobile`.

jdanyow: table-stacked-mobile

css/src/mixins/media-queries.scss
jdanyow: Not needed? Could we use mobile-first css for table-responsive?

css/src/components/table.scss
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	th {
+		&:not([scope='row']) {
+			border-top: 0;
+			border-bottom: $table-border-width;
+		}
+	}
+
+	&.is-striped {
```
jdanyow, line 57: table-striped?

css/src/components/table.scss
```diff
@@ -0,0 +1,128 @@
...
+	td,
+	th {
+		display: table-cell;
+		word-wrap: break-word;
+		padding: $table-cell-padding-xxs $table-cell-padding-xs;
+		border-top: $table-border-width solid $border;
+		border-right: 0;
+		border-bottom: 0;
+		border-left: 0;
+		border-style: solid;
```
jdanyow, line 33: border-style redundant? Could we remove some of the border-right/bottom/left stuff and replace with border collapse?
https://developer.mozilla.org/en-US/docs/Web/CSS/border-collapse

css/src/components/table.scss
```diff
@@ -0,0 +1,128 @@
...
+	caption {
+		padding: $font-size-6;
+		border-width: 0 0 $table-border-width;
+		border-style: solid;
+		border-color: $border;
+		font-size: $table-caption-size;
+		font-weight: $weight-semibold;
+		letter-spacing: $table-caption-spacing;
+		text-align: $user-right;
+		text-transform: uppercase;
```
jdanyow, line 21: double checking whether this transform should happen by default or instead use a utility

css/src/components/table.scss
```diff
@@ -0,0 +1,128 @@
...
+		}
+	}
+
+	> caption + thead,
+	> colgroup + thead,
+	> thead:first-child {
+		> tr:first-child td {
+			border-top: 0;
+		}
+	}
```
jdanyow, line 107: is this required after using border-collapse?
https://developer.mozilla.org/en-US/docs/Web/CSS/border-collapse

css/src/components/table.scss
ollips: Removed everything but 'is-stacked-mobile`

css/src/components/table.scss
ollips: Cleaned up a lot of things, please revisit

site/src/components/table.md
ollips: Removed that table completely, since only one modifier left. Will add it back when we'll add more modifiers

css/src/components/table.scss
ollips: this modifier was removed for now

css/src/components/table.scss
ollips: Updated

css/src/components/table.scss
ollips: Cleaned up!

css/src/mixins/media-queries.scss
wibjorn: @jdanyow that was the going opinion for quite some time; we've come up with a few things were it's just a lot easier to target mobile. Not something we want to do generally, but there are definitely a few places were doing mobile first is a big pain. We (I) thought we could name it something that no one would guess and that might do a double take before using it.

wibjorn: An example I came across was a full width button on mobile. Pretty annoying to try and unset widths on a higher resolution.

```
.button.button-full-width-mobile {
  width: 100%;
  @tablet {
    ??? 
  }
}

File: css/src/tokens/table.scss [hunk 12]
```diff
@@ -0,0 +1,7 @@
+$table-border-width: 1px;
```
wibjorn, line 1: I think these are not really proper tokens, they're variables that should live closer to the table class itself.

css/src/atomics/table.scss
wibjorn: qualifying selector type doesn't like the pattern `a.class-name`. It's hard to see because of the nesting, but it doesn't like `td.nowrap, th.is-narrow` etc.

wibjorn: Looks like we'll probably be taking a lot of that stuff out, so perhaps we won't need to worry about it?

css/src/tokens/table.scss
ollips: Moved to the top of the `components/table.scss`

css/src/atomics/table.scss
ollips: I've got another lint error that I had to disable cause I don't get how to resolve it. Here is the structure:

```
.table {
	...

	th,
	td {
		display: table-cell;
		...

		&:not([align]) {
			text-align: $user-left;
		}
	}

	thead th {
		border-top: 0;
		border-bottom: 1px solid $border;
	}
}
```

Here is an error I'm getting:

```
Expected selector ".table thead th" to come before selector ".table th:not([align])"   no-descending-specificity
```

I tried to move `thead th { ... }` block before the `th, td { ... }`, but then I'm getting this error:

```
Expected selector ".table th" to come before selector ".table thead th"   no-descending-specificity
```

So it's a circle 😵

wibjorn:
```
	th:not([align]),
	td:not([align]) {
		text-align: $user-left;
	}
```

css/src/mixins/media-queries.scss
ollips: Renamed to `discouraged-mobile`

site/src/components/table.md
```diff
@@ -0,0 +1,125 @@
+---
+title: Table
+description: Table element and it's modifiers.
```
wibjorn, line 3: Super nit! `its`

site/src/components/table.md
```diff
@@ -0,0 +1,125 @@
...
+template: standard
+---
+
+# Table
+
+Table element helps organize content into rows and columns. `table` class adds basic styles to the tables fast and easily. Adding supported modifiers will get a custom and responsive look to the table.
```
wibjorn, line 9: The tablet element ....
The `table` class ...

site/src/components/table.md
```diff
@@ -0,0 +1,125 @@
...
+		<tr>
+			<th>Footer row</th>
+			<th>Cell</th>
+			<th>Cell</th>
+		</tr>
+	</tfoot>
+</table>
+```
+
+For cases there are quite a lot of columns in the table - the `table-wrapper` class available, that should be applied to the table's parent element, to handle the overflow case:
```
wibjorn, line 54: For case in which there are quite ....

site/src/components/table.md
```diff
@@ -0,0 +1,125 @@
...
+	</table>
+</div>
+```
+
+## Modifiers
+
+_`table-stacked-mobile`_ - stacks table cells on mobile screen sizes.
+
+```html
+<table class="table table-stacked-mobile">
```
wibjorn, line 99: nit: can we change the name is `table-responsive`?

css/src/mixins/media-queries.scss
```diff
@@ -40,3 +40,14 @@
 		@content;
 	}
 }
+
+/* 
+ * EXCEPTIONS
+ * Please be careful with using these classes. They don't follow the mobile-first approach but might be extremely useful in some use-cases.
```
wibjorn, line 46: > Please avoid using this mixin if possible. It doesn't follow our the mobile-first approach to writing CSS.

Also, move comment to directly above mixin?

site/src/components/table.md
ollips: @wibjorn  It feels to me that `table-responsive` gives the expectation that the table might somehow adjust on different screen sizes and it's a good idea to use it commonly. When the `table-stacked-mobile` makes you really think if you need it.

css/src/mixins/media-queries.scss
ollips: Updated and moved right on top of the mixin

site/src/components/table.md
ollips: updated

site/src/components/table.md
ollips: updated

site/src/components/table.md
ollips: updated

site/src/components/table.md
wibjorn: k you've convinced me :D.

css/src/components/table.scss
```diff
@@ -0,0 +1,67 @@
+$table-cell-padding-xxs: $spacer-4 !default;
```
wibjorn, line 1: `$table-cell-vertical-padding`?
`$table-cell-horizontal-padding`?

-xss and -xs make it feel like there are other options, but that isn't really the case

css/src/components/table.scss
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.table {
+	width: 100%;
+	font-size: $font-size-8;
+	border-collapse: collapse;
+	table-layout: auto;
+
+	caption {
+		padding: $spacer-6;
```
wibjorn, line 20: Should `$spacer-6` be placed into a variable too?

css/src/components/table.scss
```diff
@@ -0,0 +1,67 @@
+$table-cell-padding-xxs: $spacer-4 !default;
+$table-cell-padding-xs: $spacer-5 !default;
+
+$table-caption-size: $font-size-9 !default;
+$table-caption-spacing: 0.125rem !default;
```
wibjorn, line 5: `$table-caption-letter-spacing` - should this be referencing another letter spacing token?

We have `$letter-spacing-wide` in our typography tokens, but the value is too large. Should we introduce `$letter-spacing-medium`?

css/src/components/table.scss
ollips: Updated

css/src/components/table.scss
ollips: Updated

css/src/components/table.scss
ollips: Updated

plugins/parcel-transformer-markdown-html/lib/index.js
```diff
@@ -91,6 +91,16 @@ const markedOptions = {
 				</${element}>
 			</div>
 			`;
+	},
+	table(header, body) {
+		return `
+			<div class="markdown table-wrapper">
+				<table class="table margin-top-s">
```
wibjorn, line 98: Should we be adding `margin-top-s` here? Or should this work automatically?

ollips: Without `margin-top` the rendered table would stick to the paragraph or whatever other element is rendered before, so some spacing between is needed.

css/src/components/table.scss
```diff
@@ -0,0 +1,68 @@
...
+
+	thead th {
+		border-top: 0;
+		border-bottom: 1px solid $border;
+	}
+
+	th:not([align]),
+	td:not([align]) {
+		text-align: $user-left;
+	}
```
jdanyow, line 47: I believe this is only needed for `th:not([align])` to ensure table headings are left-aligned by default rather than centered (the "none" case below).

Some test html and markdown:

| left | center | right | none |
|:-----|:------:|------:|------|
|😊    |😊     |😊    |😊    |
|aaaaaaaaaaaaaaaaaaaaa|aaaaaaaaaaaaaaaaaaa|aaaaaaaaaaaaaaaaa|aaaaaaaaaaaaaaaaa|

```html
<table>
<thead>
<tr>
<th align="left">left</th>
<th align="center">center</th>
<th align="right">right</th>
<th>none</th>
</tr>
</thead>
<tbody>
<tr>
<td align="left">😊</td>
<td align="center">😊</td>
<td align="right">😊</td>
<td>😊</td>
</tr>
<tr>
<td align="left">aaaaaaaaaaaaaaaaaaaaa</td>
<td align="center">aaaaaaaaaaaaaaaaaaa</td>
<td align="right">aaaaaaaaaaaaaaaaa</td>
<td>aaaaaaaaaaaaaaaaa</td>
</tr>
</tbody>
</table>
```
```md
| left | center | right | none |
|:-----|:------:|------:|------|
|😊    |😊     |😊    |😊    |
|aaaaaaaaaaaaaaaaaaaaa|aaaaaaaaaaaaaaaaaaa|aaaaaaaaaaaaaaaaa|aaaaaaaaaaaaaaaaa|
```

ollips: When I was testing it last week, I believe without these rules text was aligned in table cells to the right side, which was very surprising to me, but today I noticed as well that it takes effect on `th` only. Was it some testing accident? Maybe, I'm not sure :D 

I feel like for consistency I would prefer to keep it for both `td` and `th` but probably it's not needed.

plugins/parcel-transformer-markdown-html/lib/index.js
ollips: moved `margin-top-s` next to the `table-wrapper` so not it's on a correct element

css/src/components/table.scss
ollips: Updated

plugins/parcel-transformer-markdown-html/lib/index.js
wibjorn: margin-top to the table-wrapper, so it's more clear?

wibjorn (approved):

css/src/components/table.scss
ollips: I think it should happen by default. In docs it's done that way and as base styles, designers said we're okay with keeping the same styles.

css/src/mixins/media-queries.scss
```diff
@@ -40,3 +40,10 @@
 		@content;
 	}
 }
+
+// Please avoid using this mixin if possible. It doesn't follow our the mobile-first approach to writing CSS.
```
v-mingho, line 44: nit: `the` should be removed.

ollips: Actually decided to remove the `our` instead of `the` from the sentence: "It doesn't follow **our** the mobile-first approach to writing CSS."

v-mingho (approved):

css/src/components/table.scss
```diff
@@ -0,0 +1,67 @@
+$table-cell-vertical-padding: $spacer-4 !default;
+$table-cell-horizontal-padding: $spacer-5 !default;
+
+$table-caption-padding: $spacer-6 !default;
+$table-caption-size: $font-size-9 !default;
+$table-caption-spacing: $letter-spacing-medium !default;
+
```
wibjorn, line 7: I think for migration purposes, it will be useful if we can include a feature flag here:

```

$features-button: true !default;

@if $features-button == true {
[...]
```

css/src/components/index.scss
```diff
@@ -1 +1,2 @@
 @import './markdown.scss'; // must be ordered after code-block
+@import './table.scss'; // must be ordered after code-block
```
wibjorn, line 2: remove comment

.changeset/witty-owls-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-css': minor
+---
+
+Adding basic table tokens and atomics.
```
wibjorn, line 5: `Add table component`.

ollips: Updated

css/src/components/index.scss
ollips: Updated

css/src/components/table.scss
ollips: Updated

wibjorn (approved):

v-mingho (approved):

PR #214, opened 2021-06-07
Create Atomics overview page

v-mingho:
Task: task-[431989](https://dev.azure.com/ceapex/Engineering/_boards/board/t/Fox%20Team/Stories/?workitem=431989)

Link: preview-[214](https://design.docs.microsoft.com/pulls/214/atomics/overview.html)

This PR adds an overview page for Atomics. The content of the page is imported from the Atomics readme so the two files are synced. 

Todo: do more research on a different approach `<include>` + posthtml transform md -> html 

## Testing

1. Visit preview-[214](https://design.docs.microsoft.com/pulls/214/atomics/overview.html). Compare the content with the [CSS Atomics readme file](https://github.com/microsoft/atlas-design/blob/main/css/src/atomics/README.md)

### DIFF
--- a/plugins/parcel-transformer-markdown-html/lib/index.js
+++ b/plugins/parcel-transformer-markdown-html/lib/index.js
@@ -145,7 +145,11 @@ module.exports = new Transformer({
 		const code = await asset.getCode();
 		const { body, attributes } = frontMatter(code);
 
-		const parsedCode = marked(body);
+		const parsedCode = attributes.import
+			? await options.inputFS
+					.readFile(path.join(process.cwd(), attributes.import), 'utf-8')
+					.then(content => marked(content))
+			: marked(body);
 
 		if (attributes.template) {
 			const workingDir = process.cwd();
@@ -155,7 +159,6 @@ module.exports = new Transformer({
 			const tocFilename = config.contents.tocPath
 				? path.join(workingDir, config.contents.tocPath)
 				: null;
-			let fileToImport, importFileParsedCode;
 
 			logger.verbose({
 				message: `Resolved paths:\n${workingDir}\n${templateDir}\n${templateFilename}\n${tocFilename}`,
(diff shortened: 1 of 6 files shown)


plugins/parcel-transformer-markdown-html/lib/index.js
```diff
@@ -195,14 +196,22 @@ module.exports = new Transformer({
 
 			asset.addIncludedFile(templateFilename);
 
+			if (attributes.import) {
+				fileToImport = await new Promise(resolve =>
```
v-mingho, line 200: I decided to go with the current approach to avoid requiring extra files. We're already using `marked` on line 148 for transforming markdown to HTML.  	

The same result can be achieved by using `posthtml`, but it involves installing plug-ins that use either `marked` or `markdown-it` under the hood. 

[posthtml-markdownit](https://github.com/posthtml/posthtml-markdownit).
```
const posthtml = require('posthtml')
const markdown = require('posthtml-markdownit')

importFileParsedCode = await posthtml([markdown()])
					.process(`<div class="markdown"><markdown src=${attributes.import}></markdown></div>`)
					.then(result => {
						return result.html;
					});
```

plugins/parcel-transformer-markdown-html/lib/index.js
```diff
@@ -155,6 +155,7 @@ module.exports = new Transformer({
 			const tocFilename = config.contents.tocPath
 				? path.join(workingDir, config.contents.tocPath)
 				: null;
+			let fileToImport, importFileParsedCode;
```
wibjorn, line 158: I think most of the changes in this PR (including standard.html) can be reverted in favor of this terribly hard to read fancy oneliner:

```
const parsedCode = attributes.import
            ? await options.inputFS
                    .readFile(path.join(process.cwd(), attributes.import), 'utf-8')
                    .then(contents => marked(contents))
            : marked(body);
```

site/toc.js
```diff
@@ -116,6 +117,9 @@ async function createToc(subDir) {
 
 		if (entry.isDirectory) {
 			entry.children = await createToc(srcPath);
+			entry.children.sort(
```
wibjorn, line 120: I've been thinking about this for a time, I wonder if we need to come up with some other method for sorting ...

This will work for now though 👍

plugins/parcel-transformer-markdown-html/lib/index.js
v-mingho: Indeed, updated

wibjorn (approved):

ollips (approved):

PR #217, opened 2021-06-11
add initial version of theme switch to Atlas side

wibjorn:
Task: task-436851 (buttons)

Link: preview-217

Note: button looks a tad off, because buttons haven't gone into main yet.

## Testing

1. Open preview link.
2. See theme button.
3. Click it to change the theme. Should rotate - light -> dark -> high-contrast -> light
4. Select any non light theme, refresh, it should persist. You should not see a flash.
5. Themes are persisted within local storage only.
Note: if you have prefers dark theme set, you should see it first.


ollips (dismissed):

ollips:
Not related to this PR, but I think the `atomics-filter` output HTML should be re-visited and the border should be applied in the different place:

image

Review link (check on high-contrast) - https://design.docs.microsoft.com/pulls/217/atomics/spacing.html

site/src/scaffold/scripts/theming.ts
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * SetTheme is the public way of changing the visual theme. It ensures we use the correct EventBus instance, while allow the user of the function to remain free of such implementation detail..
+ *
+ * @param theme The ThemeType to apply to the current window.
+ */
+export function setTheme(appliedTheme: ThemeType) {
+	const previousTheme = currentTheme;
+	setGlobalThemeValue(appliedTheme);
+	setThemeClass(appliedTheme);
+	if (previousTheme === appliedTheme) {
```
v-mingho, line 102: Should this condition go right after line 99? I'm thinking that if the appliedTheme is the same as previousTheme, then there wouldn't be a need to set the current theme value and class again.

wibjorn: whoops! Yes this should be changed

wibjorn: Updated per your suggestion

ollips (approved):

v-mingho (approved):

PR #224, opened 2021-07-01
Adding CSP to site

v-joelwatson:
[Task-451986](https://dev.azure.com/ceapex/Engineering/_workitems/edit/451986)

[Preview-PR](https://design.docs.microsoft.com/pulls/224/)

Adding Content Security Policy meta tag for all pages.

## Testing

1. Follow the above link to preview the PR
2. Click through the table of contents

Expected result: 
All pages should load successfully


site/src/scaffold/standard.html
```diff
@@ -8,6 +8,7 @@
 		<meta property="og:title" content="{{title}}" />
 		<meta property="og:type" content="website" />
 		<meta name="color-scheme" content="light dark" />
+		<meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'self'; style-src 'self' 'unsafe-hashes' 'sha256-xcaKW0OyR0UBm0XPKdo2gWdgJW81zFPp1ovF7iW/x0c=' 'sha256-gyrAVcbmnI8a2UtMNEZq61n4Hd0+BbNaI7mhujFCsY0=' 'sha256-5lij/kYf1SCLu25NaZsAVsoxtnSyilNGmv0srOQE/NE='; connect-src 'self'; frame-src 'self' https://www.figma.com/">
```
wibjorn, line 11: What's the storywith the unsafehashes, sha256 hashes? Can we simplify this this a default source of self and the frame source you've included here>

v-joelwatson: There are some inline styles in the plugins/parcel-transformer-markdown-html/lib/index.js file that aren't allowed unless you allow unsafe-hashes and then give the hash of the css...Unfortunately I found out later its not well supported (safari and firefox don't have it at all). I think we will have to pull the inline style attributes out into a style tag and give it a nonce.

We can definitely simplify it by setting default source to self. We only used 'none' because we were trying to make it as restritive as possible.

We could update it to 
content="default-src 'self'; style-src 'self' nonce-######; frame-src 'self' https://www.figma.com/"

wibjorn: Most of the inline styles were created before we had a good set of atomic classes to use. Perhaps they could be removed and replaced with something else. 

CC @ollips @v-mingho

v-mingho: It's mainly the min/max-height: 30vh for the code blocks/atomic classes list. We've talked about trying to avoid making new classes just for that as a last resort.

ollips: The code-block with filter causes issues, where we setting up the `min/max height` and `overflow-y` for the container. And there are no any generic classes for limiting the height at the moment in Atlas and I don't know if we should add it. That's why we've tried a different way around with hashes yesterday. I also tested locally different approach - with moving styles into the `<style>`  and assigning `nonce` to it and seems to be working as well. Let's discuss offline what we want to do

v-joelwatson: there is also some inline styles in the button.md file for color/background-color

ollips: Yes, which btw should be updated with hex values instead of words, cause it's not working on a page as it should.

asrivastava0308, line 11: I was wondering if we need to add tag for `img-src` . we do have CSS targeting `img` in `markdown.css` . For now, we are not importing image in atlas but for future .  What do you think ?

v-joelwatson: I also noticed that it looks like its blocking favicon.ico in FireFox. I think since we are changing default-src to 'self' it will set img-src to 'self' as well so we wont have to explicitly set it.

asrivastava0308 (changes requested):
Remove the `package-lock.json` file from the PR .

site/src/scaffold/code-block.scss
```diff
@@ -5,6 +5,10 @@ $code-block-header-font-size: 0.8em !default;
 .code-block {
 	position: relative;
 
+	&.code-block-min-height {
```
wibjorn, line 8: Thoughts on instead adding `.max-height-30vh`?

site/src/scaffold/code-block.scss
```diff
@@ -39,5 +43,10 @@ $code-block-header-font-size: 0.8em !default;
 
 	.code-block-body {
 		@include code-block();
+
+		&.code-bock-body-max-height {
```
wibjorn, line 47: should these two code-block specific classes be move to just some height related classes?

```
.overflow-x-scroll (this one looks like it could be included in the framework, but we can hold off on that for now)
.max-height-30vh
.min-height-30vh

```

v-joelwatson: I could see that making a lot of sense. I think we landed on the code-block.scss because the inline css we were replacing was only being placed on the code-blocks elements.

site/src/scaffold/standard.html
```diff
@@ -8,8 +8,9 @@
 		<meta property="og:title" content="{{title}}" />
 		<meta property="og:type" content="website" />
 		<meta name="color-scheme" content="light dark" />
+		<meta http-equiv="Content-Security-Policy" content="default-src 'self'; frame-src 'self' https://www.figma.com/">
 		<link rel="stylesheet" href="~/src/scaffold/index.scss">
-		<script type="module" src="~/src/scaffold/index.ts"></script>
+		<script type="module" src="~/src/scaffold/index.ts"></script>
```
asrivastava0308, line 13: nit pick : extra spacing at the end

v-joelwatson: Oops I didn't even notices that. Thanks

site/src/scaffold/code-block.scss
ollips: Let's create a new scss file in `site\src\scaffold\styles\` folder (call it `misc.scss`?) and add these three new classes into it. Just don't firget to import it in `site\src\scaffold\index.scss` as well ;)

site/src/scaffold/index.scss
```diff
@@ -1,3 +1,4 @@
 @import '@microsoft/atlas-css/src/index.scss';
 @import './styles/layout.scss';
+@import './styles/misc.scss';
```
wibjorn, line 3: Just getting some PTSD regarding a previous misc.scss file. Can we keep them specific for now at least?

@import './styles/height.scss';
@import './styles/overflow.scss';

site/src/scaffold/styles/misc.scss
```diff
@@ -0,0 +1,11 @@
+.overflow-y-scroll {
```
wibjorn, line 1: Move overflow to @import './styles/overflow.scss'?

site/src/scaffold/styles/misc.scss
```diff
@@ -0,0 +1,11 @@
...
+	overflow-y: scroll;
```
wibjorn, line 2: All single class atomic classes need `!important`.

.changeset/old-stingrays-know.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-site': minor
+---
+
+Added a CSP to both standard and token pages
```
v-mingho, line 5: nit: This sentence will show up in the change log, so it's best to add a period at the end to make it a proper sentence.

wibjorn (approved):

wibjorn:
Looking good!

asrivastava0308 (approved):

PR #230, opened 2021-08-05
Upgrade to Node 16.6.X

v-mingho:
Task: task-[463876](https://dev.azure.com/ceapex/Engineering/_workitems/edit/463876/)

Link: preview-[230](https://design.docs.microsoft.com/pulls/230)

This PR bumps Node version to 16.6.X. Due to the new version's incompatibility with `sass-extract`'s dated dependencies, I have replaced `sass-extract` with `sass-export` to extract the tokens in JSON format. 

Sass-export: https://github.com/plentycode/sass-export

## Testing

1. Visit the [preview link](https://design.docs.microsoft.com/pulls/230). Check out different pages and test all three themes to make sure everything still works as before.
2. Checkout `main` branch, run `npm run build:tokens` to generate `tokens.json`. Save a copy somewhere to compare with the new version. 
3. Checkout this branch `ming/sass-export`. Run `npm i` and  `npm run build:tokens` to generate `tokens.json`. Compare with the version from step 2. You should also be able to run `npm run serve:site` without errors.
4. Open one of the token files and remove the `sass-export-section` comments. Run `npm run build:tokens` and the console should print a warning about `sass-export-section` comments being missing.

Expected result:  The output should be identical except for color values.


css/tokens/index.js
```diff
@@ -1,30 +1,38 @@
...
-const { convertSassVariable } = require('./conversion');
 const { quicktype, InputData, jsonInputForTargetLanguage } = require('quicktype-core');
+const { exporter } = require('sass-export');
 
 createTokens();
 
 async function createTokens() {
-	const indexFile = require.resolve('../src/tokens/index.scss');
+	const options = {
+		inputFiles: [
```
v-mingho, line 10: In theory, I should be able to include only the index file and add the includePaths option with the tokens root path to resolve the imports in index.scss. However, I tried every path combination that I could think of and couldn't get it to work.

css/src/atomics/colors.scss
```diff
@@ -1,6 +1,6 @@
 // Framework colors
 
-@each $name, $color-set in $colors {
+@each $name, $color-set in $framework-colors {
```
wibjorn, line 3: I don't mind this change, but was there a particular reason for it? If not, can we revert?

css/src/tokens/features.scss
```diff
@@ -1 +1,5 @@
+/**
```
wibjorn, line 1: @ollips Did we forget to clean this one up?

css/src/atomics/colors.scss
v-mingho: `$colors: $framework-colors` was problematic because the compiled value provided by Sass-export was a string of the values.

`"$colors": "(\"primary\": (var(--theme-primary-base), var(--theme-primary-background), ...), \"secondary\": (var(--theme-secondary-base), var(--theme-secondary-background), var(--theme-secondary-dark),..."` 

I can try to parse/handle this string if we need to keep `$colors: $framework-colors`. I just thought that it'd be simpler to use `$framework-colors` directly instead of nesting it within another token.

Unlike simpler tokens, it didn't provide the map value, which made it easy to form the desired object.  

`"colors": {
	"primary": "var(--theme-primary-base), var(--theme-primary-background), ...
	"secondary": "var(--theme-secondary-base), var(--theme-secondary-background), ...
}`

css/tokens/index.js
wibjorn: I played around an encountered some of the same issues that you did. I don't see a major issue with this set up, but I'd like to avoid a few things:

1. every time a token file is added, we have to think about this build array
2. every time we add a token file, we may forget the required comments.

So what about this, we instead use [`fast-glob`](https://www.npmjs.com/package/fast-glob
) to get the inputFiles in a more programatic way?

```js
const glob = require('fast-glob');

createTokens();

async function createTokens() {
	const include = './src/tokens/*.scss';
	const ignore = ['./src/tokens/index.scss'];

	console.log({ include }); // testing only
	console.log({ ignore }); // testing only
	const filePaths = await glob(include, {
		ignore
	});

	console.log({ filePaths }); // testing only
	[...]
```

For item 2, we can deal with that as a follow up perhaps, but I think the logic would be something similar to this: 
If a file is not in the ignore array, but does not contain the required sass-export comments, we should show a warning in the console.

css/src/atomics/colors.scss
wibjorn: If memory serves (it so rarely does), I included $colors because it was the variable that made the most sense, but didn't remove framework colors because it's used in docs.

css/tokens/index.js
v-mingho: Yes I didn't like the idea of encountering the two issues either. I'll give fast-glob a try.

css/src/atomics/colors.scss
wibjorn: Update: as usually my memory can't be trusted. I think we used to mush brand and framework colors together, but it seems we're no longer doing that in docs.

wibjorn: In short, I think we can remove the line `$colors: $framework-colors`, and just rename `$framework-colors` to `$colors`.

css/src/tokens/features.scss
ollips: Yes, that should be removed. If I remember it correctly, I added the feature flag before we've decided we're not going to move that way.

As for the docs - table styles are not connected there yet.

css/src/atomics/colors.scss
v-mingho: Yes, I couldn't find where the combining was actually happening, hence the changes. We can go with the renaming. No parsing will be needed as long as the token structure is `$tokenName: value` and not `$tokenName: anotherTokenName -> an object`.

css/src/tokens/features.scss
v-mingho: Removed the file.

css/src/atomics/colors.scss
v-mingho: Renamed to `$framework-colors` to `$colors`.

css/tokens/index.js
v-mingho: The only problematic token file is `palette` since` colors` comes first and uses palette tokens. I added palette path first to the files list, then the paths from fast-glob.

jdanyow:
Just here to say this is awesome. Looking forward to dropping the node-sass dep. 👏

css/tokens/index.js
```diff
@@ -1,30 +1,36 @@
...
+	const ignore = ['./src/tokens/index.scss'];
+	const filePaths = await glob(include, { ignore });
 
-	/** @type {import('./types').SassConvertRendered }} **/
-	const rendered = await sassExtract.render({
-		file: indexFile
+	filePaths.forEach(path => {
+		try {
+			fs.readFile(path, 'utf8', function read(err, result) {
+				if (err) throw err;
```
wibjorn, line 17: small code convention nit, but keep `{ }` around throw err?

css/tokens/index.js
```diff
@@ -1,30 +1,36 @@
...
+				}
+			});
+		} catch (err) {
+			throw new Error(`Problem reading token files: ${err}`);
+		}
 	});
 
-	const collection = collectTokenSources(rendered);
+	const options = {
+		inputFiles: [require.resolve('../src/tokens/palette.scss'), ...filePaths]
```
wibjorn, line 28: After giving this a second (and third!) thought, I'm thinking perhaps I steered you in the wrong direction with file-glob. It's definitely true that this globbing works fine, but if need to respect import order, perhaps we should provide a more robust solution to do so always (not just for the one where we need to now). I had a little tinker with this and it seemed to produce the right filePaths in an array and respect import order:

```js
const fs = require('fs-extra');
const path = require('path');

async function getSortOrder(indexPath, filePathStem) {
	try {
		const indexFile = (await fs.readFile(indexPath)).toString();
		const lines = indexFile.split('\n').reduce((arr, line) => {
			console.log(line);  // ⚠ remove me
			if (line.includes('@import')) {
				const filePath = line.replace('@import', '').replace(`'`, '').trim();
				console.log({ filePath });   // ⚠ remove me
				arr.push(path.join(filePathStem, filePath));
			}
			return arr;
		}, []);
	} catch (err) {
		throw err;
	}
}
const filePathStem = path.join(process.cwd(), './css/src/tokens'); // ⚠ this might need changing
const indexPath = path.resolve(filePathStem, 'index.scss'); // ⚠ this might need changing
getSortOrder(indexPath, filePathStem);

```

v-mingho: Replaced the previous approach with the function above. Also added a sort function and refactored the comment check into its own function.

css/tokens/index.js
v-mingho: Added curly braces.

wibjorn:
So close! This looks awesome. I did notice this though, looks like booleans aren't being treated quite the same way. Do we have any easy ways to mitigate this? I think it would be ideal to treat them as bools rather than strings.

image

wibjorn:
This seems fine to me, only noting that the shape is quite different for colors. I like it though, particularly after (just) noticing that the hex values for non-opaque colors were wrong.

New:
```scss
      "$palette-black-opacity-10": "rgba(0, 0, 0, 0.1)",
      "$palette-black-opacity-0": "rgba(0, 0, 0, 0)",
      "$palette-white": "#fff",
```

Old: 

```scss
      "$palette-black": { "r": 0, "g": 0, "b": 0, "a": 1, "hex": "#000000" },
      "$palette-black-opacity-90": {
        "r": 0,
        "g": 0,
        "b": 0,
        "a": 0.9,
        "hex": "#000000"
      },
      "$palette-black-opacity-80": {
        "r": 0,
        "g": 0,
        "b": 0,
        "a": 0.8,
        "hex": "#000000"
      },
      "$palette-black-opacity-70": {
        "r": 0,
        "g": 0,
        "b": 0,
        "a": 0.7,
        "hex": "#000000"
      },
```

v-mingho:
> So close! This looks awesome. I did notice this though, looks like booleans aren't being treated quite the same way. Do we have any easy ways to mitigate this? I think it would be ideal to treat them as bools rather than strings.
> 
> Left is updated, right is current (0.13.0).
> 
> image

I noticed that as well. Unfortunately, `sass-export` doesn't provide Sass rules like `sass-extract` does, so all values are treated as strings. In theory, we can check each value to see if it matches `"true"` or `"false"` and convert to a boolean.

wibjorn:
Can we coerce "true" and "false" to their boolean counterparts?

css/tokens/index.js
```diff
@@ -34,39 +28,127 @@ async function createTokens() {
...
+	}
 	return collection;
 }
 
+/**
+ *
+ * @param {import('./types').SassExportTokenItem} child
+ * @returns {{[key: string]: import('./types').SassExportTokenNestedItem}}
+ */
+function getNestedTokens(child) {
```
wibjorn, line 139: Placing comment here only because I'm not sure where else to do so. Seeing some extra spaces in sass arrays. Is that us or sass-export?

image

css/tokens/index.js
```diff
@@ -34,39 +28,127 @@ async function createTokens() {
...
+		return { [name]: compiledValue ? compiledValue : value };
+	}
+	/** @type {{[key: string]: import('./types').SassExportTokenNestedItem}} */
+	const childMap = {};
+	child.mapValue.forEach(subChild => {
+		childMap[child.name] = { ...childMap[child.name], ...getNestedTokens(subChild) };
+	});
+	return childMap;
+}
+
```
wibjorn, line 151: Unfortunately, after looking at themes, this update will need some fairly major changes. I think we need the values here, rather only the variable names.

image

v-mingho: Add a section to parse compiledValue and retrieve the values.  
image

css/tokens/index.js
v-mingho: It's sass-export, add logic to remove extra spacing.

css/tokens/index.js
v-mingho: Also coerced booleans 
image

v-joelwatson (dismissed):

css/tokens/index.js
```diff
@@ -34,39 +28,165 @@ async function createTokens() {
...
+ * @returns {{[key: string]: string | boolean | import('./types').SassExportTokenNestedItem}}
+ */
+function getNestedTokens(child, parent) {
+	if (!child.mapValue) {
+		const { name, compiledValue, value } = child;
+		let newCompiledValue;
+
+		//For nested maps, the compiled value for each key is retrieved by parsing the parent's compiledValue string.
+		if (!compiledValue) {
+			newCompiledValue = parent?.compiledValue
```
wibjorn, line 152: Parsing the parent here, in the child, means that we're parsing the exact same thing many times over.

Each time a child property on `light` theme is parsed, the whole parent is parsed again. Should we clean this up to ensure we're not doing duplicate work?

wibjorn: In fact, it looks to me like traversing all the child nodes may not be entirely necessary. The following riff on your code smacks them all into an object very similar (the same?) to the one being output:

```js
const parsedParent = parent?.compiledValue
.replace(/^\(/, '')
.replace(/\)$/, '')
.split(/\,(?![^(]*\))/)
.reduce((collection, val) => {
	const [name, value] = val.replaceAll(' ', '').replaceAll('"', '').split(':');
	collection[name] = convertBoolean(value);
	return collection;
}, {});

Comment (wibjorn) [reply css/tokens/index.js:152]: values of `parsedParent` is 

```js
{
  text: "#171717",
  "text-subtle": "#505050",
  "text-invert": "#fff",
  "text-glow-high-contrast": "#171717",
  "box-shadow-light": "rgba(0,0,0,0.11)",
  "box-shadow-medium": "rgba(0,0,0,0.13)",
  "box-shadow-heavy": "rgba(0,0,0,0.18)",
  "box-shadow-extra-heavy": "rgba(0,0,0,0.22)",
  overlay: "rgba(0,0,0,0.7)",
  "overlay-invert": "#fff",
  "body-background": "#fff",
  "body-background-medium": "#f2f2f2",
  "alternate-background": "#171717",
  "alternate-background-medium": "#2f2f2f",
  border: "#e6e6e6",
  "table-header": "#e6e6e6",
  "table-row-header": "#fafafa",
  "table-border-dark": "#d2d2d2",
  "hover-base": "#2f2f2f",
  "hover-invert": "#fafafa",
  "border-white-high-contrast": "rgba(255,255,255,0)",
  "border-yellow-high-contrast": "rgba(0,0,0,0)",
  "code-header": "#e6e6e6",
  "code-block": "#f2f2f2",
  "inline-code": "#e6e6e6",
  "code-highlight-background": "#fff4ce",
  visited: "#624991",
  "score-low-off": "rgba(168,0,0,0.3)",
  "score-low": "#b62626",
  "score-medium-off": "rgba(255,187,0,0.3)",
  "score-medium": "#ffb900",
  "score-high-off": "rgba(72,157,72,0.3)",
  "score-high": "#2a8b2a",
  hyperlink: "#0065b3",
  "primary-base": "#0078d4",
  "primary-background": "#d7eaf8",
  "primary-background-glow-high-contrast": "#d7eaf8",
  "primary-dark": "#004173",
  "primary-hover": "#0065b3",
  "primary-active": "#00579a",
  "primary-box-shadow": "rgba(0,101,179,0.3)",
  "primary-invert": "#fff",
  "secondary-base": "#a2a2a2",
  "secondary-background": "#e6e6e6",
  "secondary-background-glow-high-contrast": "#e6e6e6",
  "secondary-dark": "#2f2f2f",
  "secondary-hover": "#d2d2d2",
  "secondary-active": "#d2d2d2",
  "secondary-box-shadow": "rgba(0,0,0,0.3)",
  "secondary-invert": "#000",
  "tertiary-base": "#243a5e",
  "tertiary-background": "#061329",
  "tertiary-background-glow-high-contrast": "#061329",
  "tertiary-dark": "#a4b2c9",
  "tertiary-hover": "#14294c",
  "tertiary-active": "#70819f",
  "tertiary-box-shadow": "rgba(38,23,63,0.3)",
  "tertiary-invert": "#fff",
  "success-base": "#107c10",
  "success-background": "#dff6dd",
  "success-background-glow-high-contrast": "#dff6dd",
  "success-dark": "#054b16",
  "success-hover": "#0b6413",
  "success-active": "#054b16",
  "success-box-shadow": "rgba(72,157,72,0.3)",
  "success-invert": "#fff",
  "info-base": "#8661c5",
  "info-background": "#efd9fd",
  "info-background-glow-high-contrast": "#efd9fd",
  "info-dark": "#3b2e58",
  "info-hover": "#8661c5",
  "info-active": "#624991",
  "info-box-shadow": "rgba(32,24,67,0.3)",
  "info-invert": "#fff",
  "warning-base": "#ffb900",
  "warning-background": "#fff4ce",
  "warning-background-glow-high-contrast": "#fff4ce",
  "warning-dark": "#6a4b16",
  "warning-hover": "#d19501",
  "warning-active": "#966802",
  "warning-box-shadow": "rgba(255,187,0,0.3)",
  "warning-invert": "#000",
  "danger-base": "#a80000",
  "danger-background": "#fde7e9",
  "danger-background-glow-high-contrast": "#fde7e9",
  "danger-dark": "#470001",
  "danger-hover": "#870000",
  "danger-active": "#630001",
  "danger-box-shadow": "rgba(168,0,0,0.3)",
  "danger-invert": "#fff",
  "facepile-red": "#a80000",
  "facepile-teal": "#328e9f",
  "facepile-blue": "#0078d4",
}

```

wibjorn (changes requested):

css/tokens/index.js
```diff
@@ -34,39 +28,165 @@ async function createTokens() {
...
+	} catch (err) {
+		throw err;
+	}
+}
+
+/**
+ *
+ * @param {string[]} paths token file paths
+ */
+// Print a warning if a token file lacks the required sass-export-section comments.
```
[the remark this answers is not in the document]
v-mingho: Updated

css/tokens/index.js
```diff
@@ -1,30 +1,24 @@
...
 
-	/** @type {import('./types').SassConvertRendered }} **/
-	const rendered = await sassExtract.render({
-		file: indexFile
-	});
+	const filePaths = await getInputFilesFromIndex(filePathStem, indexPath);
+	checkFileComments(filePaths);
 
-	const collection = collectTokenSources(rendered);
+	const options = {
```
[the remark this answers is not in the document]
v-mingho: How about `/** @type {{inputFiles: string[]}} */`?

css/tokens/index.js
```diff
@@ -34,39 +28,165 @@ async function createTokens() {
...
+	return childMap;
+}
+
+/**
+ *
+ * @param {string} string
+ * @returns {boolean | string}
+ */
+function convertBoolean(string) {
+	if (string === 'true') return true;
```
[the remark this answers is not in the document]
v-mingho: updated

css/tokens/index.js
```diff
@@ -34,39 +28,165 @@ async function createTokens() {
...
+	if (string === 'false') return false;
```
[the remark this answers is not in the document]
v-mingho: updated

css/tokens/index.js
```diff
@@ -34,39 +28,165 @@ async function createTokens() {
...
+	});
+	return childMap;
+}
+
+/**
+ *
+ * @param {string} string
+ * @returns {boolean | string}
+ */
+function convertBoolean(string) {
```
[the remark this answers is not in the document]
v-mingho: updated

css/tokens/index.js
```diff
@@ -34,39 +28,165 @@ async function createTokens() {
...
+
+/**
+ *
+ * @param {string[]} paths token file paths
+ */
+// Print a warning if a token file lacks the required sass-export-section comments.
+function checkFileComments(paths) {
+	paths.forEach(path => {
+		try {
+			fs.readFile(path, 'utf8', function read(err, result) {
```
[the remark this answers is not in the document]
v-mingho: updated

css/tokens/index.js
v-mingho: Pushed the latest version that doesn't require passing parent.

css/tokens/index.js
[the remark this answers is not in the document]
wibjorn: That will do in a pinch. The reason I was asking was because it would be awesome for developers have a better understanding of what the full list of options are - so a more full type would be good. We can see if we can get to that later.

.changeset/twelve-cheetahs-hunt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-css': patch
```
wibjorn, line 2: Just realized this is a breaking change. Major upgrade because of the change in our token shape.

wibjorn: Updated

```json
"text": "#171717",
```

Previous

```json
"text": { "r": 23, "g": 23, "b": 23, "a": 1, "hex": "#171717" },
```

css/tokens/index.js
```diff
@@ -129,64 +126,93 @@ function collectTokens(tokens) {
...
+ * @returns {import('./types').SassExportNestedCollection}
+ */
+function getNestedTokens(tokenItem) {
+	/* NOTE: sass-export seems to have trouble parsing maps that have more two levels. */
+	const { name, compiledValue, mapValue } = tokenItem;
+
+	/** @type {import('./types').SassExportNestedCollection} */
+	let newCompiledValue = {};
+
+	let nested = containsMapValue(tokenItem);
```
wibjorn, line 146: const

css/tokens/index.js
```diff
@@ -129,64 +126,93 @@ function collectTokens(tokens) {
...
+ * Aggregate nested values when token value is a Sass map
+ * @param {import('./types').SassExportTokenItem} tokenItem
+ * @returns {import('./types').SassExportNestedCollection}
+ */
+function getNestedTokens(tokenItem) {
+	/* NOTE: sass-export seems to have trouble parsing maps that have more two levels. */
+	const { name, compiledValue, mapValue } = tokenItem;
+
+	/** @type {import('./types').SassExportNestedCollection} */
+	let newCompiledValue = {};
```
wibjorn, line 144: const

css/tokens/index.js
```diff
@@ -129,64 +126,93 @@ function collectTokens(tokens) {
...
 
 	return collection;
 }
 
 /**
- *
- * @param {import('./types').SassExportTokenItem} child
- * @param {import('./types').SassExportTokenItem} [parent]
- * @returns {{[key: string]: string | boolean | import('./types').SassExportTokenNestedItem}}
+ * Aggregate nested values when token value is a Sass map
```
wibjorn, line 135: Second level of comment: 
```
* Note, this currently only targets to levels of nesting.
```

css/tokens/index.js
```diff
@@ -129,64 +126,93 @@ function collectTokens(tokens) {
...
+	if (mapValue && nested) {
+		for (const child of mapValue) {
+			parseCompiledValue(child.compiledValue, child.name, newCompiledValue);
+		}
+	}
+	return { [name]: nested ? newCompiledValue : newCompiledValue[name] };
+}
+
+/**
+ * Return a boolean if the token item has nested maps
```
wibjorn, line 161: Second level of comment: 
```
* Note, this currently only targets to levels of nesting.
```

css/tokens/index.js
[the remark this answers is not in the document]
v-mingho: I think it would be most prudent to test all the options before we add the available options (since we found out that inputPaths doesn't actually work).

.changeset/twelve-cheetahs-hunt.md
v-mingho: Updated to major

css/tokens/index.js
v-mingho: updated

css/tokens/index.js
v-mingho: updated

css/tokens/index.js
v-mingho: updated

css/tokens/index.js
v-mingho: updated

css/tokens/index.js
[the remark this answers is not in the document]
wibjorn: I'm not sure what the harm would be in having a more accurate type here, but I agree that it isn't a priority now.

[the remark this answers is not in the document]
v-mingho: Added a new type taken from Sass-export.

wibjorn (approved):

v-joelwatson (approved):

asrivastava0308 (approved):

PR #254, opened 2021-09-27
add overflow atomics

wibjorn:
Task: task-488692

Link: preview-254 
Click "Overflow" in navigation, under "Atomics". Visit page.


site/src/atomics/overflow.md
```diff
@@ -0,0 +1,52 @@
...
+template: standard
+---
+
+# Overflow
+
+At times, you'll need to determine the overflow behavior of an element. Atlas provides several classes to do this.
+
+| cssproperty                            | value              | screensize |
+| -------------------------------------- | ------------------ | ---------- |
+| `overflow`, `overflow-x`, `overflow-y` | `scroll`, `hidden` | N/A        |
```
ollips, line 13: What about the `auto` value? I think in some cases it might be useful as well, for example, for tables.

css/src/atomics/overflow.scss
```diff
@@ -0,0 +1,15 @@
+$overflow-values: hidden, scroll;
```
jdanyow, line 1: 1. Do we need utilities for `scroll`? 
2. Do we not need utilities for `visible`?
3. I think some of these need `-webkit-overflow-scrolling: touch`
4. Would be good to reconcile with `is-vertically-scrollable`

jdanyow (changes requested):

css/src/atomics/overflow.scss
```diff
@@ -0,0 +1,29 @@
+$overflow-values: auto hidden;
+
+@each $value in $overflow-values {
```
jdanyow, line 3: Is the looping needed? Thinking we'd might only need overflow-hidden. The classes defined in scroll.scss are what we'd want to direct developers towards.

css/src/components/scroll.scss
```diff
@@ -0,0 +1,11 @@
+.scroll-horizontal {
```
ollips, line 1: Should we have a doc page presenting these classes? I think might be useful, otherwise, people won't know about their existence

css/src/atomics/overflow.scss
wibjorn: You are right, just updated

site/src/atomics/overflow.md
```diff
@@ -0,0 +1,55 @@
...
+	</div>
+</div>
+```
+
+### Scolling
+
+You can force the appearance of either horizontal or vertical scollbars with `.overflow-x-auto` and `.overflow-y-auto`.
+
+```html
+<div class="overflow-x-auto">
```
jdanyow, line 34: The result of this would be the browser defaulting overflow-y to auto, and there would be no momentum scrolling. 

Thinking these classes should be removed:

overflow-auto
.overflow-x-hidden
.overflow-x-auto
.overflow-y-hidden
.overflow-y-auto

site/src/atomics/overflow.md
```diff
@@ -0,0 +1,55 @@
...
+		</p>
+	</div>
+</div>
+```
+
+## Available classes
+
+List of all available classes:
+
+```atomics-filter
```
ollips, line 33: Didn't we agree to list the available classes if there are 20+ of them? Otherwise, it's quite a lot of empty space at the bottom of the page. Also, if the Figma component would be presented at the bottom as well that would look really odd

css/src/atomics/overflow.scss
```diff
@@ -0,0 +1,25 @@
...
+	}
+
+	.overflow-x-hidden-tablet {
+		overflow-x: $value;
+	}
+
+	.overflow-y-hidden-tablet {
+		overflow-y: $value;
+	}
+}
```
jdanyow, line 25: $value -> hidden

site/src/atomics/overflow.md
```diff
@@ -0,0 +1,55 @@
...
+
+```html
+<div class="overflow-x-auto">
+	<div class="background-color-success padding-xl">
+		<p class="color-success-invert font-size-xl">
+			This container has a horizontal scrollbar.
+			----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------->
+		</p>
+	</div>
+</div>
```
ollips: I don't see a scroll here:

image

wibjorn: yeah gotta remove that example, good catch!

site/src/atomics/overflow.md
```diff
@@ -0,0 +1,40 @@
...
+| cssproperty                            | value    | screensize |
+| -------------------------------------- | -------- | ---------- |
+| `overflow`, `overflow-x`, `overflow-y` | `hidden` | N/A        |
+
+## Usage
+
+As an examplem we'll use a parent element's border radius in combination with `.overflow-hidden` to ensure a child element's background is clipped. This is primarily useful with images nested deeploy in content, but here we'll use a background color for ease.
+
+```html
+<div class="overflow-hidden border-radius-l">
```
ollips, line 20: `border-radius-l` -> `border-radius-lg`

site/src/atomics/overflow.md
```diff
@@ -0,0 +1,40 @@
...
+
+At times, you'll need to determine the overflow behavior of an element. Atlas provides several classes to do this.
+
+| cssproperty                            | value    | screensize |
+| -------------------------------------- | -------- | ---------- |
+| `overflow`, `overflow-x`, `overflow-y` | `hidden` | N/A        |
+
+## Usage
+
+As an examplem we'll use a parent element's border radius in combination with `.overflow-hidden` to ensure a child element's background is clipped. This is primarily useful with images nested deeploy in content, but here we'll use a background color for ease.
```
ollips, line 17: nit: "As an exampl**e** ..."

ollips, line 17: nit: " ... with images **deeply** nested ... "

css/src/atomics/overflow.scss
wibjorn: fixed

css/src/components/scroll.scss
wibjorn: updated and added 👍

site/src/atomics/overflow.md
wibjorn: removed!

site/src/atomics/overflow.md
wibjorn: that example was moved to the /scroll page

site/src/atomics/overflow.md
wibjorn: fixed!

wibjorn: updated!

site/src/atomics/overflow.md
wibjorn: updated

site/src/components/scroll.md
```diff
@@ -0,0 +1,70 @@
...
+		</p>
+	</div>
+</div>
+
+## Vertical scrolling
+
+Force an elements overflow to be horizontally scrollable by using `.scroll-vertical`. Note that this will only occur if the elements height is constrained.
+
+<div class="scroll-vertical example background-color-success max-height-30vh margin-top-md">
+	<p class="color-success-invert font-size-h1 margin-bottom">
```
ollips, line 29: Can we adjust the second example's styles to match the first one, so visually it looks better together? Update the font size to be `font-size-xl` and add `padding-xl` to the parent `div`

site/src/components/scroll.md
```diff
@@ -0,0 +1,70 @@
+---
```
ollips, line 1: Also, should it be in the components folder? I feel like the component is something complex, with additional helpers created specifically for it 🤔 I think Atomics is a better place, but I'm not 100% sure

ollips: And I realized there are no more `overflow-x-hidden` nor overflow-y-hidden` helpers, I assume you removed these two assuming people would use` scroll-horizontal` \ `scroll-vertical` instead, so I think it would be good to keep that info close to each other.

wibjorn: Yep, I think these are both components. Agreed that they are simple, but components need not be complex. Atomics are tied whenever possible to a single CSS property (or maybe two if we're being a little forgiving). That properties name is reflected by CSS. The name alone here is a clue that these aren't Atomics as we've defined them until now.

site/src/components/scroll.md
wibjorn: 👍Good call! Updated!

jdanyow (approved):

site/src/components/scroll.md
```diff
@@ -0,0 +1,70 @@
...
+
+# Scrolling
+
+There are two components that help modify the scroll direction of a particular container.
+
+## Horizontal scrolling
+
+Force an elements overflow to be horizontally scrollable by using `.scroll-horizontal`.
+
+<div class="scroll-horizontal border background-color-success padding-xl margin-top-md">
```
ollips, line 15: nit: `border` is needed?

site/src/components/scroll.md
```diff
@@ -0,0 +1,70 @@
...
+	<div class="">
```
ollips, line 16: Is it also needed? Sorry missed those initially 🤦🏻‍♀️

site/src/components/scroll.md
```diff
@@ -0,0 +1,70 @@
...
+		|
+		|
+		|
+		|
+		|
+		|
+		▼
+		</pre>
+	</p>
+
```
ollips, line 69: super nit: remove an extra empty line

site/src/components/scroll.md
```diff
@@ -0,0 +1,70 @@
...
+			<pre class="color-success-invert ">-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------></pre>
+		</p>
+	</div>
+</div>
+
+## Vertical scrolling
+
+Force an elements overflow to be horizontally scrollable by using `.scroll-vertical`. Note that this will only occur if the elements height is constrained.
+
+<div class="scroll-vertical max-height-30vh example background-color-success margin-top-md padding-xl">
```
ollips, line 28: Remove `example` class, doesn't seem to be needed here

site/src/components/scroll.md
wibjorn: updated!

site/src/components/scroll.md
wibjorn: updated!

site/src/components/scroll.md
wibjorn: updated!

site/src/components/scroll.md
wibjorn: updated!

ollips (approved):

jdanyow (approved):

PR #256, opened 2021-10-01
adaptive button with .button.button-clear.border

wibjorn:
Task: task-495356

Link: preview-256

## Testing

1. Step one
2. Step two
   Expected result:

## Additional information

[Optional]

### DIFF
--- a/site/src/components/button.md
+++ b/site/src/components/button.md
@@ -147,14 +147,18 @@ Ensure the user knows they need to wait for some event (like a fetch request) in
 <button class="button button-filled button-warning is-loading">Warning</button>
 <button class="button button-filled button-info is-loading">Info</button>
 <button class="button button-filled button-danger is-loading">Danger</button>
+<button class="button button-clear border is-loading">Clear</button>
 ```
 
 ## Adaptive buttons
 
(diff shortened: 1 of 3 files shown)


site/src/components/button.md
```diff
@@ -147,14 +147,18 @@ Ensure the user knows they need to wait for some event (like a fetch request) in
...
 <button class="button button-filled button-danger is-loading">Danger</button>
+<button class="button button-clear border is-loading">Clear</button>
 ```
 
 ## Adaptive buttons
 
-The default clear button picks up the color the text set on a container. It is best used as an color-accessible button on a color not part of one of the Atlas themes.
+The default clear button picks up the color the text set on a container. It is best used as an color-accessible button on a color not part of one of the Atlas themes. These button do not support loading states.
+
+Note! These buttons are not design to be full featured, but rather to work when our typical themed buttons won't satisfy color requirements for non-standard background. Use with care, and _always test for a contrast ratio of greater than 4.5:1 between foreground and background_. When using them, it is recommended that you use a static text color, or to use in conjuction with a theme class to prevent accidental effects on other themes.
```
ollips, line 157: 👍👍👍

ollips (approved):

disimmon (approved):

PR #277, opened 2021-11-02
fix lint with quotes and ampersands

wibjorn:
Task: task-511585

Link: preview-277

Previous build with a lint error committed: https://github.com/microsoft/atlas-design/runs/4086544836?check_suite_focus=true

## Testing

See link above.

1. Pull locally.
2. Make a CSS change that will fail link (like putting height last in .button)
3. Run lint.
4. Should fail.

All DevOps here, so no changeset.


css/src/components/button.scss
```diff
@@ -42,7 +43,6 @@ $button-font-weight: $weight-semibold !default;
 	text-decoration: none;
 	cursor: pointer;
 	padding-#{$user-left}: $button-padding-horizontal;
```
wibjorn, line 45: Note: this must be reverted; we want this to fail a build

wibjorn: reverted

ann-liu (approved):

ollips (approved):

PR #282, opened 2021-11-10
Make the dark theme box shadow dark

ollips:
Task: task-515459

Link: preview-282

## Testing

1. Step one
2. Step two
   Expected result:

## Additional information

[Optional]


jdanyow (approved):

css/src/tokens/themes.scss
```diff
@@ -197,10 +197,10 @@ $themes: (
 		'text-subtle': $palette-white,
 		'text-invert': $palette-black,
 		'text-glow-high-contrast': $palette-yellow-high-contrast,
-		'box-shadow-light': $palette-white-opacity-0,
-		'box-shadow-medium': $palette-white-opacity-0,
-		'box-shadow-heavy': $palette-white-opacity-0,
-		'box-shadow-extra-heavy': $palette-white-opacity-0,
+		'box-shadow-light': $palette-black-opacity-0,
```
wibjorn, line 200: nit: shall we just keep it the same as the others, to indicate to any future themes that they should probably remain the same too?

```
	'box-shadow-light': $palette-black-opacity-108,
	'box-shadow-medium': $palette-black-opacity-132,
	'box-shadow-heavy': $palette-black-opacity-18,
	'box-shadow-extra-heavy': $palette-black-opacity-22,
```

ollips: I like how we currently have for the high-contrast theme values set up with 0 opacity but thinking about more possible themes - making an exception here makes more confusion than benefits. I'll update in a sec

ollips: Updated!

wibjorn (approved):

jdanyow (approved):

PR #284, opened 2021-11-11
add breadcrumbs component and relevant documentation

wibjorn:
Task: task-516176
Link: preview-284

Creates the CSS for a breadcrumb component.

https://design.docs.microsoft.com/pulls/284/components/breadcrumbs.html

## Additional information

It looks like Atlas /core needs to begin setting a visited style that is theme aware.


css/src/components/breadcrumbs.scss
```diff
@@ -0,0 +1,51 @@
...
+$breadcrumb-spacing-horizontal-start: 0.5em !default;
+$breadcrumb-spacing-horizontal-end: 0.25em !default;
+$breadcrumb-padding-vertical: 0.375em !default;
+$breadcrumb-separator: '\00200B/' !default;
+$breadcrumb-color: $hyperlink !default;
+
+.breadcrumbs {
+	font-size: $breadcrumb-font-size;
+
+	.breadcrumb-overflow {
```
jdanyow, line 11: Thinking this would not be needed if breadcrumbs was limited to list of crumbs and did not contain the overflow. You'd of course need a modifier to determine whether there should be a leading slash which would need to be added programmatically when overflow occurred at the same time as the overflow menu... thoughts?

css/src/components/breadcrumbs.scss
```diff
@@ -0,0 +1,51 @@
+$breadcrumb-font-size: $document-font-size !default;
+$breadcrumb-spacing-horizontal-start: 0.5em !default;
+$breadcrumb-spacing-horizontal-end: 0.25em !default;
+$breadcrumb-padding-vertical: 0.375em !default;
+$breadcrumb-separator: '\00200B/' !default;
```
jdanyow, line 5: I know we're doing this already, so it has passed several accessibility reviews, just curious how this reads and whether it should be using a rotated border instead. Feel free to resolve.

css/src/components/breadcrumbs.scss
```diff
@@ -0,0 +1,51 @@
...
+$breadcrumb-color: $hyperlink !default;
```
jdanyow, line 6: I think it would be more flexible / separate concerns if breadcrumbs didn't set link colors.

css/src/components/breadcrumbs.scss
jdanyow: something along the lines of this?

`.breadcrumbs`
```
bla / bla / bla /
```
`.breadcrumbs.breadcrumbs-overflowed`
```
/ bla / bla / bla /
```

css/src/components/breadcrumbs.scss
wibjorn: It seems pretty straightforward to me that breadcrumbs would ensure the link color.

css/src/components/breadcrumbs.scss
wibjorn: Thoughts are that the current version will allow us to be a little more exacting in terms of spacing on design consistency - and that `.breadcrumbs.breadcrumbs-overflowed` is an less easily understandable concept than an optional subcomponent.

css/src/components/breadcrumbs.scss
wibjorn: It reads "Slash".

css/src/components/breadcrumbs.scss
wibjorn: I'd also be open to making the `breadcrumbs-overflow` some day, so as to even include the overflow menu and details - but seems out of scope for this PR.

PR #286, opened 2021-12-02
Adding breadcrumbs component

ollips:
Task: task-516176

Link: preview-286

Creates the CSS for a breadcrumb component.

https://design.docs.microsoft.com/pulls/286/components/breadcrumbs.html

### DIFF
--- a/css/src/components/breadcrumbs.scss
+++ b/css/src/components/breadcrumbs.scss
@@ -1,0 +1,30 @@
+$breadcrumb-font-size: $font-size-8 !default;
+$breadcrumb-spacing-horizontal: 0.25em !default;
+$breadcrumb-spacing-vertical: 0.375em !default;
+$breadcrumb-separator: '\00200B/' !default;
+
+.breadcrumbs {
+	font-size: $breadcrumb-font-size;
+	list-style: none;
+
+	.breadcrumb-item {
+		display: inline;
+
+		@include desktop {
+			display: inline-block;
+			padding-block: $breadcrumb-spacing-vertical;
+		}
+
+		&::after,
+		&::before {
+			display: inline-block;
+			margin-inline: $breadcrumb-spacing-horizontal;
+			color: $text-subtle;
+		}
+
+		&:not(.breadcrumb-slashless)::after {
+			color: $text-subtle;
+			content: $breadcrumb-separator;
+		}
+	}
+}
--- a/css/src/components/breadcrumbs.scss
+++ b/css/src/components/breadcrumbs.scss
@@ -5,7 +5,6 @@ $breadcrumb-separator: '\00200B/' !default;
 
 .breadcrumbs {
 	font-size: $breadcrumb-font-size;
-	list-style: none;
 
 	.breadcrumb-item {
 		display: inline;
@@ -15,8 +14,7 @@ $breadcrumb-separator: '\00200B/' !default;
 			padding-block: $breadcrumb-spacing-vertical;
 		}
 
-		&::after,
-		&::before {
+		&::after {
 			display: inline-block;
 			margin-inline: $breadcrumb-spacing-horizontal;
 			color: $text-subtle;
--- a/css/src/components/breadcrumbs.scss
+++ b/css/src/components/breadcrumbs.scss
@@ -7,12 +7,8 @@ $breadcrumb-separator: '\00200B/' !default;
 	font-size: $breadcrumb-font-size;
 
 	.breadcrumbs-item {
-		display: inline;
-
-		@include desktop {
-			display: inline-block;
-			padding-block: $breadcrumb-spacing-vertical;
-		}
+		display: inline-block;
+		padding-block: $breadcrumb-spacing-vertical;
 
 		&:not(.breadcrumbs-item-slashless)::after {
 			display: inline-block;
--- a/css/src/components/breadcrumbs.scss
+++ b/css/src/components/breadcrumbs.scss
@@ -6,7 +6,7 @@ $breadcrumb-separator: '\00200B/' !default;
 .breadcrumbs {
 	font-size: $breadcrumb-font-size;
 
-	.breadcrumb-item {
+	.breadcrumbs-item {
 		display: inline;
 
 		@include desktop {
@@ -14,13 +14,9 @@ $breadcrumb-separator: '\00200B/' !default;
 			padding-block: $breadcrumb-spacing-vertical;
 		}
 
-		&::after {
+		&:not(.breadcrumbs-item-slashless)::after {
 			display: inline-block;
 			margin-inline: $breadcrumb-spacing-horizontal;
-			color: $text-subtle;
-		}
-
-		&:not(.breadcrumb-slashless)::after {
 			color: $text-subtle;
 			content: $breadcrumb-separator;
 		}
--- a/site/src/components/breadcrumbs.md
+++ b/site/src/components/breadcrumbs.md
@@ -12,48 +12,28 @@ Use breadcrumbs to show navigation from the current location to less specific lo
 ```html
 <nav>
 	<ol class="breadcrumbs">
-		<li class="breadcrumb-item">
+		<li class="breadcrumbs-item">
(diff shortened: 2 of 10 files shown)


css/src/components/breadcrumbs.scss
```diff
@@ -0,0 +1,30 @@
...
+	font-size: $breadcrumb-font-size;
+	list-style: none;
+
+	.breadcrumb-item {
+		display: inline;
+
+		@include desktop {
+			display: inline-block;
+			padding-block: $breadcrumb-spacing-vertical;
+		}
```
jdanyow, line 16: To me the vertical padding is external to this css component and would be something applied by a utility class as-needed.

css/src/components/breadcrumbs.scss
```diff
@@ -0,0 +1,30 @@
...
+
+		&::after,
+		&::before {
```
jdanyow, line 19: ::before not needed

css/src/components/breadcrumbs.scss
```diff
@@ -0,0 +1,30 @@
...
+			display: inline-block;
+			margin-inline: $breadcrumb-spacing-horizontal;
+			color: $text-subtle;
+		}
+
+		&:not(.breadcrumb-slashless)::after {
```
jdanyow, line 25: We could add the .breadcrumb-slashless modifier later, as-needed. In the new design every crumb is terminated with a slash.

css/src/components/breadcrumbs.scss
```diff
@@ -0,0 +1,30 @@
+$breadcrumb-font-size: $font-size-8 !default;
+$breadcrumb-spacing-horizontal: 0.25em !default;
+$breadcrumb-spacing-vertical: 0.375em !default;
+$breadcrumb-separator: '\00200B/' !default;
+
+.breadcrumbs {
+	font-size: $breadcrumb-font-size;
```
jdanyow, line 7: I think font-size should be determined externally, via utility, but I don't feel strongly about that.

css/src/components/breadcrumbs.scss
```diff
@@ -0,0 +1,30 @@
...
+	list-style: none;
```
jdanyow, line 8: Is this required? I thought we had reset list styles.

site/src/components/breadcrumbs.md
```diff
@@ -0,0 +1,116 @@
...
+		<li class="breadcrumb-item breadcrumb-slashless">
+			<a href="#">Species</a>
+		</li>
+	</ol>
+</nav>
+```
+
+## Representing the current location
+
+Note that the final item is omitted in the default example. If for some reason it must be included add `.breadcrumb-slashless` to the final item and wrap it into `span` instead of an `anchor` tag.
```
jdanyow, line 65: If we were to keep slashless around me might instead keep the current page's breadcrumb a link and use `aria-current="page"`

jdanyow (changes requested):

css/src/components/breadcrumbs.scss
ollips: 🤦🏻‍♀️

css/src/components/breadcrumbs.scss
ollips: After thinking about it for a while I still think it's good to add it now since Atlas is meant to be used by anyone. And that's a high chance people would want to include the last crumb, even though in our project it's not the case.

css/src/components/breadcrumbs.scss
ollips: If anyone who uses Atlas would want to overwrite the value they would be able to set up the `$breadcrumb-font-size` to whatever fits their need. So I think it should stay as is.

site/src/components/breadcrumbs.md
```diff
@@ -0,0 +1,116 @@
+---
+title: Breadcrumbs
+description: The Breadcrumb component in the Atlas Design System
+template: standard
+figmaEmbed: https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FNWYugObOGcAOlekKyjkEkT%2F%25E2%25AD%2590%25EF%25B8%258F-Atlas---Page-Templating%3Fnode-id%3D204%253A1220
+---
+
+# Breadcrumbs
+
+Use breadcrumbs to show navigation from the current location to less specific locations. Note that the final item is not presented.
```
homing1, line 10: Can you clarify the second sentence? In this example, the final item is "Species". Does it mean that I should not see "Species" in the breadcrumb?

ollips: I mean the current page breadcrumb is not presented. I'll update the wording

css/src/components/breadcrumbs.scss
ollips: I don't remember exactly why added it but seems to be working just fine without it. Will update!

css/src/components/breadcrumbs.scss
ollips: Added the vertical spacing for those cases when there are a lot of breadcrumbs and don't fit into one line, so when wraps - having some space would look nicer than without it:

image

site/src/components/breadcrumbs.md
ollips: I think the most common case when the last breadcrumb represents the current page is not a link, so I'd prefer to keep it in `span`. Added info about the `aria-current="page"` attribute, thanks for catching that!

site/src/components/breadcrumbs.md
ollips: Updated

css/src/components/breadcrumbs.scss
ollips: updated

css/src/components/breadcrumbs.scss
ollips: updated

css/src/components/breadcrumbs.scss
jdanyow: Makes sense! In that case it should be applied unconditionally since it's a function of the number/length of crumbs rather than viewport width which will introduce the wrapping.

css/src/components/breadcrumbs.scss
jdanyow: We need to be careful here- Atlas is *not* built such that it serves everyone's project needs. The purpose of Atlas is to serve the Docs platform, developed in the open to exhibit our patterns and practices with respect to modern css and accessibility.
That said, I do not see a major issue in including such a small extension like this. 

To clean this up, let's combine with the preceding selector and update the modifier class name to match our standards.

```scss
&::after:not(.breadcrumb-item-slashless) {
	display: inline-block;
	margin-inline: $breadcrumb-spacing-horizontal;
	color: $text-subtle;
	content: $breadcrumb-separator;
}
```

site/src/components/breadcrumbs.md
```diff
@@ -0,0 +1,116 @@
...
+			<a href="#">Genus</a>
+		</li>
+		<li class="breadcrumb-item">
+			<a href="#">Species</a>
+		</li>
+	</ol>
+</nav>
+```
+
+## Removing slashes
```
jdanyow, line 43: Let's remove this section since it duplicates "Representing the current location" which is the use-case for slashless.

css/src/components/breadcrumbs.scss
ollips: Updated!

site/src/components/breadcrumbs.md
ollips: Updated!

css/src/components/breadcrumbs.scss
ollips: Updated. Kept `inline-block` + `padding` to keep some spacing all the time. `inline` + `padding` combo doesn't do what I need and it's not worth spending more time on this, especially knowing we're going to truncate it

jdanyow (approved):

homing1 (approved):

PR #290, opened 2021-12-07
Popover component

ollips:
Task: task-516181

Link: preview-290

Adding popover component and relevant documentation.

## Testing

1. https://design.docs.microsoft.com/pulls/290/components/popover.html

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,10 +8,5 @@
 	"css.validate": false,
 	"less.validate": false,
 	"scss.validate": true,
-	"eslint.workingDirectories": [
-		{
-			"directory": "site",
-			"changeProcessCWD": true
-		}
-	]
+	"eslint.workingDirectories": [{ "directory": "site", "changeProcessCWD": true }]
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,5 +8,10 @@
 	"css.validate": false,
 	"less.validate": false,
 	"scss.validate": true,
-	"eslint.workingDirectories": [{ "directory": "site", "changeProcessCWD": true }]
+	"eslint.workingDirectories": [
+		{
+			"directory": "site",
+			"changeProcessCWD": true
+		}
+	]
 }
--- a/css/src/components/popover.scss
+++ b/css/src/components/popover.scss
@@ -1,5 +1,6 @@
 $popover-background-color: $body-background !default;
 $popover-border: $border-width solid $border-white-high-contrast !default;
+$popover-border-radius: $border-radius !default;
 $popover-shadow: $box-shadow-heavy !default;
 $popover-width: 224px !default;
 
@@ -8,20 +9,19 @@ $popover-width: 224px !default;
 	position: relative;
 
 	.popover-summary {
-		color: $hyperlink;
 		list-style: none;
-		cursor: pointer;
 	}
 
 	.popover-content {
 		position: absolute;
-		min-width: $popover-width;
+		width: $popover-width;
 		margin-top: $spacer-3;
 		padding: $spacer-5;
 		border: $popover-border;
-		border-radius: $border-radius;
+		border-radius: $popover-border-radius;
 		background-color: $popover-background-color;
 		box-shadow: $popover-shadow;
+		overflow: hidden;
 		z-index: $zindex-popover;
 	}
 
--- a/css/src/components/popover.scss
+++ b/css/src/components/popover.scss
@@ -8,7 +8,7 @@ $popover-width: 224px !default;
 	display: inline-block;
 	position: relative;
 
-	.popover-summary {
+	summary {
 		list-style: none;
 	}
 
--- a/css/src/components/popover.scss
+++ b/css/src/components/popover.scss
@@ -1,0 +1,36 @@
+/* stylelint-disable scss/at-extend-no-missing-placeholder */
+.popover {
+	display: inline-block;
+	position: relative;
+
+	.popover-summary {
+		@extend .button;
+
+		min-height: unset;
+	}
+
+	.popover-content {
+		position: absolute;
+		min-width: 150px;
+		margin-top: $spacer-3;
+		padding: $spacer-6;
+		border: 1px solid $border;
+		border-radius: $border-radius;
+		background-color: $body-background;
+		box-shadow: $box-shadow-heavy;
+		z-index: $zindex-popover;
+
+		&.popover-content-align-left {
+			left: 0;
+		}
+
+		&.popover-content-align-right {
+			right: 0;
+		}
+
+		&.popover-content-align-center {
+			left: 50%;
+			transform: translateX(-50%);
+		}
+	}
+}
--- a/css/src/components/popover.scss
+++ b/css/src/components/popover.scss
@@ -1,12 +1,11 @@
-/* stylelint-disable scss/at-extend-no-missing-placeholder */
 .popover {
 	display: inline-block;
 	position: relative;
 
 	.popover-summary {
-		@extend .button;
-
-		min-height: unset;
+		color: $hyperlink;
+		list-style: none;
+		cursor: pointer;
 	}
 
 	.popover-content {
@@ -19,16 +18,16 @@
 		background-color: $body-background;
 		box-shadow: $box-shadow-heavy;
 		z-index: $zindex-popover;
+	}
 
-		&.popover-content-align-left {
-			left: 0;
-		}
-
-		&.popover-content-align-right {
+	&.popover-right {
+		.popover-content {
 			right: 0;
 		}
+	}
 
-		&.popover-content-align-center {
+	&.popover-center {
+		.popover-content {
 			left: 50%;
 			transform: translateX(-50%);
 		}
--- a/css/src/components/popover.scss
+++ b/css/src/components/popover.scss
@@ -1,3 +1,8 @@
+$popover-background-color: $body-background !default;
+$popover-border: $border-width solid $border-white-high-contrast !default;
+$popover-shadow: $box-shadow-heavy !default;
+$popover-width: 224px !default;
+
 .popover {
 	display: inline-block;
 	position: relative;
@@ -10,13 +15,13 @@
 
 	.popover-content {
 		position: absolute;
-		min-width: 150px;
+		min-width: $popover-width;
 		margin-top: $spacer-3;
-		padding: $spacer-6;
-		border: 1px solid $border;
+		padding: $spacer-5;
+		border: $popover-border;
 		border-radius: $border-radius;
-		background-color: $body-background;
-		box-shadow: $box-shadow-heavy;
+		background-color: $popover-background-color;
+		box-shadow: $popover-shadow;
 		z-index: $zindex-popover;
 	}
 
--- a/site/src/components/popover.md
+++ b/site/src/components/popover.md
@@ -10,7 +10,7 @@ The popover component allows content to be placed in the popup that appears afte
 
 ```html
 <details class="popover">
-	<summary class="popover-summary link-button">Click for details</summary>
+	<summary class="link-button">Click for details</summary>
 	<div class="popover-content">
 		<p class="margin-bottom-xs">It can be a paragraph of text.</p>
 		<p class="margin-bottom-xxs">Or a list:</p>
@@ -32,11 +32,11 @@ The popover component allows content to be placed in the popup that appears afte
 
 ## Styling the trigger
 
-You can apply almost any classes in combination with the `.popover-summary` element to achieve the look you need. The following example applies `.button`.
+You can apply almost any classes on the `summary` element to achieve the look you need. The following example applies `.button`.
 
 ```html
 <details class="popover">
-	<summary class="popover-summary button border">
+	<summary class="button border">
 		<span class="icon color-primary">
 			<svg class="fill-current-color" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 2048 2048">
 				<path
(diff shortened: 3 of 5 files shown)


site/src/components/popover.md
```diff
@@ -0,0 +1,66 @@
...
+description: The Popover component in the Atlas Design System
+template: standard
+---
+
+# Popover
+
+The popover component allows both simple and more complex content to be placed in the popup that appears after clicking on a button. Popover is based on top of the [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#customizing_the_disclosure_widget) HTML component.
+
+```html
+<details class="popover breadcrumbs-overflow" data-popover-details>
```
homing1, line 12: What's the usage of the `data-popover-` attributes for?

css/src/components/popover.scss
```diff
@@ -0,0 +1,36 @@
+/* stylelint-disable scss/at-extend-no-missing-placeholder */
+.popover {
+	display: inline-block;
+	position: relative;
+
+	.popover-summary {
+		@extend .button;
```
jdanyow, line 7: To ensure this component is composable in multiple ways it should not force the presentation of the summary element  to be .button. If the design calls for a button appearance folks can add the .button class to the `<summary>` element.

css/src/components/popover.scss
```diff
@@ -0,0 +1,36 @@
...
+		padding: $spacer-6;
+		border: 1px solid $border;
+		border-radius: $border-radius;
+		background-color: $body-background;
+		box-shadow: $box-shadow-heavy;
+		z-index: $zindex-popover;
+
+		&.popover-content-align-left {
+			left: 0;
+		}
```
jdanyow, line 25: is this required? Thinking left alignment happens by default even if left:0 is not set.

ollips: Oh, you're right, forgot to remove it. Initial idea was to keep it in case of the need to overwrite another rule.

site/src/components/popover.md
ollips: First of all, it was they were in the suggested template from Will. Secondly, I thought to use it for the targeting elements to add event listeners (which I'm going to add in docs). But I can do that by pulling the class name. Will do some clean up

wibjorn: If I recall our conversation from a few weeks ago (high likelihood I'm misremembering!), I think we discussed trying to do a web component first.

wibjorn: I'm any case the main thing we need in this PR is the css. We should try and reduce extra information in examples as much as possible.

.vscode/settings.json
```diff
@@ -8,5 +8,10 @@
 	"css.validate": false,
 	"less.validate": false,
 	"scss.validate": true,
-	"eslint.workingDirectories": [{ "directory": "site", "changeProcessCWD": true }]
+	"eslint.workingDirectories": [
```
wibjorn, line 11: Revert?

site/src/components/popover.md
ollips: I don't remember the details either tbh, so I'll clean it up, and if needed - I'll add it back later

css/src/components/popover.scss
```diff
@@ -0,0 +1,36 @@
...
+
+		&.popover-content-align-right {
```
wibjorn, line 27: I personally prefer top level classes for this type of thing. Taking the `.buttons` component as an example: `.buttons.buttons-right` is more pleasant for the developer. It will require some changing of the selectors but could instead do something like:

```scss
.popover {
  &.popover-right {
    right: 0;
  }
}
```

site/src/components/popover.md
```diff
@@ -0,0 +1,66 @@
+---
+title: Popover
+description: The Popover component in the Atlas Design System
+template: standard
+---
+
+# Popover
+
+The popover component allows both simple and more complex content to be placed in the popup that appears after clicking on a button. Popover is based on top of the [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#customizing_the_disclosure_widget) HTML component.
```
wibjorn, line 9: Suggestion only. remove a few words: ~~both simple and complex~~

site/src/components/popover.md
wibjorn, line 12: Remove breadcrumb overflow class?

wibjorn (changes requested):

css/src/components/popover.scss
ollips: I haven't seen any design specifically for the popover component outside the breadcrumbs design file, and it definitely looks like a button over there. Removing the buttons styles make it look a little off and not obvious that you should click on it, so I'm working on trying to figure out what should I do if it's not a button

css/src/components/popover.scss
ollips: I thought about that too, but the alignment should be applied on the `popover-content`, so I thought it would make more sense. But I can update to this:

```css
.popover {
  ...

  &.popover-right {
    .popover-content {
      right: 0;
    }
  }
}
```

site/src/components/popover.md
ollips: 🤦🏻‍♀️👍

ollips: Cleaned up!

css/src/components/popover.scss
ollips: removed that one

.vscode/settings.json
ollips: reverted

css/src/components/popover.scss
ollips: Updated

site/src/components/popover.md
ollips: Updated

site/src/components/popover.md
ollips: Cleaned up

css/src/components/popover.scss
ollips: This one is in discussion with the design team, for now, I've updated the summary a little so it looks like a clickable element, and in combination with the `.button` class, we can get the button look.

Here is a popover design discussion: https://github.com/microsoft/atlas-design/discussions/165

css/src/components/popover.scss
wibjorn: Yes, perfect!

css/src/components/popover.scss
```diff
@@ -0,0 +1,35 @@
...
+		list-style: none;
+		cursor: pointer;
+	}
+
+	.popover-content {
+		position: absolute;
+		min-width: 150px;
+		margin-top: $spacer-3;
+		padding: $spacer-6;
+		border: 1px solid $border;
```
wibjorn, line 16: Figma designs don't have a border as far as I can tell.

css/src/components/popover.scss
```diff
@@ -0,0 +1,35 @@
+.popover {
+	display: inline-block;
+	position: relative;
+
+	.popover-summary {
+		color: $hyperlink;
```
wibjorn, line 6: Remove `color`? If we're going for composability here, we don't need it.

site/src/components/popover.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Popover
+description: The Popover component in the Atlas Design System
+template: standard
+---
+
+# Popover
+
+The popover component allows content to be placed in the popup that appears after clicking on a button. Popover is based on top of the [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#customizing_the_disclosure_widget) HTML component.
```
wibjorn, line 9: > Popover is built on top of the `<details>` ....

css/src/components/popover.scss
```diff
@@ -0,0 +1,35 @@
...
+		list-style: none;
+		cursor: pointer;
+	}
+
+	.popover-content {
+		position: absolute;
+		min-width: 150px;
```
wibjorn, line 13: 1. extract to variable at top of file? `$popover-min-width: 160px;`
2. 160px? makes it more divisible by 4s and 8s and 16s.

site/src/components/popover.md
```diff
@@ -0,0 +1,71 @@
...
+			<li>
+				<a href="#">List item</a>
+			</li>
+		</ul>
+		<p class="margin-top-xs">Or basically anything else you can think of.</p>
+	</div>
+</details>
+```
+
+You can also use the `.button` class in combination with `.popover-summary` to get the button look.
```
wibjorn, line 33: ## Styling the trigger

You can apply almost any classes in combination with the `.popover-summary` element to achieve the look you need. The following example applies `.button`.

css/src/components/popover.scss
```diff
@@ -0,0 +1,35 @@
...
+		margin-top: $spacer-3;
+		padding: $spacer-6;
+		border: 1px solid $border;
+		border-radius: $border-radius;
```
wibjorn, line 17: I think a few of these items can be extracted to variables with `!default` at the top of the file. 

bg color, box-shadow,  border (defaulting to or the only in high contract border),

css/src/components/popover.scss
ollips: Added it for the contrast.

css/src/components/popover.scss
ollips: Well, I did that to make it clear that's element is clickable. Asked Courtny about it, by the end of the week design should be done so I'm planning to update it then.

css/src/components/popover.scss
ollips: It was a random number tbh cause I didn't have any design when was writing this. I'll update according to the file Courtny will provide and your recommendation about divisibility by 4/8/16.

css/src/components/popover.scss
wibjorn: Yes, the issue is composability though. If we don't set anything here, we're free to style it however we like with other classes.

css/src/components/popover.scss
ollips: Updated to show the border for the high-contrast theme only

site/src/components/popover.md
ollips: Updated

css/src/components/popover.scss
ollips: Updated to 224px. Matches the design and divisible by 4/8/16

site/src/components/popover.md
ollips: Updated

css/src/components/popover.scss
ollips: Updated

css/src/components/popover.scss
wibjorn: PR looks good, just waiting on this one.

css/src/components/popover.scss
wibjorn: Sorry, missed this in my previous comment, but I think border-radius as well

css/src/components/popover.scss
```diff
@@ -0,0 +1,40 @@
...
+
+	.popover-summary {
+		color: $hyperlink;
+		list-style: none;
+		cursor: pointer;
+	}
+
+	.popover-content {
+		position: absolute;
+		min-width: $popover-width;
```
wibjorn, line 18: Think we need to set a reasonable maximum too (or just `width`). See what happens if there's a long string of single line text? (We'll need to do something like this for breadcrumb overflow, for example).

image

css/src/components/popover.scss
ollips: Sounds good, I'll update the default one to link so there is an example of that view!

css/src/components/popover.scss
ollips: I don't know why the rule is `min-width` when it should be just `width`. Probably a mistake, updated! Thanks for the catch 👍 Also, added the `overflow: hidden` here.

css/src/components/popover.scss
ollips: You're right, updated!

css/src/components/popover.scss
ollips: Updated!

css/src/components/popover.scss
```diff
@@ -0,0 +1,40 @@
+$popover-background-color: $body-background !default;
+$popover-border: $border-width solid $border-white-high-contrast !default;
+$popover-border-radius: $border-radius !default;
+$popover-shadow: $box-shadow-heavy !default;
+$popover-width: 224px !default;
+
+.popover {
+	display: inline-block;
+	position: relative;
+
```
wibjorn, line 10: There's a csstricky way to dismiss the popover when the user clicks off of it:

```scss
       // line 10
	&[open] .popover-summary:after {
		content: '';
		position: fixed;
		top: 0;
		right: 0;
		left: 0;
		bottom: 0;
	}
```

css/src/components/popover.scss
```diff
@@ -0,0 +1,40 @@
...
+	.popover-summary {
```
jdanyow, line 11: I think we should remove the need for this class and use `summary`. We need to take a hard dependency on the details element's `open` attribute so requiring `popover-summary` to reset the default summary styles or get the dismiss pseudo element seems like it complicates usage unnecessarily.

css/src/components/popover.scss
wibjorn: Spoke offline, we agreed to scrap this for now, since it's more of a modal pattern.

ollips: As agreed over the teams - not going to do that.

css/src/components/popover.scss
ollips: updated

wibjorn (approved):

css/src/components/popover.scss
jdanyow: Normally I would agree- this could be done as a separate step. In this case we need this selector for the behavior to more closely match the existing Docs .dropdown component.

jdanyow (approved):

PR #291, opened 2021-12-10
Increase specificity of selector to ensure a.button doesn't underline

wibjorn:
Task: task-531035

### DIFF
--- a/css/src/components/button.scss
+++ b/css/src/components/button.scss
@@ -43,6 +43,10 @@ $button-font-weight: $weight-semibold !default;
 	padding-#{$user-left}: $button-padding-horizontal;
 	padding-#{$user-right}: $button-padding-horizontal;
 
+	&:hover {
+		text-decoration: none;
+	}
+
 	strong {
 		color: inherit;
 	}
--- a/.changeset/strange-dolphins-scream.md
+++ b/.changeset/strange-dolphins-scream.md
@@ -1,0 +1,5 @@
+---
+'@microsoft/atlas-css': patch
+---
+
+Increase specifity of selector to ensure a.button shows no underlines on hover.


ollips (approved):

css/src/components/button.scss
```diff
@@ -43,6 +43,10 @@ $button-font-weight: $weight-semibold !default;
 	padding-#{$user-left}: $button-padding-horizontal;
 	padding-#{$user-right}: $button-padding-horizontal;
 
+	&:hover {
+		text-decoration: none;
+	}
```
jdanyow, line 48: change looks good- throwing this out there as an alternative- we can decrease the specificity of the docs link style using [:where](https://developer.mozilla.org/en-US/docs/Web/CSS/:where#:~:text=Copy%20to%20Clipboard-,the%20difference%20between,-%3Awhere()%20and%20%3Ais)

https://gist.run/?id=6629a07ede8cb8a1f755cf49765f90a2

wibjorn: That works well (tested in docs). As a framework author though, I'd expect the current change rather than the assumption that downstream authors have used :where. What do you think? I'm about 60/40, but both are easy.

jdanyow: I think the generic anchor styles should move in here with as low specificity as possible

wibjorn: in that case, fine with this change, I'll abandon this PR for now

jdanyow: Elaborating on this a bit more for our review on Monday:

- I think it would be good for `a { text-decoration:none}` and `a:hover { text-decoration:underline}` to move into atlas as soon as possible.
- When these do move in, I think it would be beneficial to keep the specificity of `a:hover` the same as the `a` selector by using `a:where(:hover)`. I think this is a good practice for any of the high specificity "base / generic" selectors that target elements directly rather than via classes.
- I think this practice should be extended to the types of selectors found in the `.content` / `.markdown` classes to ensure they're composable with class-based components.
    ```diff
    - .content ul {
    + :where(.content) ul {
    ```
- All this said, I have not tested any of these, but this is generally where my heads at based on my understanding of how :where works and patterns I'm starting to see emerge.

PR #296, opened 2021-12-14
small modification to breadcrumb slash spacing

wibjorn:
Task: task-529932

Link: preview-296

Upon taking a longer look at this component, I think we need a small update to the spacing of the slashes.

### DIFF
--- a/css/src/components/breadcrumbs.scss
+++ b/css/src/components/breadcrumbs.scss
@@ -1,5 +1,5 @@
 $breadcrumb-font-size: $font-size-8 !default;
-$breadcrumb-spacing-horizontal: 0.25em !default;
+$breadcrumb-spacing-horizontal: 0.375em !default;
 $breadcrumb-spacing-vertical: 0.375em !default;
 $breadcrumb-separator: '\00200B/' !default;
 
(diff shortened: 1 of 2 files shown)


homing1 (approved):

ollips (approved):

disimmon (approved):

homing1 (approved):

css/src/components/breadcrumbs.scss
```diff
@@ -12,7 +12,7 @@ $breadcrumb-separator: '\00200B/' !default;
 
 		&:not(.breadcrumbs-item-slashless)::after {
 			display: inline-block;
-			margin-inline: $breadcrumb-spacing-horizontal;
+			margin-inline-end: $breadcrumb-spacing-horizontal;
```
ollips, line 15: Revert?

ollips (changes requested):

css/src/components/breadcrumbs.scss
wibjorn: Intention was right here, but looks like the incorrect spacing in Docs was not a symptom of any issues here in Atlas, just a hallmark of the default `white-space: normal`, which multiple white space chars into a single character. This PR doesn't need to merge after all. 😔 Apologies folks

PR #299, opened 2021-12-20
Adding event listeners to popover

ollips:
Task: task-533553

Link: preview-299

Adding event listeners to the Atlas' site.

## Testing

1. Go to [popover](https://design.docs.microsoft.com/pulls/299/components/popover.html) page
2. Try to open different popovers. Just one popover can be opened at a time
3. Trigger the popover. After clicking the Escape button popover should disappear, focus is on the details' summary
4. Trigger the popover. Click somewhere on a page. An opened popover should disappear
5. Trigger the popover. Using keyboard move focus away from the popover. Popover should close
6. Testing nesting popovers. Using the dev tools directly in the browser, paste one popover inside another. Opening the inner popover won't close the parent. Clicking inside the child popover content - won't close any of the popovers. Clicking the parent's popover content, while the inner one is open, would close the child popover

### DIFF
--- a/.changeset/grumpy-experts-smell.md
+++ b/.changeset/grumpy-experts-smell.md
@@ -2,4 +2,4 @@
 '@microsoft/atlas-site': minor
 ---
 
-Adding event listeners to popover
+Adding event listeners to popover.
--- a/.changeset/grumpy-experts-smell.md
+++ b/.changeset/grumpy-experts-smell.md
@@ -1,0 +1,5 @@
+---
+'@microsoft/atlas-site': minor
+---
+
+Adding event listeners to popover
--- a/css/src/components/popover.scss
+++ b/css/src/components/popover.scss
@@ -11,7 +11,7 @@ $popover-width: 224px !default;
 	summary {
 		list-style: none;
 
-		&::marker {
+		&::-webkit-details-marker {
 			display: none;
 		}
 	}
--- a/css/src/components/popover.scss
+++ b/css/src/components/popover.scss
@@ -25,7 +25,6 @@ $popover-width: 224px !default;
 		border-radius: $popover-border-radius;
 		background-color: $popover-background-color;
 		box-shadow: $popover-shadow;
-		overflow: hidden;
 		z-index: $zindex-popover;
 	}
 
--- a/css/src/components/popover.scss
+++ b/css/src/components/popover.scss
@@ -10,6 +10,10 @@ $popover-width: 224px !default;
 
 	summary {
 		list-style: none;
+
+		&::marker {
+			display: none;
+		}
 	}
 
 	.popover-content {
--- a/site/src/components/popover.md
+++ b/site/src/components/popover.md
@@ -53,21 +53,21 @@ You can apply almost any classes on the `summary` element to achieve the look yo
 
 ## Alignments
 
-Popover is left aligned by default, but also can be centered or right aligned relatively to the summary's width. Apply any of the corresponding classes to the `popover` element: `popover-right`, `popover-center`.
+Popover is left aligned by default, but also can be centered or right aligned relatively to the `summary`'s width. Apply any of the corresponding classes to the `popover` element: `popover-right`, `popover-center`.
 
 ```html
 <details class="popover margin-xxs">
 	<summary class="button">Popover aligned to the left</summary>
-	<div class="popover-content">Popover content.</div>
+	<div class="popover-content border">Popover content.</div>
 </details>
 
 <details class="popover popover-center margin-xxs">
 	<summary class="button">Popover centered</summary>
-	<div class="popover-content">Popover content.</div>
+	<div class="popover-content border">Popover content.</div>
 </details>
 
 <details class="popover popover-right margin-xxs">
 	<summary class="button">Popover aligned to the right</summary>
-	<div class="popover-content">Popover content.</div>
+	<div class="popover-content border">Popover content.</div>
 </details>
 ```
--- a/site/src/components/popover.md
+++ b/site/src/components/popover.md
@@ -53,7 +53,7 @@ You can apply almost any classes on the `summary` element to achieve the look yo
 
 ## Alignments
 
-Popover is left aligned by default, but also can be centered or right aligned. Apply any of the corresponding classes to the `popover` element: `popover-right`, `popover-center`.
+Popover is left aligned by default, but also can be centered or right aligned relatively to the summary's width. Apply any of the corresponding classes to the `popover` element: `popover-right`, `popover-center`.
 
 ```html
 <details class="popover margin-xxs">
--- a/site/src/scaffold/scripts/popover.ts
+++ b/site/src/scaffold/scripts/popover.ts
@@ -1,52 +1,58 @@
 export function initPopovers(container: HTMLElement) {
-	container.addEventListener('click', (event: Event) => {
-		const targetPopover =
-			(event.target instanceof Element &&
-				(event.target.closest('details.popover') as HTMLDetailsElement)) ||
-			(event.target instanceof Element &&
-				(event.target.shadowRoot?.activeElement?.closest('details.popover') as HTMLDetailsElement));
+	container.addEventListener(
+		'toggle',
+		(event: Event) => {
+			const targetPopover =
+				(event.target instanceof Element &&
+					(event.target.closest('details.popover') as HTMLDetailsElement)) ||
+				(event.target instanceof Element &&
+					(event.target.shadowRoot?.activeElement?.closest(
+						'details.popover'
+					) as HTMLDetailsElement));
 
-		if (!targetPopover || targetPopover.open) {
-			return;
-		}
-
-		const keyHandler = (event: KeyboardEvent) => {
-			if (event.key === 'Escape') {
-				closePopovers();
-			}
-		};
-
-		const checkTarget = (event: Event) => {
-			if (!(event.target instanceof Element)) {
+			if (!targetPopover || !targetPopover.open) {
 				return;
 			}
 
-			if (!targetPopover?.contains(event.target)) {
-				closePopovers();
-			}
-		};
+			const keyHandler = (event: KeyboardEvent) => {
+				if (event.key === 'Escape') {
+					closePopovers();
+				}
+			};
 
-		const blurHandler = () => {
-			if (document.activeElement?.nodeName?.toLowerCase() === 'iframe') {
-				closePopovers();
-			}
-		};
+			const checkTarget = (event: Event) => {
+				if (!(event.target instanceof Element)) {
+					return;
+				}
 
-		const closePopovers = () => {
-			container.removeEventListener('focus', checkTarget, true);
-			container.removeEventListener('click', checkTarget);
-			container.removeEventListener('touchstart', checkTarget);
-			container.removeEventListener('keydown', keyHandler);
-			window.removeEventListener('blur', blurHandler);
-			if (targetPopover?.open) {
-				targetPopover.removeAttribute('open');
-			}
-		};
+				if (!targetPopover?.contains(event.target)) {
+					closePopovers();
+				}
+			};
 
-		container.addEventListener('focus', checkTarget, true);
-		container.addEventListener('click', checkTarget);
-		container.addEventListener('touchstart', checkTarget);
-		container.addEventListener('keydown', keyHandler);
-		window.addEventListener('blur', blurHandler);
-	});
+			const blurHandler = () => {
+				if (document.activeElement?.nodeName?.toLowerCase() === 'iframe') {
+					closePopovers();
+				}
+			};
+
+			const closePopovers = () => {
+				container.removeEventListener('focus', checkTarget, true);
+				container.removeEventListener('click', checkTarget);
+				container.removeEventListener('touchstart', checkTarget);
+				container.removeEventListener('keydown', keyHandler);
+				window.removeEventListener('blur', blurHandler);
+				if (targetPopover?.open) {
+					targetPopover.removeAttribute('open');
+				}
+			};
+
+			container.addEventListener('focus', checkTarget, true);
+			container.addEventListener('click', checkTarget);
+			container.addEventListener('touchstart', checkTarget);
+			container.addEventListener('keydown', keyHandler);
+			window.addEventListener('blur', blurHandler);
+		},
+		true
+	);
 }
--- a/site/src/scaffold/scripts/popover.ts
+++ b/site/src/scaffold/scripts/popover.ts
@@ -1,13 +1,5 @@
 export function initPopovers(container: HTMLElement) {
 	container.addEventListener('click', (event: Event) => {
-		const allPopovers: HTMLDetailsElement[] = Array.from(
-			container.querySelectorAll('details.popover')
-		);
-
-		if (allPopovers.length === 0) {
-			return;
-		}
-
 		let targetPopover: HTMLDetailsElement | null;
 		targetPopover =
 			event.target instanceof Element &&
@@ -29,10 +21,14 @@ export function initPopovers(container: HTMLElement) {
 			return;
 		}
 
-		if (!targetPopover.hasAttribute('open')) {
+		const allPopovers: HTMLDetailsElement[] = Array.from(
+			container.querySelectorAll('details.popover')
+		);
+
+		if (!targetPopover.open) {
 			const keyHandler = (event: KeyboardEvent) => {
 				if (event.key === 'Escape') {
-					closePopovers(allPopovers);
+					closePopovers();
 				}
 			};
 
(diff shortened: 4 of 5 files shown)


site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,33 @@
...
+				}
+				p.focus();
+			}
+		})
+	);
+
+	container.addEventListener('click', event => {
+		const targetPopover =
+			event.target instanceof Element && (event.target.closest('details') as HTMLDetailsElement)
+				? (event.target.closest('details') as HTMLDetailsElement)
```
wibjorn, line 22: selector issue again

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,33 @@
+export function initPopover(container: Element) {
+	const popovers = container.querySelectorAll('details');
```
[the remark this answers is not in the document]
wibjorn: Array.from()

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,33 @@
...
+		})
+	);
+
+	container.addEventListener('click', event => {
+		const targetPopover =
+			event.target instanceof Element && (event.target.closest('details') as HTMLDetailsElement)
+				? (event.target.closest('details') as HTMLDetailsElement)
+				: null;
+
+		const otherPopovers = Array.from(popovers).filter(el => el !== targetPopover);
```
wibjorn, line 25: One issue with this is that this will all work as the page loads, but we cannot assume that a `.popover` element might not be added to the DOM later. I think you'll have to querySelector popovers on keypress and on click.

wibjorn (dismissed):

site/src/scaffold/scripts/popover.ts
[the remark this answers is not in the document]
wibjorn: Also, see comment below about timing of selection

site/src/scaffold/scripts/popover.ts
ollips: updated

site/src/scaffold/scripts/popover.ts
ollips: updated

site/src/scaffold/scripts/popover.ts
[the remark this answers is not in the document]
ollips: updated

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,33 @@
+export function initPopover(container: Element) {
```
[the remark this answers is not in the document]
ollips: updated

.changeset/grumpy-experts-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-site': minor
+---
+
+Adding event listeners to popover
```
homing1, line 5: supernit: period

ollips: Updated

homing1 (approved):

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,37 @@
...
+	container.addEventListener('click', event => {
+		const popovers: HTMLDetailsElement[] = Array.from(
+			container.querySelectorAll('details.popover')
+		);
+
+		if (popovers.length === 0) {
+			return;
+		}
+
+		popovers.forEach(p =>
```
wibjorn, line 11: Does this add new event listeners on each subsequent click?

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,37 @@
...
+
+		const targetPopover =
+			event.target instanceof HTMLElement &&
+			(event.target.closest('details.popover') as HTMLDetailsElement)
+				? (event.target.closest('details.popover') as HTMLDetailsElement)
+				: null;
+
+		const otherPopovers = popovers.filter(el => el !== targetPopover);
+
+		otherPopovers.forEach(p => {
```
wibjorn, line 27: nit: can be single line

site/src/scaffold/scripts/popover.ts
wibjorn: also consider, for ... of loop instead

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,37 @@
+export function initPopover(container: HTMLElement) {
```
wibjorn, line 1: `initPopovers` (with s)?

site/src/scaffold/scripts/popover.ts
jdanyow: I don't think the loop will be required. This code from dropdown.ts is still good and covers touch, virtual dom, and edge cases with iframes.
https://dev.azure.com/ceapex/Engineering/_git/docs-ui?path=/packages/scripts/src/dropdown.ts&version=GBdevelop&line=38&lineEnd=90&lineStartColumn=1&lineEndColumn=49&lineStyle=plain&_a=contents

site/src/scaffold/scripts/popover.ts
ollips: Updated

site/src/scaffold/scripts/popover.ts
ollips: Updated

site/src/scaffold/scripts/popover.ts
ollips: Changed the approach, please revisit. I don't assign the key event listener to every popover with random clicks on a page.

site/src/scaffold/scripts/popover.ts
wibjorn: Forgot to push? 

`otherPopovers.forEach(p => closePopover(p))`?

site/src/scaffold/scripts/popover.ts
wibjorn: Revisit the link Jeremy posted above?

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,80 @@
...
+				}
+
+				if (!(targetPopover && targetPopover.contains(event.target))) {
+					closePopovers(allPopovers);
+				}
+			};
+
+			const blurHandler = () => {
+				if (
+					((document.activeElement && document.activeElement.nodeName) || '').toLowerCase() ===
```
jdanyow, line 51: nit- would be nice to change this to the newer syntax `(document.activeElement?.nodeName ?? '') === 'IFRAME'`

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,80 @@
+export function initPopovers(container: HTMLElement) {
+	container.addEventListener('click', (event: Event) => {
+		const allPopovers: HTMLDetailsElement[] = Array.from(
+			container.querySelectorAll('details.popover')
+		);
+
+		if (allPopovers.length === 0) {
+			return;
+		}
```
jdanyow, line 9: I think the logic can be reduced by using the `<details>` element's toggle event and by removing the looping over all popovers, similar to the .dropdown handling.

```ts
container.addEventListener('toggle', event => {
  // Was a details.popover element toggled *open*?
  if (!(event.target instanceof HTMLDetailsElement) || !event.target.open || !event.target.classList.contains('popover')) {
    return;
  }
  
  const details = event.target;
  
  // Synchronize aria-expanded with the details element's open state.
  details.setAttribute('aria-expanded', details.open);

  // Make sure the details element collapses automatically...

  const checkTarget = (event: Event) => {
    if (!(event.target instanceof Element)) {
      return;
    }
    
    if (!details.contains(event.target) /*|| (event.type === 'click' && event.target.closest('.dropdown-menu-close')*/) {
      collapse();
      /*if (event.target.classList.contains('dropdown-menu-close')) {
        trigger.focus();
      }*/
    }
  };
  
  const keyHandler = (event: KeyboardEvent) => {
    if (event.keyCode === keyCodes.escape) {
      collapse();
      details.querySelector('summary')!.focus();
    }
  };
  
  const blurHandler = async () => {
    if ((document.activeElement?.nodeName ?? '') ==='IFRAME') {
      collapse();
    }
  };
  
  collapse = () => {
    container.removeEventListener('focus', checkTarget);
    container.removeEventListener('click', checkTarget);
    container.removeEventListener('touchstart', checkTarget);
    container.removeEventListener('keydown', keyHandler);
    window.removeEventListener('blur', blurHandler);
    details.removeEventListener('toggle', collapse);
    details.setAttribute('aria-expanded', 'false');
    details.toggleAttribute('open', false);
  };
  
  const useCapture = true;
  container.addEventListener('focus', checkTarget, useCapture);
  container.addEventListener('click', checkTarget);
  container.addEventListener('touchstart', checkTarget);
  container.addEventListener('keydown', keyHandler);
  window.addEventListener('blur', blurHandler);
  details.addEventListener('toggle', collapse);
});
```

site/src/scaffold/scripts/popover.ts
ollips: Updated

site/src/scaffold/scripts/popover.ts
ollips: The idea is great, but you still have to assign it to each popover first. And because we can not guarantee that the popover component won't be added to the page sometime after the initial content loaded, we still need to keep the body's click event to verify the amount of popover's on a page. And every with every click we would loop through each popover and assign it the `toggleHandler` event, and to avoid tons of them assigned to the same popover we need to clean them up somehow too. So it doesn't really look like a path to follow, to be honest. Please correct me if I'm wrong :) 

The way I was also thinking about using the `toggle` event is it could be useful in switching the `aria-expanded` value, but it's the whole other story - when I added that attribute and was testing locally with Narrator and NVDA I found out that having this additional attribute doesn't make a difference in how the screen readers announce (does not actually) when the details component opens/closes. So I'm keeping that PR about event listeners, and want to work on accessibility in a separate one.

css/src/components/popover.scss
```diff
@@ -21,7 +25,6 @@ $popover-width: 224px !default;
 		border-radius: $popover-border-radius;
 		background-color: $popover-background-color;
 		box-shadow: $popover-shadow;
-		overflow: hidden;
```
ollips, line 24: _Reviewers note:_ removed the `overflow: hidden;` to cover those cases when one popover exists in another one to make its content visible.

site/src/scaffold/scripts/popover.ts
ollips: Updated

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,72 @@
+export function initPopovers(container: HTMLElement) {
+	container.addEventListener('click', (event: Event) => {
+		let targetPopover: HTMLDetailsElement | null;
```
wibjorn, line 3: I may be reading this incorrectly, but it seems to me like we can condense these two lines without sacrificing readability, thanks to our friend the `?.` optional chaining operator. 

```ts
	const targetPopover =
			event.target instanceof Element &&
			(event.target.closest('details.popover') as HTMLDetailsElement) ||
			// or
			event.target instanceof Element &&
			event.target.shadowRoot?.activeElement?.closest('details.popover') as HTMLDetailsElement;
```

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,72 @@
...
+		targetPopover =
+			event.target instanceof Element &&
```
wibjorn, line 5: Mind explaining why we're using `Element` here instead of `HTMLElement`?

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,72 @@
...
+
+		if (!targetPopover) {
+			return;
+		}
+
+		const allPopovers: HTMLDetailsElement[] = Array.from(
+			container.querySelectorAll('details.popover')
+		);
+
+		if (!targetPopover.open) {
```
wibjorn, line 28: question: can this be placed above the querySelection of `allPopovers`? so that `if (!targetPopover.open)` can be removed. Just reduces the amount of nesting.

```ts
if (targetPopover.open) {
  return;
}

File: site/src/scaffold/scripts/popover.ts [hunk 15]
```diff
@@ -0,0 +1,72 @@
...
+			const keyHandler = (event: KeyboardEvent) => {
+				if (event.key === 'Escape') {
+					closePopovers();
+				}
+			};
+
+			const clickHandler = (event: Event) => {
+				if (!(event.target instanceof Element)) {
```
wibjorn, line 36: same q about `HTMLElement`.

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,72 @@
...
+					return;
+				}
+
+				if (!targetPopover?.contains(event.target)) {
+					targetPopover ? closePopovers([targetPopover]) : closePopovers();
```
wibjorn, line 41: I'm not certain about this, and it seems fine to me, but I have a feeling the `no unused expresssion` lint rules that are enabled in other repos will complain about it. solution would just be to un-nest and use `if else` (I think). Might be worth pasting this into another repo and running lint.

wibjorn (changes requested):

BenDMyers:
As I was validating the popovers on the Atlas docs, I was tripped up by my own assumption that the popovers would be left/center/right-aligned relative to the trigger, not the parent width. Can we add a comment to the docs page that clarifies that?

site/src/scaffold/scripts/popover.ts
ollips: Because if there is a SVG icon pasted on the summary (as in one of the examples) the target would be `SVGElement`, so instead of writing here `event.target instanceof HTMLElement || event.target instanceof SVGElement` I decided to go with `event.target instanceof Element`

site/src/scaffold/scripts/popover.ts
ollips: Sure 👍🏻

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,72 @@
...
+				}
+			};
+
+			const blurHandler = () => {
+				if ((document.activeElement?.nodeName || '').toLowerCase() === 'iframe') {
```
BenDMyers, line 46: I think we can remove the need for the `|| ''` fallback by continuing optional chaining on the `toLowerCase` method:

```suggestion
				if (document.activeElement?.nodeName?.toLowerCase() === 'iframe') {
```

Thoughts?

site/src/scaffold/scripts/popover.ts
ollips: Answered above

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,72 @@
...
+		);
+
+		if (!targetPopover.open) {
+			const keyHandler = (event: KeyboardEvent) => {
+				if (event.key === 'Escape') {
+					closePopovers();
+				}
+			};
+
+			const clickHandler = (event: Event) => {
```
jdanyow, line 35: Thinking this should be named something like checkTarget or checkShouldCollapse to describe what it's doing rather than the event it's handling. Plus it's used with focus, touch, and click.

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,72 @@
...
+				}
+			};
+
+			const blurHandler = () => {
+				if ((document.activeElement?.nodeName || '').toLowerCase() === 'iframe') {
+					closePopovers();
+				}
+			};
+
+			const closePopovers = (popovers: HTMLDetailsElement[] = allPopovers) => {
```
jdanyow, line 51: related to other comment- this only needs to close the targetPopover. Other open popovers will have their own event listeners and will close themselves.

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,72 @@
...
+					: null;
+		}
+
+		if (!targetPopover) {
+			return;
+		}
+
+		const allPopovers: HTMLDetailsElement[] = Array.from(
+			container.querySelectorAll('details.popover')
+		);
```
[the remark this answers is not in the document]
ollips: 🤦🏻‍♀️ Updated

site/src/scaffold/scripts/popover.ts
ollips: Updated

site/src/scaffold/scripts/popover.ts
ollips: 🤦🏻‍♀️ Updated

site/src/scaffold/scripts/popover.ts
ollips: Updated

site/src/scaffold/scripts/popover.ts
ollips: Updated

site/src/scaffold/scripts/popover.ts
ollips: Thanks for the suggestion, updated!

site/src/scaffold/scripts/popover.ts
```diff
@@ -0,0 +1,52 @@
+export function initPopovers(container: HTMLElement) {
+	container.addEventListener('click', (event: Event) => {
```
jdanyow, line 2: thinking this could be toggle since we only care when the details menu is opening at this point in the code.

site/src/scaffold/scripts/popover.ts
ollips: Refactored that section, don't think it's relevant anymore.

ollips:
> As I was validating the popovers on the Atlas docs, I was tripped up by my own assumption that the popovers would be left/center/right-aligned relative to the trigger, not the parent width. Can we add a comment to the docs page that clarifies that?

Updated! Let me know if my words look good to you 😁

site/src/components/popover.md
```diff
@@ -53,21 +53,21 @@ You can apply almost any classes on the `summary` element to achieve the look yo
...
 <details class="popover popover-center margin-xxs">
 	<summary class="button">Popover centered</summary>
-	<div class="popover-content">Popover content.</div>
+	<div class="popover-content border">Popover content.</div>
 </details>
 
 <details class="popover popover-right margin-xxs">
 	<summary class="button">Popover aligned to the right</summary>
-	<div class="popover-content">Popover content.</div>
+	<div class="popover-content border">Popover content.</div>
```
ollips, line 71: @jdanyow is that the border update you meant?

image

site/src/scaffold/scripts/popover.ts
ollips: I finally figured out why it didn't work for me on my previous attempts, updated!

site/src/components/popover.md
jdanyow: that will do it!

jdanyow (dismissed):

wibjorn (approved):

BenDMyers (approved):

BenDMyers (approved):

jdanyow (approved):

PR #300, opened 2021-12-21
Adding tables modifiers

ollips:
Task: task-442333

Link: preview-300

Adding more table modifiers:

| Class                         |
| ----------------------------- |
| `.table-simple` |
| `.table-complex` |
| `.table-borderless` |
| `.table-striped` |
| `.table-sm` |
| `.table-lg` |

## Testing

1. Visit [table](https://design.docs.microsoft.com/pulls/300/components/table.html) page
2. Review the new modifiers in use and share your feedback :)


css/src/components/table.scss
```diff
@@ -64,4 +67,59 @@ $table-caption-spacing: $letter-spacing-medium !default;
...
+	}
+
+	&.table-lg {
+		th,
+		td {
+			padding: $table-cell-padding-loose;
+		}
+	}
+
+	&.table-width-reset {
```
wibjorn, line 116: Width modifier `.width-auto { width: auto !important; }`?

css/src/components/table.scss
```diff
@@ -64,4 +67,59 @@ $table-caption-spacing: $letter-spacing-medium !default;
...
+		width: auto;
+	}
+
+	&.table-interactive {
```
wibjorn, line 120: I think we should remove `tablet-interactive` for now as well. We don't have direct use case for this right now, and don't know if it may require JS eventually.

site/src/components/table.md
```diff
@@ -93,6 +94,231 @@ For cases in which a table has many columns, the `table-wrapper` class is availa
...
+		<tr>
+			<td>Row three</td>
+			<td>Cell</td>
+			<td>Cell</td>
+		</tr>
+	</tbody>
+</table>
+```
+
+_`table-complex`_ - complex table style is a completely bordered cell layout with a heavy stylized header row.
```
wibjorn, line 128: Should we use headings?

wibjorn:
```md
### Complex
```

site/src/components/table.md
```diff
@@ -93,6 +94,231 @@ For cases in which a table has many columns, the `table-wrapper` class is availa
 
 ## Modifiers
 
+_`table-simple`_ - simple table style only shows horizontal cell borders and no table border.
```
wibjorn, line 97: Thinking that this page has gotten extremely long. Just floating the idea, but perhaps we could just abbreviate or remove the code blocks for examples?

css/src/components/table.scss
ollips: In the [tables discussion](https://github.com/microsoft/atlas-design/discussions/170#discussioncomment-847305) Jeremy mentioned there is a need in highlighting the row, so I though we can do at leat the hover effect. And later figure out what to do (or not do) with keyboard focusing on the rows.

css/src/components/table.scss
ollips: It feels like it can be just a general helper, but I don't think we have a space for listing that kind of random classes. Any thoughts on what's the best place for that? Or should I keep it inside the `tables.scss` file? If so, I think it should be called `table-width-auto`.

site/src/components/table.md
ollips: I don't like how long the page became either, but I couldn't find out the best solution to shorten it. Removing all examples won't work, because I think some cases should be presented such as stacking on mobile, or how the complex table looks since it's pretty different from the default. And at the same time having examples for a few cases and not all feel weird to me. I'll try to shorten the code examples, showing only 2 rows in the table instead of 3 might work.

css/src/components/table.scss
wibjorn: I'd say -- if we really need this at all -- we can place our first width atomic in `/atomics/width.scss`. Otherwise, we can remove it entirely.

css/src/components/table.scss
wibjorn: Row highlighting is quite different from a hover effect background color change. Still seems best to me that we remove this until a feature really requires it and we have a better understanding of it's accessibility concerns.

ollips: Sounds good, will remove it!

site/src/components/table.md
wibjorn: What about just removing the code blocks, not the examples?

css/src/components/table.scss
ollips: Updated

site/src/components/table.md
ollips: Updated

site/src/components/table.md
ollips: Updated

css/src/components/table.scss
ollips: I remember there were cases before where we needed to reset the table's width but can't find any cases right now. So skipping on adding this right now, but I think we should add it eventually.

site/src/components/table.md
homing1: If we want to shorten the page, I'm wondering if the first example can be combined with the second example?

ollips: I think it's fine the way it is right now after the latest update - having two code examples is much better than 9 as it used to be :)

wibjorn: I agree that the second table-wrapper example code seems like it could be removed.

site/src/components/table.md
```diff
@@ -93,10 +94,205 @@ For cases in which a table has many columns, the `table-wrapper` class is availa
...
+			<td>Row three</td>
+			<td>Cell</td>
+			<td>Cell</td>
+		</tr>
+	</tbody>
+</table>
+
+### Stacked on mobile
+
+_`table-stacked-mobile`_ stacks table cells on mobile screen sizes.
```
wibjorn, line 293: On mobile, this doesn't look so good, I also don't see any code related to it.

image

wibjorn (changes requested):

site/src/components/table.md
ollips: Ugh with a new default border around the table I forgot to take a closer look at this view. Good catch, updated!

site/src/components/table.md
ollips: Updated

.changeset/nice-news-peel.md
```diff
@@ -0,0 +1,6 @@
+---
+'@microsoft/atlas-css': minor
+'@microsoft/atlas-site': minor
+---
+
+Adding table modifiers; updating site with relevant documenation
```
homing1, line 6: "documentation."?

wibjorn (dismissed):

.changeset/nice-news-peel.md
ollips: Updated

homing1: still seeing "documenation" instead of "documentation"

ollips: Oh, my fault, I thought you were pointing the missed comma! Updated now

wibjorn (dismissed):

site/src/components/table.md
```diff
@@ -86,17 +86,28 @@ For cases in which a table has many columns, the `table-wrapper` class is availa
...
 ## Modifiers
 
-_`table-stacked-mobile`_ - stacks table cells on mobile screen sizes.
+All modifier classes should be applied in conjunction with the `.table` class.
 
-```html
-<table class="table table-stacked-mobile">
+### Simple
+
+_`table-simple`_ gives table has only horizontal cell borders.
```
BenDMyers, line 108: I think we may want to tweak the writing a bit here:

```suggestion
_`table-simple`_ gives the table only horizontal cell borders.
```

site/src/components/table.md
```diff
@@ -122,4 +133,202 @@ _`table-stacked-mobile`_ - stacks table cells on mobile screen sizes.
 		</tr>
 	</tbody>
 </table>
-```
+
+### Complex
+
+_`table-complex`_ - complex table style is a completely bordered cell layout with a heavy stylized header row.
```
BenDMyers, line 139: I've got a little recommendation here to align how we talk about `table_complex` a bit more similarly to how we talk about the other classes:

```suggestion
_`table-complex`_ gives the table a completely bordered cell layout with a heavily stylized header row.
```

site/src/components/table.md
ollips: 🤦🏻‍♀️

heskew (dismissed):

site/src/components/table.md
ollips: Updated

site/src/components/table.md
ollips: Updated

wibjorn (approved):

BenDMyers (approved):

BenDMyers (approved):

wibjorn (approved):

PR #304, opened 2022-01-03
update breadcrumb to allow for an initial slash, document article-header pattern

wibjorn:
Task: task-[work-item-number]

Link: preview-304

Some of the tandem work, such as automatic truncation of breadcrumb items, requires greater control over the container within which they'll be placed on the page. To that end, here is `article-header`, which is a container that can house and overflow menu (or two!) and some overflow-hidden breadcrumbs. Made a few changes to breadcrumbs as well to add the ability to configure an initial slash. 

I've included SVG icons, which were mangled by parcel. You can assume for now that those icons won't be messed up in a normal environment. I will continue researching a fix for the SVG mangling.

## Testing

1. [Article Header](https://design.docs.microsoft.com/pulls/304/patterns/article-header.html)
2. [Breadcrumbs](https://design.docs.microsoft.com/pulls/304/components/breadcrumbs.html)
3. [Spacing](https://design.docs.microsoft.com/pulls/304/atomics/spacing.html)


css/src/components/breadcrumbs.scss
```diff
@@ -11,10 +17,21 @@ $breadcrumb-separator: '\00200B/' !default;
...
-			color: $text-subtle;
-			content: $breadcrumb-separator;
 		}
 	}
+
+	// stylelint-disable selector-max-specificity
+
+	&.breadcrumbs-initial-slash {
+		.breadcrumbs-item:first-child:not(.breadcrumbs-items-slashless)::before,
+		.breadcrumbs-item[hidden] + .breadcrumbs-item:not(.breadcrumbs-items-slashless)::before {
```
jdanyow, line 30: feel free to resolve: do we need `:not(.breadcrumbs-items-slashless)` in these selectors? I was thinking `.breadcrumbs-items-slashless` was to disable the default trailing slash.

jdanyow (approved):

css/src/components/breadcrumbs.scss
ollips, line 30: Misspelled - `breadcrumbs-item-slashless`. Also, agree with Jeremy's comment - feels like the `:not()` is not needed here

ollips (approved):

jdanyow (approved):

PR #313, opened 2022-01-07
Add checkbox component

homing1:
Task: task-[533239](https://dev.azure.com/ceapex/Engineering/_boards/board/t/Rendering%20Team/Stories/?workitem=533239)

Link: preview-[313](https://design.docs.microsoft.com/pulls/313/components/checkbox.html)

This PR adds the checkbox component. The former structure is kept with some edits. Docons is no longer used for the checkmark. 

Things to consider: 
- There is no styling for the [indeterminate](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox#attr-indeterminate) state.

## Testing

1. Visit https://design.docs.microsoft.com/pulls/313/components/checkbox.html
2. Review code.


css/src/components/checkbox.scss
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+
+	&:hover {
+		.checkbox-check::before {
+			@include checkmark;
+		}
+	}
+
+	&.is-small {
```
wibjorn, line 56: This we'll need to rename this one: `checkbox-sm` see https://design.docs.microsoft.com/components/overview.html

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,118 @@
...
+			}
+		}
+
+		&[disabled] + .checkbox-check {
+			opacity: 0.6;
+			cursor: not-allowed;
+		}
+	}
+
+	&.is-secondary {
```
wibjorn, line 97: remove?

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,118 @@
...
+				background-color: $secondary;
+
+				&::before {
+					color: $secondary-invert;
+				}
+			}
+		}
+	}
+
+	.checkbox-check.is-checked {
```
wibjorn, line 110: `is-checked` is a state based class, so it can actually stay this way I think. What do you think?

https://design.docs.microsoft.com/components/overview.html

css/src/mixins/checkbox.scss
```diff
@@ -0,0 +1,11 @@
+@mixin checkmark {
+	display: block;
+	position: absolute;
+	top: 1px;
+	width: calc($checkbox-icon-size / 2);
+	height: $checkbox-icon-size;
+	transform: rotate(45deg);
+	border: solid $secondary-hover;
```
wibjorn, line 8: Is there a way we can use something other than secondary? That particular color isn't very well supported.

css/src/components/checkbox.scss
homing1: I believe that the only use case we have for this is in facet search. Should we still remove? 
image

css/src/components/checkbox.scss
homing1: I agree. Something like `checkbox-check-checked` would be too confusing.

css/src/mixins/checkbox.scss
homing1: Updated to $text-subtle

css/src/components/checkbox.scss
homing1: I wonder if the color distinction between parent and child is actually needed for this scenario. 
image

css/src/components/checkbox.scss
homing1: Indeed, updated

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,118 @@
+/* stylelint-disable selector-max-specificity, max-nesting-depth, selector-no-qualifying-type */
+
+$checkbox-border-radius: $radius-sm !default;
+$checkbox-border-color: $text !default;
+$checkbox-check-hover-color: $secondary !default;
+$checkbox-icon-size: 0.875rem !default;
```
wibjorn, line 6: Internal spacing for components should be based on `em` whenever possible, so that we can easily modify sizes with font-size alone.

css/src/components/checkbox.scss
wibjorn: I was done by design, so we should keep something similar. Perhaps we need to circle back with design regarding this? .checkbox-muted could be the color modifier? .checkbox could be the standard blue while -muted could be the new 'secondary'. Thoughts?

It's probably too much to hope that `border` or `background-color-medium` color would work, on all themes, any chance there?

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,118 @@
...
+		display: flex;
+		flex-shrink: 0;
+		align-items: center;
+		justify-content: center;
+		width: $checkbox-size;
+		height: $checkbox-size;
+		transition-duration: $checkbox-duration;
+		transition-property: background, border, border-color;
+		transition-timing-function: $checkbox-timing-function;
+		border: 1px solid $checkbox-border-color;
```
wibjorn, line 37: Think that our border needs some work, since it needs to be the same color as the background when its checked:

This PR: 
image

Figma
image

Old:
image

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,118 @@
...
+
+	&:hover {
+		.checkbox-check::before {
+			@include checkmark;
+		}
+	}
+
+	&.checkbox-sm {
+		.checkbox-check {
+			width: $checkbox-size-small;
```
wibjorn, line 58: ideally, after converting most inner spacing to `em` this rule becomes something like:

```
.checkbox-sm {
  font-size: 14px; // or whatever size.
}

File: site/src/components/checkbox.md [hunk 8]
```diff
@@ -0,0 +1,111 @@
...
+		</thead>
+		<tbody>
+			<tr>
+				<td>Unchecked</td>
+				<td>
+					<label class="checkbox">
+						<input type="checkbox" />
+						<span class="checkbox-check" role="presentation" aria-hidden="true"></span>
+						<span class="checkbox-text" aria-hidden="true">Default</span>
+					</label>
```
jdanyow, line 29: Now that there's better support for [accent-color](https://developer.mozilla.org/en-US/docs/Web/CSS/accent-color) let's retire the special markup to enable comething more like this:
```html
<label class="checkbox"><input type="checkbox" /> Reindeer</label>
```

It would be worth considering whether we need a class at all or whether accent-color should be part of our base styles.

cc @wibjorn

homing1: Without the custom checkbox, it would look like something on the left 
image

Here's a codepen to play with: https://codepen.io/homing1/pen/mdBzRLV

Some implications:
- The checkmark color will be defined by the accent color/browser. There will be no color difference from current behavior with all themes so this is ok.
- There will be a subtle background color change on hover checked, but looks ok.
- We would lose the grey checkmark on hover unchecked.

Should we loop in design? 
cc @Courtny

jdanyow: Yes- please loop in design. I'm thinking the accent-color should be set globally.

```css
:root {
 accent-color: var(--theme-primary-base);
}

.checkbox2 {
 display: inline-flex;
 align-items: center;
 gap: 0.1875em;
}
```

With markup like this:

```html
<label class="checkbox2"><input type="checkbox"> Hello world</label>
```

homing1: Moving discussion to here https://github.com/microsoft/atlas-design/discussions/152

wibjorn: Seems to me the accent color approach has two major downsides:
- we still need to style the check, because the browser check is pretty ugly.
- we lose the smooth (albiet) subtle transition of the background color on state changes.

css/src/components/checkbox.scss
homing1: I think `border` ($palette-grey-30) and `body-background-medium` ($palette-grey-20) is too light with light theme and doesn't pass 3:1 color contrast ratio. 

I just realized that `$secondary` ($palette-grey-60) doesn't pass either when paired with the darker background in my screenshot above, even though it's what we have in prod right now.

The sweet spot seems to be `$palette-grey-80`, but that's currently unused by color/theme tokens. 

I updated to `$text-subtle` ($palette-grey-90) for now, which works with all themes. It's more like a `.checkbox-dark` rather than muted though.

css/src/components/checkbox.scss
homing1: Updated to `em`.

css/src/components/checkbox.scss
homing1: Updated border color. I made a slight adjustment to the checkmark size, but it may require more tweaking.

css/src/components/checkbox.scss
homing1: updated

site/src/components/checkbox.md
```diff
@@ -0,0 +1,111 @@
...
+```html
+<label class="checkbox">
+	<input type="checkbox" />
+	<span class="checkbox-check" role="presentation" aria-hidden="true"></span>
+	<span class="checkbox-text" aria-hidden="true">Checkbox 1</span>
+</label>
+
+<label class="checkbox">
+	<input type="checkbox" />
+	<span class="checkbox-check is-checked" role="presentation" aria-hidden="true"></span>
```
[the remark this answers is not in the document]
homing1: The use case I've found for this modifier (perhaps the only one) is on the assessment answer summary, where we show the selected answers, but the answers are not interactive. I'm assuming that the modifier was added solely for this purpose.

image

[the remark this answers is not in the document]
ollips: Right, I meant to remove it from that specific example, cause for demonstrating that class usage you have another example further on a page and it is explained over there. On a first usage example, I wouldn't expect to see the non-interactive checkbox 🤔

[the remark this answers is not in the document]
homing1: Ah you're right, updated

site/src/components/checkbox.md
homing1: It doesn't seem like there's a good overall solution to this. Design is against losing the color change on hover unchecked. Let's not pursue this approach until there's a way to make it work for our purposes.

css/src/components/checkbox.scss
homing1: Design is good with the combo of $palette-grey-90 with a white checkmark.

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,121 @@
...
+		}
+	}
+
+	&:hover {
+		.checkbox-check::before {
+			@include checkmark;
+		}
+	}
+
+	&.checkbox-sm {
```
wibjorn, line 55: after the `em` sizing changes, can't we just use `.checkbox-sm { font-size: $font-size-7; }`?

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,121 @@
+/* stylelint-disable selector-max-specificity, max-nesting-depth, selector-no-qualifying-type */
+
+$checkbox-border-radius: $radius-sm !default;
+$checkbox-border-color: $text !default;
+$checkbox-check-hover-color: $secondary !default;
```
wibjorn, line 5: audit for unused variables?

css/src/components/checkbox.scss
wibjorn: (and cut the rest)

homing1: Not quite. At `.checkbox-sm`, 1.25em size equates to 17.5px, which is too large so the unit needs to be adjusted to 1em. Updated to `$font-size-8` (14px) and removed extra `.checkbox-text` rule.

css/src/components/checkbox.scss
homing1: Good catch, `$checkbox-check-hover-color` is now used in the checkmark mixin.

css/src/components/checkbox.scss
homing1: Updated so that `1em` size works for both.

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,117 @@
...
+
+.checkbox {
+	display: flex;
+	position: relative;
+	align-items: center;
+	font-size: $font-size-5;
+	line-height: 1.25;
+	cursor: pointer;
+
+	+ .checkbox {
```
wibjorn, line 22: Should we remove the `.checkbox + .checkbox` rule entirely? This type of spacing is what atomics are for.

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,117 @@
...
+$checkbox-icon-size: 0.8em !default;
+$checkbox-timing-function: $input-timing-function !default;
+$checkbox-duration: $input-transition-duration !default;
+$checkbox-spacing: $spacer-3;
+
+.checkbox {
+	display: flex;
+	position: relative;
+	align-items: center;
+	font-size: $font-size-5;
```
wibjorn, line 18: top level font-size looks like it can be removed as it matches the document's font size

css/src/mixins/checkbox.scss
```diff
@@ -0,0 +1,11 @@
+@mixin checkmark {
```
wibjorn, line 1: I think that this mixin should live in the file with checkbox. The mixins folder is a better place for mixins that need to be globally available, such as `visually-hidden`.

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,117 @@
...
+		width: $checkbox-size;
+		height: $checkbox-size;
+		transition-duration: $checkbox-duration;
+		transition-property: background, border, border-color;
+		transition-timing-function: $checkbox-timing-function;
+		border: 1px solid $checkbox-border-color;
+		border-radius: $checkbox-border-radius;
+
+		&:not(:last-child) {
+			margin-#{$user-right}: 0.5rem;
```
wibjorn, line 41: Use logical prop instead `margin-inline-end`

css/src/components/checkbox.scss
wibjorn: Toyed around with this a bit locally.

```
	&.checkbox-sm {
		font-size: 14px;
	}
```

Is sufficient to resize the whole box, and keep things in proportion. Need to update borders to use `em` sizing too.

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,117 @@
+/* stylelint-disable selector-max-specificity, max-nesting-depth, selector-no-qualifying-type */
+
+$checkbox-border-radius: $radius-sm !default;
+$checkbox-border-color: $text !default;
+$checkbox-color: $primary !default;
+$checkbox-check-color: $primary-invert !default;
+$checkbox-check-hover-color: $text-subtle !default;
+$checkbox-size: 1em !default;
```
wibjorn, line 8: I've just put a comment in the Figma related to this, but I think we should keep the text size of the default tied to 1rem, which to keep relatively similar proportions means we'd want to update the following to variables:

```
$checkbox-size: 1.25em !default;
$checkbox-icon-size: 0.875em !default;

File: css/src/components/checkbox.scss [hunk 17]
```diff
@@ -0,0 +1,117 @@
+/* stylelint-disable selector-max-specificity, max-nesting-depth, selector-no-qualifying-type */
+
+$checkbox-border-radius: $radius-sm !default;
+$checkbox-border-color: $text !default;
```
wibjorn, line 4: nit for clarity: `$checkbox-unchecked-border-color`

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,117 @@
...
+			margin-#{$user-right}: 0.5rem;
+		}
+
+		&::before {
+			font-size: $checkbox-icon-size;
+			content: '\0020';
+		}
+	}
+
+	.checkbox-text {
```
wibjorn, line 50: As said above, question out to design in Figma, but I think this rule should be removed, in favor of using the document font size - or the size modifier on the parent (e.i. `checkbox-sm`).

wibjorn (changes requested):

css/src/components/checkbox.scss
homing1: removed

css/src/components/checkbox.scss
homing1: removed

css/src/mixins/checkbox.scss
homing1: Moved to the checkbox file

css/src/components/checkbox.scss
homing1: updated

css/src/components/checkbox.scss
homing1: updated

css/src/components/checkbox.scss
homing1: Reverted back to the sizes in your comment.

css/src/components/checkbox.scss
homing1: Updated

css/src/components/checkbox.scss
homing1: removed

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,113 @@
...
+$checkbox-size: 1.25em !default;
+$checkbox-icon-size: 0.875em !default;
+$checkbox-timing-function: $input-timing-function !default;
+$checkbox-duration: $input-transition-duration !default;
+$checkbox-spacing: $spacer-3;
+
+@mixin checkmark {
+	display: block;
+	position: absolute;
+	top: 0.2em;
```
wibjorn, line 17: should the em sizes in this file map a little better to font sizes at 16px?
1px = 0.0625
2px = .125
3px = .1875 etc.

homing1: updated

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,112 @@
...
+
+@mixin checkmark {
+	display: block;
+	position: absolute;
+	top: 0.1875em;
+	width: 0.3125em;
+	height: $checkbox-icon-size;
+	transform: rotate(45deg);
+	border: solid $checkbox-check-hover-color;
+	border-width: 0 0.078em 0.078em 0;
```
wibjorn, line 22: Should border widths be pixel based? `0.0625` is one pixel

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,112 @@
...
+		&:not(:last-child) {
+			margin-inline-end: 0.5rem;
+		}
+
+		&::before {
+			content: '\0020';
+		}
+	}
+
+	&:hover {
```
wibjorn, line 55: nit: change to single selector, nesting doesn't provide any less duplication after recent changes  `.checkbox-check::before:hover`

site/src/components/checkbox.md
```diff
@@ -0,0 +1,111 @@
+---
+title: Checkbox
+description: The checkbox component in the Atlas Design System
+template: standard
+figmaEmbed: https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FuVA2amRR71yJZ0GS6RI6zL%2F%25F0%259F%258C%259E-Atlas-Design-Library%3Fnode-id%3D855%253A1014"
+---
+
+# Checkbox
+
+A selectable menu list from which an user can make one or more selections.
```
wibjorn, line 10: I think that this may not quite be the right definition, perhaps something more akin to what MDN says? 
> A checkbox allows you to select single values for submission in a form

css/src/components/checkbox.scss
homing1: Updated to 0.0625. 2 px is too large.

css/src/components/checkbox.scss
homing1: `.checkbox-check::before:hover` seems to lose the hover effect, but `.checkbox-check:hover::before` seems to work.

site/src/components/checkbox.md
homing1: Updated

css/src/components/checkbox.scss
homing1: Actually, that removed the checkmark on label hover. Updated to `&:hover .checkbox-check::before`

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,110 @@
...
+		&:not(:last-child) {
+			margin-inline-end: 0.5rem;
+		}
+
+		&::before {
+			content: '\0020';
+		}
+	}
+
+	&:hover .checkbox-check::before {
```
ollips, line 55: 👍🏻👍🏻👍🏻

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,110 @@
...
+$checkbox-icon-size: 0.75em !default;
+$checkbox-timing-function: $input-timing-function !default;
+$checkbox-duration: $input-transition-duration !default;
+$checkbox-spacing: $spacer-3;
+
+@mixin checkmark {
+	display: block;
+	position: absolute;
+	top: 0.1875em;
+	width: 0.3125em;
```
ollips, line 18: Shouldn't these two values `top` and `width` be variables? As it's done for height. It feels like if height can be changed, these two should be flexible as well since they're tight together.

homing1: Added variables for top, width, and border-width.

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,114 @@
+/* stylelint-disable selector-max-specificity, max-nesting-depth, selector-no-qualifying-type */
+
+$checkbox-border-radius: $radius-sm !default;
+$checkbox-border-width: 0.0625em !default;
+$checkbox-unchecked-border-color: $text !default;
+$checkbox-color: $primary !default;
+$checkbox-check-color: $primary-invert !default;
+$checkbox-check-hover-color: $text-subtle !default;
+$checkbox-size: 1.25em !default;
+$checkbox-icon-size: 0.75em !default;
```
ollips, line 10: rename to `$checkbox-icon-height` to match the purpose.

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,114 @@
...
+$checkbox-check-top: 0.1875em !default;
+$checkbox-check-width: 0.3125em !default;
```
ollips, line 12: These three variable names should match and be either `$checkbox-icon-{rule}` or `$checkbox-check-{rule}` or `$checkbox-icon-check-{rule}`

ollips: Oh, same for `$checkbox-checkmark-thickness` - use either `icon` or `check` or checkmark` in all of them

css/src/components/checkbox.scss
homing1: Updated to `$checkbox-check-height`

css/src/components/checkbox.scss
homing1: Let's go with `$checkbox-check-*`, updated

ollips (approved):

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,114 @@
...
+
+$checkbox-border-radius: $radius-sm !default;
+$checkbox-border-width: 0.0625em !default;
+$checkbox-unchecked-border-color: $text !default;
+$checkbox-color: $primary !default;
+$checkbox-check-color: $primary-invert !default;
+$checkbox-check-hover-color: $text-subtle !default;
+$checkbox-size: 1.25em !default;
+$checkbox-check-height: 0.75em !default;
+$checkbox-check-top: 0.1875em !default;
```
wibjorn, line 11: To my eyes, checkbox top looks better as `.125em;` (more centered vertically). 3px seems to be a bit low

2px (.125em)
image

3px (.1875em)
image

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,114 @@
...
+$checkbox-check-width: 0.3125em !default;
+$checkbox-check-thickness: 0.0625em !default;
+$checkbox-timing-function: $input-timing-function !default;
+$checkbox-duration: $input-transition-duration !default;
+$checkbox-spacing: $spacer-3 !default;
+
+@mixin checkmark {
+	display: block;
+	position: absolute;
+	top: $checkbox-check-top;
```
wibjorn, line 21: This requirement sneakily slid in during this PR, but we want to do logical properties whenever possible, so `top` becomes `inset-block-start`.

css/src/components/checkbox.scss
homing1: Updated

css/src/components/checkbox.scss
homing1: Updated

wibjorn (approved):

wibjorn:
Long road! But great work! The markup is _much_ improved!

css/src/components/checkbox.scss
```diff
@@ -0,0 +1,114 @@
...
+			&::before {
+				@include checkmark;
+
+				border-color: $checkbox-check-color;
+			}
+		}
+
+		&[disabled] + .checkbox-check {
+			opacity: 0.6;
+			cursor: not-allowed;
```
jdanyow, line 87: future PR if folks agree:

Should `cursor: not-allowed` be on .checkbox?

jdanyow (approved):

css/src/components/checkbox.scss
ollips: I feel a bit weird about seeing different cursors on the checkbox and label. Or at least it shouldn't be a `pointer` on the label, the default arrow would be perfect!

ollips (approved):

PR #320, opened 2022-01-13
width atomics

wibjorn:
Task: task-546818

Link: preview-320
https://design.docs.microsoft.com/pulls/320/atomics/width.html

## Testing

View documentation page.

## Notes

Atomics to constrain set the `width` property on elements. Major decision points here:
- no min/max width rules
- `unset` provides strategry for responsively reset widths on tablet and larger.
- globally applicable widths do not exceed common mobile boundaries
- there are some widths that exist only on tablet +
- we'll use pixel widths based on 50s, because although they not be as transmutable as base 8, they're a hell of a lot easier to remember.
- widths go up to 450

### DIFF
--- a/css/src/atomics/width.scss
+++ b/css/src/atomics/width.scss
@@ -6,6 +6,10 @@ $all-widths: list.join($universal-widths, $mobile-incompatible-widths);
 
 // widths
 
+.width-full {
+	width: 100% !important;
+}
+
 @each $width in $universal-widths {
 	$unit: if($width != auto, 'px', '');
 	.width-#{$width} {
@@ -14,6 +18,10 @@ $all-widths: list.join($universal-widths, $mobile-incompatible-widths);
 }
 
 @include tablet {
+	.width-full-tablet {
+		width: 100% !important;
+	}
+
 	@each $width in $all-widths {
 		$unit: if($width != auto and $width != unset, 'px', '');
 		.width-#{$width}-tablet {
@@ -23,6 +31,10 @@ $all-widths: list.join($universal-widths, $mobile-incompatible-widths);
 }
 
 @include desktop {
+	.width-full-desktop {
+		width: 100% !important;
+	}
+
 	@each $width in $all-widths {
 		$unit: if($width != auto and $width != unset, 'px', '');
 		.width-#{$width}-desktop {
--- a/css/src/atomics/width.scss
+++ b/css/src/atomics/width.scss
@@ -1,37 +1,41 @@
 @use 'sass:list';
 
-$universal-widths: auto 100, 150, 200, 250, 300, 350 !default;
-$tablet-plus-widths: 400, 450, unset !default;
-$all-widths: list.join($universal-widths, $tablet-plus-widths);
+$universal-widths: auto, 100, 150, 200, 250, 300, 350 !default;
+$mobile-incompatible-widths: 400, 450, unset !default;
+$all-widths: list.join($universal-widths, $mobile-incompatible-widths);
 
 // widths
 
 @each $width in $universal-widths {
+	$unit: if($width != auto, 'px', '');
 	.width-#{$width} {
-		width: #{$width}px !important;
+		width: #{$width}#{$unit} !important;
 	}
 }
 
 @include tablet {
 	@each $width in $all-widths {
+		$unit: if($width != auto and $width != unset, 'px', '');
 		.width-#{$width}-tablet {
-			width: #{$width}px !important;
+			width: #{$width}#{$unit} !important;
 		}
 	}
 }
 
 @include desktop {
 	@each $width in $all-widths {
+		$unit: if($width != auto and $width != unset, 'px', '');
 		.width-#{$width}-desktop {
-			width: #{$width}px !important;
+			width: #{$width}#{$unit} !important;
 		}
 	}
 }
 
 @include widescreen {
 	@each $width in $all-widths {
+		$unit: if($width != auto and $width != unset, 'px', '');
 		.width-#{$width}-widescreen {
-			width: #{$width}px !important;
+			width: #{$width}#{$unit} !important;
 		}
 	}
 }
--- a/css/src/atomics/width.scss
+++ b/css/src/atomics/width.scss
@@ -1,19 +1,19 @@
-$widths: 100, 150, 200, 250, 300, 350, 400, 450, 500 !default;
-// do we need a maximum for globally applicable widths?
-// i.e. .width-500 on a 414px wide screen would be a problem
-// do we need min and max? They are used sometimes.
-// But we're definitely generating more of every category of atomic than we need.
+@use 'sass:list';
+
+$universal-widths: 100, 150, 200, 250, 300, 350 !default;
+$tablet-plus-widths: 400, 450 !default;
+$all-widths: list.join($universal-widths, $tablet-plus-widths);
 
 // widths
 
-@each $width in $widths {
+@each $width in $universal-widths {
 	.width-#{$width} {
 		width: #{$width}px !important;
 	}
 }
 
 @include tablet {
-	@each $width in $widths {
+	@each $width in $all-widths {
 		.width-#{$width}-tablet {
 			width: #{$width}px !important;
 		}
@@ -21,7 +21,7 @@ $widths: 100, 150, 200, 250, 300, 350, 400, 450, 500 !default;
 }
 
 @include desktop {
-	@each $width in $widths {
+	@each $width in $all-widths {
 		.width-#{$width}-desktop {
 			width: #{$width}px !important;
 		}
@@ -29,73 +29,9 @@ $widths: 100, 150, 200, 250, 300, 350, 400, 450, 500 !default;
 }
 
 @include widescreen {
-	@each $width in $widths {
+	@each $width in $all-widths {
 		.width-#{$width}-widescreen {
 			width: #{$width}px !important;
-		}
-	}
-}
-
-// max-widths
-
-@each $width in $widths {
-	.max-width-#{$width} {
-		max-width: #{$width}px !important;
-	}
-}
-
-@include tablet {
-	@each $width in $widths {
-		.max-width-#{$width}-tablet {
-			max-width: #{$width}px !important;
-		}
-	}
-}
-
-@include desktop {
-	@each $width in $widths {
-		.max-width-#{$width}-desktop {
-			max-width: #{$width}px !important;
-		}
-	}
-}
-
-@include widescreen {
-	@each $width in $widths {
-		.max-width-#{$width}-widescreen {
-			max-width: #{$width}px !important;
-		}
-	}
-}
-
-// min-widths
-
-@each $width in $widths {
-	.min-width-#{$width} {
-		min-width: #{$width}px !important;
-	}
-}
-
-@include tablet {
-	@each $width in $widths {
-		.min-width-#{$width}-tablet {
-			min-width: #{$width}px !important;
-		}
-	}
-}
-
-@include desktop {
-	@each $width in $widths {
-		.min-width-#{$width}-desktop {
-			min-width: #{$width}px !important;
-		}
-	}
-}
-
-@include widescreen {
-	@each $width in $widths {
-		.min-width-#{$width}-widescreen {
-			min-width: #{$width}px !important;
 		}
 	}
 }
--- a/css/src/atomics/width.scss
+++ b/css/src/atomics/width.scss
@@ -30,12 +30,3 @@ $all-widths: list.join($universal-widths, $mobile-incompatible-widths);
 		}
 	}
 }
-
-@include widescreen {
-	@each $width in $all-widths {
-		$unit: if($width != auto and $width != unset, 'px', '');
-		.width-#{$width}-widescreen {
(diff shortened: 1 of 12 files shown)


css/src/atomics/width.scss
```diff
@@ -0,0 +1,101 @@
...
+
+@include desktop {
+	@each $width in $widths {
+		.width-#{$width}-desktop {
+			width: #{$width}px !important;
+		}
+	}
+}
+
+@include widescreen {
```
ollips, line 31: I think all the `widescreen` rules can be removed. I doubt it would be any difference there knowing the widescreen breakpoint is 1800 px.

css/src/atomics/width.scss
```diff
@@ -0,0 +1,101 @@
+$widths: 100, 150, 200, 250, 300, 350, 400, 450, 500 !default;
```
ollips, line 1: I think `500` can be removed. At first, I couldn't find any existing case using that width. Secondly, not providing that option would eliminate the number of problems with mobile screens. Even though I understand some others still might be an issue, but at least just one less 🤔

ollips, line 1: Another thought - I think that would be the right place to add a `width-full` helper.

Oh, also it might be great to have `width-auto` 🤔

site/src/atomics/width.md
```diff
@@ -0,0 +1,175 @@
...
+	<p>450</p>
+</div>
+<div class="width-500 box-shadow-heavy padding-sm font-weight-semibold">
+	<p>500</p>
+</div>
+```
+
+### Responsive rules
+
+If there are screen-specific class variations available they can be described in this section.
```
ollips, line 53: Update the description 😁

site/src/atomics/width.md
```diff
@@ -0,0 +1,175 @@
...
+
+```html
+<div class="width-200 width-400-tablet box-shadow-heavy padding-sm font-weight-semibold">
+	<p>300 ➡ 500</p>
```
ollips, line 57: Per applied classes it should be `200 ➡ 400`

css/src/atomics/width.scss
wibjorn: perhaps we should eliminate 450 too?

wibjorn: yes, good idea

site/src/atomics/width.md
wibjorn: 🤦‍♂️ thanks!

css/src/atomics/width.scss
```diff
@@ -0,0 +1,101 @@
...
+
+@include widescreen {
+	@each $width in $widths {
+		.width-#{$width}-widescreen {
+			width: #{$width}px !important;
+		}
+	}
+}
+
+// max-widths
```
jdanyow, line 39: Adding a note related to what we talked about on teams- would be good to know how folks are using max-width and min-width and whether we want folks to continue using those approaches.

I think the documentation in this PR describes what is available but needs examples around how these classes should be used based on the existing usages in docs-ui.

css/src/atomics/width.scss
homing1: Perhaps 500 can be added once there is a use case. I see that we're currently using 450 for the chat window in LearnTV so we might want to keep that.
image

css/src/atomics/width.scss
homing1: We do have a use case of widescreen for the chat window in LearnTV,

css/src/atomics/width.scss
```diff
@@ -0,0 +1,101 @@
...
+	}
+}
+
+@include widescreen {
+	@each $width in $widths {
+		.min-width-#{$width}-widescreen {
+			min-width: #{$width}px !important;
+		}
+	}
+}
```
jdanyow, line 101: Is the output size of the css reduced by restructuring the code as follows? Also thinking this could be better in terms of ordering for responsive use?

Having looked through the docs-ui code using `has(-min|-max)?-width-.*`, many of the min-width usages could use width and many of the max-width usages have to do with truncation and could use flex grow/shrink to control width of the element being truncated.

```scss
@each $width in $widths {
  // width loop
  // min-width loop
  // max-width loop    
}

@include tablet {
  // width loop
  // min-width loop
  // max-width loop    
}

@include desktop {
  // width loop
  // min-width loop
  // max-width loop    
}

@include widescreen {
  // width loop
  // min-width loop
  // max-width loop    
}
```

css/src/atomics/width.scss
wibjorn: To me that particular instance could be either removed or changed to a slightly lower value or only applied on desktop.

css/src/atomics/width.scss
wibjorn: As said in other thread, re:LearnTV example it seems to me that particular instance could be either removed or changed to a slightly lower value or only applied on desktop.

site/src/atomics/width.md
wibjorn: 😱

css/src/atomics/width.scss
wibjorn: Team discussed in a meeting. Think we landed on the following:

-  no min width rules, they tend not to be necessary
- need a good strategy for unsetting global widths on tablet +. The width equivelent to something like `display-none display-none-tablet`. This strategry could be based on `unset`, `auto`, or `inherit` but need to do some research to figure out what's best.
- we'll need to be careful that globally applicable widths do no exceed common mobile boundaries, or that wider ones are scoped to tablet. This could be a confusing an unexpected thing. First consideration is to preclude mobile with bugs where `.width-400` for example exceed an iPhone SE's (or whatever small-ish or old phone's) horizontal boundaries.

css/src/atomics/width.scss
ollips: I would recommend changing the applied class on LearntTV to `-desktop` and don't have any `widescreen` helpers.

css/src/atomics/width.scss
```diff
@@ -0,0 +1,37 @@
+@use 'sass:list';
+
+$universal-widths: auto 100, 150, 200, 250, 300, 350 !default;
```
ollips, line 3: Comma is missed after `auto`

ollips: Also, probably it shouldn't be here. Per rules below we end up with `width: autopx !important`. Add .`width-auto` directly after every `@each` rule set.

site/src/atomics/width.md
```diff
@@ -0,0 +1,90 @@
...
+
+# Width Atomics
+
+Need to specifiy a specific width for an element? You've come to the right place! These classes should be used sparingly, and only used after testing heavily at all screen widths (starting with mobile). That said, if you need to specify and min, max or normal width, they can be a useful tool in the utility belt.
+
+Values are in pixels. Keep in mind that the `min-width` and `max-width` properties override the standard `width` property. If you aren't using the Atlas core folder as a base for you styles, these classes work best when [you're setting `box-sizing` to be `border-box` in some way](https://css-tricks.com/box-sizing/#aa-present-day-box-sizing).
+
+| cssproperty                                      | value (px)                               | screensize                                         |
+| ------------------------------------------------ | ---------------------------------------- | -------------------------------------------------- |
+| `width`                                          | `100`, `150`, `200`, `250`, `300`, `350` | all screensizes, `tablet`, `desktop`, `widescreen` |
```
ollips, line 15: `auto` is missed here as a possible value as well.

site/src/atomics/width.md
```diff
@@ -0,0 +1,90 @@
...
+	<p>300</p>
+</div>
+<div class="width-350 border padding-sm font-weight-semibold">
+	<p>350</p>
+</div>
+```
+
+### Responsive rules
+
+Certain widths are too large for many mobile screens, so they are only available on larger screensizes. However, all helpers can be applied to any screensize.
```
ollips, line 45: _However, all helpers can be applied to any screen size._ - I understand what you meant here, however, I think it might sound a little bit confusing at a first glance, cause not _all_ helpers can be applied to _any_ screen size. For example, I might think about `width-400` as to be available on all screen sizes including mobile but it doesn't exist without `{screensize}` prefix actually.

site/src/atomics/width.md
```diff
@@ -0,0 +1,90 @@
...
+	<p>250 ➡ 300 ➡ 350 ➡ 450</p>
+</div>
+```
+
+## Available classes
+
+List of all available classes:
+
+```atomics-filter
+.width-100
```
ollips, line 60: Please add `width-auto` helpers here as well

site/src/atomics/width.md
```diff
@@ -0,0 +1,90 @@
...
+title: Width Atomics
+description: Width related atomic css classes for the Atlas Design System
+template: standard
+---
+
+# Width Atomics
+
+Need to specifiy a specific width for an element? You've come to the right place! These classes should be used sparingly, and only used after testing heavily at all screen widths (starting with mobile). That said, if you need to specify and min, max or normal width, they can be a useful tool in the utility belt.
+
+Values are in pixels. Keep in mind that the `min-width` and `max-width` properties override the standard `width` property. If you aren't using the Atlas core folder as a base for you styles, these classes work best when [you're setting `box-sizing` to be `border-box` in some way](https://css-tricks.com/box-sizing/#aa-present-day-box-sizing).
```
homing1, line 11: "a base for your styles, these..."

css/src/atomics/width.scss
wibjorn: Sounds like a good plan to me

wibjorn: widescreen is gone

ollips: 🙌🏻🙌🏻🙌🏻

css/src/atomics/width.scss
ollips: Was a WI created to make that update in docs? Resolve if it was or after creating one!

ollips: What about the full width class? I do see quite a lot of usage of similar class in docs

site/src/atomics/width.md
```diff
@@ -0,0 +1,109 @@
...
+</div>
+```
+
+### Responsive rules
+
+Certain widths are too large for many mobile screens, so they are only available on larger screensizes. However, all helpers can be applied to any screensize.
+
+```html
+<div class="width-250 width-300-tablet width-400-desktop border padding-sm">
+	<p>250 ➡ 300 ➡ 350 ➡ 450</p>
```
ollips, line 63: Remove `➡ 350`, update `450` to `400` to match what's actually applied.

ollips (changes requested):

css/src/atomics/width.scss
wibjorn: done!

wibjorn: Might be better in the future to come up with a .column friendly scheme for percentage widths, but for now, pushing up only `.width-full` and the same for tablet and desktop.

site/src/atomics/width.md
wibjorn: this is gone now

site/src/atomics/width.md
wibjorn: re-written

site/src/atomics/width.md
wibjorn: 🤦‍♂️gets me every time!

css/src/atomics/width.scss
ollips: Don't forget to add it to the doc page!

site/src/atomics/width.md
ollips: I still see the same thing.

site/src/atomics/width.md
wibjorn: updated

site/src/atomics/width.md
wibjorn: 🤦‍♂️Apologies @ollips, yes, I see it here now. Your point is definitely correct, this sentence is misleading.

site/src/atomics/width.md
wibjorn: I lied, now they it's re-written

site/src/atomics/width.md
```diff
@@ -0,0 +1,112 @@
...
+
+Values are in pixels. Keep in mind that the `min-width` and `max-width` properties override the standard `width` property. If you aren't using the Atlas core folder as a base for your styles, these classes work best when [you're setting `box-sizing` to be `border-box` in some way](https://css-tricks.com/box-sizing/#aa-present-day-box-sizing). Values represent pixel unless they are a string or otherwise specified.
+
+| cssproperty                                      | value                                                    | screensize                            |
+| ------------------------------------------------ | -------------------------------------------------------- | ------------------------------------- |
+| `width`                                          | `auto`, `100`, `150`, `200`, `250`, `300`. `full` (100%) | all screensizes, `tablet`, `desktop`, |
+| `width` (available on tablet screens and larger) | `350`, `400`, `450`, `unset`                             | `tablet`, `desktop`,                  |
```
ollips, line 16: `350` is universal width class

ollips (approved):

jdanyow (approved):

PR #326, opened 2022-01-19
Adding an additional popover closing method

ollips:
Task: task-[work-item-number]

Link: preview-326

Adding the `data-close` attribute to the popover's main content link/button would collapse it after a click.

## Testing

1. Step one
2. Step two
   Expected result:

## Additional information

[Optional]

### DIFF
--- a/site/src/components/popover.md
+++ b/site/src/components/popover.md
@@ -71,3 +71,26 @@ Popover is left aligned by default, but also can be centered or right aligned re
 	<div class="popover-content border">Popover content.</div>
 </details>
 ```
+
+## Closing popover
+
+An additional way to collapse the popover is to add the `data-close` attribute to the link/button in the content.
+
+```html
+<details class="popover margin-xxs">
+	<summary class="button">Popover</summary>
+	<div class="popover-content border">
+		<ul>
+			<li class="margin-bottom-xxs">
+				<a href="#">Link</a>
+			</li>
+			<li class="margin-bottom-xxs">
+				<a href="#">Link</a>
+			</li>
+			<li>
+				<a href="#" data-close>Closing link</a>
+			</li>
+		</ul>
+	</div>
+</details>
+```
--- a/site/src/components/popover.md
+++ b/site/src/components/popover.md
@@ -71,26 +71,3 @@ Popover is left aligned by default, but also can be centered or right aligned re
 	<div class="popover-content border">Popover content.</div>
 </details>
 ```
-
-## Closing popover
-
-An additional way to collapse the popover is to add the `data-close` attribute to the link/button in the content.
-
-```html
-<details class="popover margin-xxs">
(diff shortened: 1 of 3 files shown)


site/src/components/popover.md
```diff
@@ -71,3 +71,26 @@ Popover is left aligned by default, but also can be centered or right aligned re
 	<div class="popover-content border">Popover content.</div>
 </details>
 ```
+
+## Closing popover
+
+An additional way to collapse the popover is to add the `data-close` attribute to the link/button in the content.
```
wibjorn, line 77: I think we should name scope this to popovers `data-popover-close`?

site/src/components/popover.md
```diff
@@ -71,3 +71,26 @@ Popover is left aligned by default, but also can be centered or right aligned re
...
+	<div class="popover-content border">
+		<ul>
+			<li class="margin-bottom-xxs">
+				<a href="#">Link</a>
+			</li>
+			<li class="margin-bottom-xxs">
+				<a href="#">Link</a>
+			</li>
+			<li>
+				<a href="#" data-close>Closing link</a>
```
wibjorn, line 91: Perhaps the example should be a button?

site/src/components/popover.md
```diff
@@ -71,3 +71,26 @@ Popover is left aligned by default, but also can be centered or right aligned re
 	<div class="popover-content border">Popover content.</div>
 </details>
 ```
+
```
wibjorn, line 74: Wondering if we can omit this documentation until or re-write it to be a suggested implementation. We don't yet have a way of sharing the front end scripts with other consumers (we're documenting TS that is not technically part of our css package)

site/src/scaffold/scripts/popover.ts
```diff
@@ -28,6 +28,10 @@ export function initPopovers(container: HTMLElement) {
 				if (!targetPopover?.contains(event.target)) {
 					closePopovers();
 				}
+
+				if (event.type === 'click' && event.target.hasAttribute('data-close')) {
```
wibjorn, line 32: `data-popover-close`?

site/src/components/popover.md
ollips: Updated

site/src/components/popover.md
ollips: Oh, you're right, I always forget we don't share the scripts right now. Decided to omit this portion of documentation until we do!

site/src/components/popover.md
ollips: Temporarily removed the whole example block

site/src/scaffold/scripts/popover.ts
ollips: Updated

wibjorn (approved):

homing1 (approved):

PR #343, opened 2022-02-03
Input & Textarea components

ollips:
Task: task-531896

Link: preview-343

Adding input and textarea components.

## Testing

- [Input](https://design.docs.microsoft.com/pulls/343/components/input.html)
- [Textarea](https://design.docs.microsoft.com/pulls/343/components/textarea.html)
- [Label](https://design.docs.microsoft.com/pulls/343/components/label.html)
- [Help](https://design.docs.microsoft.com/pulls/343/components/help.html)

1. Visit the pages linked above.
2. Click through the input/textarea components, ensure different states are looking good.


css/src/components/form/input.scss
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	&.input-lg {
+		@include control-lg;
+	}
+
+	&.input-narrow {
+		width: auto;
+
+		&-tablet {
```
wibjorn, line 61: we already have `width-auto` and `width-auto-tablet`. Remove?

css/src/components/form/input.scss
```diff
@@ -0,0 +1,91 @@
...
+			@include tablet {
+				width: auto;
+			}
+		}
+	}
+
+	&.input-error {
+		border-color: $input-error-border-color;
+
+		&:focus-visible {
```
wibjorn, line 71: Need to determine the state of :focus-visible, is it implemented in all browsers in our matrix (Safari was last holdout that I know of). Remember that [we have logic](https://github.com/microsoft/atlas-design/blob/11f26a3e691344d61aff657e80b5d55182c49896/css/src/core/focus.scss#L1) to configure whether we use .focus-visible or :focus-visible. Docs uses the former.

wibjorn: Consider as well whether you definitely require a separate class, or whether the rule can just be:

`&.&-error {
		border-color: $input-error-border-color;
        outline-color: $input-error-border-color;
}

css/src/components/form/textarea.scss
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	&[disabled] {
+		border-color: $textarea-disabled-border-color;
+		background-color: $textarea-disabled-background-color;
+		color: $textarea-disabled-color;
+	}
+
+	&:not([rows]) {
+		min-height: 120px;
```
wibjorn, line 53: Research question: is there a more modern way to do this?

css/src/components/form/textarea.scss
```diff
@@ -0,0 +1,106 @@
...
+
+		&-tablet {
+			@include tablet {
+				width: auto;
+			}
+		}
+	}
+
+	&.textarea-error {
+		border-color: $textarea-error-border-color;
```
wibjorn, line 84: same focus and outline related note as in input

site/src/components/input.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+## Validation states
+
+`input-error`/`input-warning`/`input-success` helps with visual reflection of validation states.
+
+| State   | Class                   |                                                                             |
+| ------- | ----------------------- | --------------------------------------------------------------------------- |
+| Error   | `.input .input-error`   | <input class="input input-error" type="text" placeholder="Placeholder" />   |
+| Warning | `.input .input-warning` | <input class="input input-warning" type="text" placeholder="Placeholder" /> |
```
wibjorn, line 58: .

site/src/components/input.md
```diff
@@ -0,0 +1,63 @@
...
+| Success | `.input .input-success` | <input class="input input-success" type="text" placeholder="Placeholder" /> |
+
+```abut-html
+<input class="input .input-error" type="text" placeholder="Placeholder" />
```
wibjorn, line 62: `.input`

css/src/components/form/input.scss
```diff
@@ -0,0 +1,91 @@
...
+$input-disabled-border-color: $table-border-dark !default;
+
+$input-error-border-color: $danger !default;
+$input-warning-border-color: $warning-active !default;
+$input-success-border-color: $score-high !default;
+
+.input {
+	@include control;
+
+	width: 100%;
```
wibjorn, line 21: I do not think we'll want to add any width related rules to input

wibjorn, line 21: This is something I never noticed about input, but seems like `input.width-full` is much more normal to me than `input.input-narrow`. Naturally this change would would make migration quite a bit trickier, but honestly, a full width input seems like a strange move to me.

site/src/components/input.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Input
+description: The input component in the Atlas Design System
+template: standard
+figmaEmbed: https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FuVA2amRR71yJZ0GS6RI6zL%2F%25F0%259F%258C%259E-Atlas-Design-Library%3Fnode-id%3D506%253A1176
+---
+
+# Input component
```
wibjorn, line 8: Most H1s are just the name - only "Input" here, not "Input component". Think the shorter version is better. Same problem with `textarea`

css/src/components/form/textarea.scss
ollips: Updated

site/src/components/input.md
ollips: Fixed

css/src/components/form/input.scss
ollips: Hmm, I'll take a look into docs and how things are done there to confirm the approach. If I'm not mistaken I saw only one instance of `.input.input-narrow`, that way it would make more sense to keep it as is I think, but need to double check.

site/src/components/input.md
ollips: Updated

css/src/components/form/input.scss
ollips: Partially updated. Still need to test the `:focus-visible` in different browsers.

css/src/components/form/input.scss
ollips: Removed `-narrow` modifiers in favor of `width-auto` usage 👍🏻

css/src/components/form/input.scss
wibjorn: https://caniuse.com/?search=%3Afocus-visible

Looks like we're still missing Safari on iOS

wibjorn: Here's an idea:

```
// /mixins/focus-visible.scss

@mixin focus-visible() {
	@if ($focus-visible-use-polyfill) {
		&.focus-visible {
			@content;
		}
	} @else {
		&:focus-visible {
			@content;
		}
	}
}

// elsewhere
.input {
    @include focus-visible {
	background-color: black !important;
    }
}

File: css/src/components/form/help.scss [hunk 9]
```diff
@@ -0,0 +1,35 @@
...
+
+	&.help-warning {
+		color: $help-warning-color;
+	}
+
+	&.help-success {
+		color: $help-success-color;
+	}
+
+	&.help-sm {
```
wibjorn, line 28: remove `-sm`/`-lg` in favor of `.font-size-` atomics?

ollips: I see your point here, however, I think I would prefer to keep these modifiers. Here is why - initially help's font size is smaller compared to the input/textarea components, so I think it would be much easier to have these two modifiers in favor of input's/textarea's size modifiers so the user doesn't have to guess which exact `.font-size-` class to use to keep the proportions between the form's interactive component and the help text. Let me know if that makes sense to you.

css/src/components/form/input.scss
ollips: I checked docs and there are just a few (3 to be specific) places where input's width was reset to auto so I think it makes sense to keep it full width by default. In my experience that's exactly what I see everywhere, that way form looks more organized rather than when fields have different (or just very small) widths or it is a lot of white space around. Idk, I can collect more opinions on this if you feel strongly about it.

wibjorn: Okay, sounds good to me

css/src/components/form/help.scss
wibjorn: I think ideally we just provide a single default and that's what's used most of the time. The .help-<size> modifiers seem extraneous to me. I also don't think we'd need to document .font-size-<size> usage as it relates to this.

ollips: Updated :)

css/src/components/form/input.scss
```diff
@@ -0,0 +1,102 @@
...
+			outline: none !important;
+			box-shadow: $input-focus-box-shadow-size $input-focus-border-color;
+		} @else {
+			@extend %focus;
+
+			outline-color: $input-focus-border-color;
+			outline-offset: 0 !important;
+			outline-style: solid !important;
+		}
+	}
```
ollips, line 57: @wibjorn I need your advice. I have two ideas on how to proceed with focus styles since we want to keep Safari looking nice. This is a first approach that keeps `focus-visible` mixin unchanged, however, I need to use the `if ... else` statement here to pass the correct CSS rules I want to be applied to the component. 

Th second approach I was thinking of is modifying the mixin like that:
```css
@mixin focus-visible($focus-color: $primary, $shadow-size: 0 0 0 0.125rem) {
	@if ($focus-visible-use-polyfill) {
		&.focus-visible {
			@content;
			outline: none !important;
			box-shadow: $shadow-size $focus-color;
		}
	} @else {
		&:focus-visible {
			@extend %focus;

			@content;
		}
	}
}
```

and then the in `input.scss` file code would look like that:

```css
@include focus-visible($input-focus-border-color, $input-focus-box-shadow-size) {
	border-color: $input-focus-border-color;
	outline-color: $input-focus-border-color;
	outline-offset: 0 !important;
	outline-style: solid !important;
}
```

Personally, I like the 2nd better, but I'm not sure if the box-shadow styles would look good on future components that might need using that mixin, including the `toggle` you're working on. Please let me know what you think.

ollips: To emulate `focus-visible` styles like in Safari follow these steps:
1. Update `$focus-visible-use-polyfill: false` to be `true` in `tokens/focus.scss` file
2. `npm start`
3. Manually add `focus-visible` class to the `input`/`textarea` components to see how it look like

wibjorn: chatted offline online, figured this out

css/src/components/form/textarea.scss
```diff
@@ -0,0 +1,97 @@
...
+		border-color: $textarea-focus-border-color;
+		outline: none !important;
+		box-shadow: $textarea-focus-box-shadow-size $textarea-focus-border-color;
+	}
+
+	&.textarea-fixed-height {
+		resize: none;
+	}
+
+	&.textarea-sm {
```
wibjorn, line 65: Can we remove these?

site/src/components/textarea.md
```diff
@@ -0,0 +1,57 @@
...
+<label class="label" for="textarea-demo">Label</label>
+<textarea class="textarea" id="textarea-demo" placeholder="Textarea placeholder"></textarea>
+```
+
+### Sizes
+
+`textarea-sm`/`textarea-lg` makes textarea a little bigger or smaller than the default.
+
+```html
+<textarea class="textarea textarea-sm" placeholder="Textarea placeholder"></textarea>
```
wibjorn, line 34: rm sizes?

css/src/components/form/help.scss
```diff
@@ -0,0 +1,24 @@
+$help-color: $text !default;
+$help-error-color: $danger !default;
+$help-warning-color: $warning-active !default;
+$help-success-color: $score-high !default;
+$help-font-size: $font-size-8 !default;
+
+.help {
+	display: block;
+	margin-block-start: 0.5rem;
```
wibjorn, line 9: I don't think we should include any margin on elements by default. We can use atomics to compose experiences.

css/src/components/form/input.scss
```diff
@@ -0,0 +1,84 @@
...
+
+.input {
+	@include control;
+
+	width: 100%;
+	max-width: 100%;
+	border-color: $input-border-color;
+	background-color: $input-background-color;
+	color: $input-color !important;
+	margin-block-start: 0.5rem;
```
wibjorn, line 28: don't think we require/should have margin here

css/src/components/form/input.scss
```diff
@@ -0,0 +1,84 @@
...
+
+$input-disabled-color: $text-subtle !default;
+$input-disabled-background-color: $body-background-medium !default;
+$input-disabled-border-color: $table-border-dark !default;
+
+$input-error-border-color: $danger !default;
+$input-warning-border-color: $warning-active !default;
+$input-success-border-color: $score-high !default;
+
+$input-focus-box-shadow-size: 0 0 0 0.125rem !default;
```
wibjorn, line 18: should the box shadow size be tied to the elemnts font size? `0.125em`?

css/src/components/form/textarea.scss
```diff
@@ -0,0 +1,97 @@
...
+	@include control;
+
+	display: block;
+	width: 100%;
+	max-width: 100%;
+	resize: vertical;
+	border-color: $textarea-border-color;
+	background-color: $textarea-background-color;
+	color: $textarea-color !important;
+	margin-block-start: 0.5rem;
```
wibjorn, line 30: rm margin here too

css/src/components/form/help.scss
ollips: I think we should provide some default spacing on it. It is meant to be used in combination with some other component, so having some spacing would be needed always. What's the point of figuring out later what additional atomics to use when we can prevent it? 🤔

css/src/components/form/input.scss
ollips: I don't really see a lot of benefits from it tbh. Also, the outline rules we have set up in `rem` too, so I would prefer to keep the same way for consistency.

css/src/components/form/input.scss
ollips: I added it here thinking about how the input comes with a label or the case when a bunch of form components comes one after another. But I can definitely see the cases when it might be not needed, so will remove it. However, I think the label component needs to have it by default instead.

css/src/components/form/help.scss
wibjorn: A general principal of all our components is that they have no margin sizing by themselves. This allows for easy composability in a variety of scenarios.

css/src/components/form/input.scss
wibjorn: okay, cool, thanks for testing!

css/src/components/form/textarea.scss
ollips: Yes! Removed :)

site/src/components/textarea.md
ollips: Removed

css/src/components/form/help.scss
ollips: Removed

css/src/components/form/input.scss
ollips: Removed

css/src/components/form/textarea.scss
ollips: Removed

css/src/components/form/help.scss
```diff
@@ -0,0 +1,23 @@
+$help-color: $text !default;
+$help-error-color: $danger !default;
+$help-warning-color: $warning-active !default;
+$help-success-color: $score-high !default;
```
wibjorn, line 4: $success, or $success-dark?

css/src/components/form/help.scss
```diff
@@ -0,0 +1,23 @@
+$help-color: $text !default;
+$help-error-color: $danger !default;
```
wibjorn, line 2: Is `danger` definitely meeting contrast reqs when paired with a medium background?

css/src/components/form/help.scss
```diff
@@ -0,0 +1,23 @@
...
+$help-warning-color: $warning-active !default;
```
wibjorn, line 3: `warning-dark` will have to do here, don't think we should pair it with a different concept

css/src/components/form/textarea.scss
```diff
@@ -0,0 +1,88 @@
...
+$textarea-hover-color: $text !default;
+$textarea-hover-border-color: $primary !default;
+$textarea-focus-border-color: $primary !default;
+
+$textarea-disabled-color: $text-subtle !default;
+$textarea-disabled-background-color: $body-background-medium !default;
+$textarea-disabled-border-color: $table-border-dark !default;
+
+$textarea-error-border-color: $danger !default;
+$textarea-warning-border-color: $warning-active !default;
```
wibjorn, line 15: same comments as related to help, these color concepts need to match more closely with their names

ollips: updated

css/src/components/form/help.scss
ollips: updated

css/src/components/form/help.scss
ollips: updated

css/src/components/form/help.scss
```diff
@@ -0,0 +1,23 @@
+$help-color: $text !default;
+$help-error-color: $danger-dark !default;
+$help-warning-color: $warning-dark !default;
```
homing1, line 3: Should the warning color be a shade of yellow? On light theme, the warning color looks like a shade of gray to me 
image

ollips: I see your point, however, we need to be able to pass the contrast accessibility tests, and the regular `$warning`, which is a brighter yellow color, doesn't meet the requirements.

homing1: Also, shouldn't error be a shade of red? Same for input

homing1: I see. I tried playing with the colors a bit. It looks to me that `$danger` should work with medium background in all themes except for `dark`. Please double check, but I think it would work if we update the dark `$danger-base` to be one shade lighter (`$palette-red-20`).

I'm just thinking if `$danger-base` or any state colors provide insufficient contrast, then perhaps we should update them so that we can actually use those variables. Otherwise, it means `$danger-base` can't realistically be used in most (if not all) cases and making it available to use seems misleading.

image

ollips: I followed exact same thoughts as you are! The only concern I had about it is how then the `danger` color would look a bit off compared to `warning` and `success` right next to it:

image

But probably that's not _that_ critical. @wibjorn what do you think? Can we update the color to make it accessible on dark mode with background medium?

css/src/components/form/help.scss
```diff
@@ -0,0 +1,23 @@
+$help-color: $text !default;
+$help-error-color: $danger-dark !default;
```
wibjorn, line 2: change name to `danger` from error

css/src/components/form/help.scss
wibjorn: It's a shame that danger-base on dark doesn't work as a text color. We could consider updating the `danger-base` property on dark to be
```


 But short of revisiting the palette now, I think using the dark variables will work just fine. They were created for this very purpose.

Comment (wibjorn) [reply css/src/components/form/help.scss:3]: We'd then end update with danger-base, success-base, and warning-dark. Which would be consistent with button.

File: css/src/components/form/help.scss [hunk 23]
```diff
@@ -0,0 +1,23 @@
...
+$help-warning-color: $warning-dark !default;
+$help-success-color: $success !default;
+$help-font-size: $font-size-8 !default;
+
+.help {
+	display: block;
+	color: $help-color;
+	font-size: $help-font-size;
+
+	&.help-error {
```
wibjorn, line 12: Any instances of error in this PR should be replaced with danger

css/src/components/form/input.scss
```diff
@@ -0,0 +1,83 @@
...
+
+	&.input-sm {
+		@include control-sm;
+	}
+
+	&.input-lg {
+		@include control-lg;
+	}
+
+	&.input-error {
```
wibjorn, line 57: here too, no new semantic color names

site/src/components/help.md
```diff
@@ -0,0 +1,37 @@
...
+<p class="help margin-top-xxs">Help message</p>
+```
+
+### Validation states
+
+`help-error`/`help-warning`/`help-success` helps with visual reflection of validation states.
+
+| State   | Class                 | Preview                                                          |
+| ------- | --------------------- | ---------------------------------------------------------------- |
+| Error   | `.help .help-error`   | <p class="help help-error margin-top-none">Error message</p>     |
```
wibjorn, line 27: +1

site/src/components/textarea.md
```diff
@@ -0,0 +1,49 @@
...
+<textarea class="textarea textarea-fixed-height" placeholder="Textarea placeholder"></textarea>
+```
+
+### Validation states
+
+`textarea-error`/`textarea-warning`/`textarea-success` helps with visual reflection of validation states.
+
+| State   | Class                         |                                                                                        |
+| ------- | ----------------------------- | -------------------------------------------------------------------------------------- |
+| Error   | `.textarea .textarea-error`   | <textarea class="textarea textarea-error" placeholder="Default textarea"></textarea>   |
```
wibjorn, line 43: +1

css/src/components/form/help.scss
ollips: Updated the color value and made all relevant updates in PR. Thanks @homing1 for bringing this up 🙏🏻

css/src/components/form/help.scss
ollips: Updated

css/src/components/form/input.scss
ollips: Updated

css/src/components/form/textarea.scss
```diff
@@ -0,0 +1,88 @@
...
+		border-color: $textarea-focus-border-color;
+		outline: none !important;
+		box-shadow: $textarea-focus-box-shadow-size $textarea-focus-border-color;
+	}
+
+	&.textarea-fixed-height {
+		resize: none;
+	}
+
+	&.textarea-error {
```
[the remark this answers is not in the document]
ollips: Updated

site/src/components/help.md
ollips: Updated. Also updated the state from "Error" to "Danger"

site/src/components/input.md
```diff
@@ -0,0 +1,51 @@
...
+<input class="input input-lg margin-top-xs" type="text" placeholder="Placeholder" />
+```
+
+### Validation states
+
+`input-error`/`input-warning`/`input-success` helps with visual reflection of validation states.
+
+| State   | Class                   |                                                                             |
+| ------- | ----------------------- | --------------------------------------------------------------------------- |
+| Error   | `.input .input-error`   | <input class="input input-error" type="text" placeholder="Placeholder" />   |
```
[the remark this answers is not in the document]
ollips: Updated

site/src/components/textarea.md
ollips: Updated

css/src/components/form/help.scss
homing1: Looks much better to me! For `warning`, `warning-dark` would work for now, but I think we need to revisit. I just realized that even for buttons, the current `$warning-base` is too light without an outline.

image

homing1 (dismissed):

wibjorn (approved):

PR #346, opened 2022-02-07
add toggle element and relevant documentation

wibjorn:
Task: task-533240

Link: preview-346

https://design.docs.microsoft.com/pulls/346/components/toggle.html

## Testing

1. Visit above page.
2. Toggle element state on/off.
3. Resize window, toggle element states.
4. Tab through various examples, ensure focus styles show.

## Additional information

[https://www.figma.com/file/uVA2amRR71yJZ0GS6RI6zL/%F0%9F%8C%9E-Atlas-Master-Styles-Library?node-id=838%3A791](https://www.figma.com/file/uVA2amRR71yJZ0GS6RI6zL/%F0%9F%8C%9E-Atlas-Master-Styles-Library?node-id=838:791)

### DIFF
--- a/css/src/components/toggle.scss
+++ b/css/src/components/toggle.scss
@@ -21,6 +21,7 @@ $toggle-circle-selected-background-color: $body-background !default;
 		border-radius: $radius-rounded;
 		background-color: $toggle-unselected-background-color;
 		cursor: pointer;
+		forced-color-adjust: none;
 
 		&::after {
 			display: block;
--- a/css/src/components/toggle.scss
+++ b/css/src/components/toggle.scss
@@ -1,4 +1,4 @@
-$toggle-transition: 200ms cubic-bezier(0.01, 1.23, 0.58, 0.96) !default;
+$toggle-transition: $input-transition-duration cubic-bezier(0.01, 1.23, 0.58, 0.96) !default;
 $toggle-circle-size: 0.75em !default;
 $toggle-circle-top: 0.1875em !default;
 $toggle-circle-distance-from-border: 1em - $toggle-circle-size;
@@ -23,8 +23,8 @@ $toggle-circle-selected-background-color: $body-background !default;
 		cursor: pointer;
 
 		@include forced-colors {
-			background-color: Canvas !important;
 			border-color: CanvasText !important;
+			background-color: Canvas !important;
 		}
 
 		&::after {
--- a/css/src/components/toggle.scss
+++ b/css/src/components/toggle.scss
@@ -1,4 +1,4 @@
-$toggle-transition: 1s cubic-bezier(0.01, 1.23, 0.58, 0.96) !default;
+$toggle-transition: 200ms cubic-bezier(0.01, 1.23, 0.58, 0.96) !default;
 $toggle-circle-size: 0.75em !default;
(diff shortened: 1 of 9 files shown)


site/src/components/toggle.md
```diff
@@ -0,0 +1,39 @@
...
+---
+
+# Toggle
+
+A selectable form element that represents a binary state (on or off).
+
+```html
+<div class="toggle">
+	<input id="toggle-id-0" type="checkbox" aria-label="Toggle me!" />
+	<label for="toggle-id-0"></label>
```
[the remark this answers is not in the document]
wibjorn: The label is the toggle element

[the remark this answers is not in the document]
ollips: 🤦🏻‍♀️🤦🏻‍♀️🤦🏻‍♀️

site/src/components/toggle.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Disabled states
+
+In order to achieve disabled styles, you'll need to add the `disabled` attribute to the inner input.
+
+### Disabled and unselected
+
+```html
+<div class="toggle">
+	<input id="toggle-id-2" type="checkbox" aria-label="Disabled toggle" disabled />
```
homing1, line 27: supernit: Should the id be `toggle-id-1` here to avoid duplicate ids?

css/src/components/toggle.scss
```diff
@@ -0,0 +1,86 @@
...
+	&.is-selected label,
+	input[type='checkbox']:checked + label {
+		/* stylelint-enable */
+		border-color: $toggle-selected-background-color;
+		background-color: $toggle-selected-background-color;
+
+		@include forced-colors {
+			border-color: LinkText !important;
+			background-color: LinkText !important;
+		}
```
BenDMyers, line 59: I might be wrong, but I think the more "semantic" system colors to use here would be the `Button` ones. Per https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#system_colors,

> **ButtonBorder**
> Base border color of controls
>
> **ButtonFace**
> Background color of controls
>
> **ButtonText**
> Foreground color of controls

The generic "controls" phrasing makes me think that's what we should be using for this. Thoughts?

wibjorn: Could be that's right, was thinking that I like that the enabled version of the toggle to be more actively colored. It looks liek this if we use ButtonText. 

image

wibjorn: Compared to current in PR: 
image

BenDMyers: What about using `ButtonFace`? It's for "Background color of controls," so that might be more colorful.

wibjorn:
@homing1 work item created for that, out of scope for this PR I think.

css/src/components/toggle.scss
wibjorn: Button Face is black in this particular case.

ollips (approved):

css/src/components/toggle.scss
BenDMyers: Gotcha - then I'm good with how we have it!

BenDMyers (approved):

PR #350, opened 2022-02-11
logo update

wibjorn:
We're not properly following brand guidelines having a monochromatic logo. Updating to follow suite with docs.microsoft.com.

preview-350


jdanyow (approved):
To fully comply I think you may need to create a theme color like "--microsoft-logo-text" or something because the word microsoft should be #fff in dark and hc themes.

homing1 (approved):

PR #355, opened 2022-02-24
Fixing select component size variations

ollips:
Task: task-531897

Link: preview-355

Fixing the select component sizes.

## Testing

1. Check the code
2. Visit [Select page](https://design.docs.microsoft.com/pulls/355/components/select.html)

### DIFF
--- a/.changeset/yellow-suits-suffer.md
+++ b/.changeset/yellow-suits-suffer.md
@@ -1,0 +1,5 @@
+---
+'@microsoft/atlas-css': minor
+---
+
+Fixing the select component size variations.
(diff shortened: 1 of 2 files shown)


homing1:
I just noticed that there's a subtle difference with the new change. The corners of the small variant is less round. Also the carats are now the same size for all size variants. Are these changes expected? 

new
image

old
image

ollips:
> I just noticed that there's a subtle difference with the new change. The corners of the small variant is less round. Also the carats are now the same size for all size variants. Are these changes expected?
> 
> new image
> 
> old image

Border change is expected and what I'm going for on current PR actually to match the input :D

Good catch on the carats, fixed now!

BenDMyers (approved):

jdanyow (approved):

PR #358, opened 2022-02-26
Radio

homing1:
Task: task-[531899](https://dev.azure.com/ceapex/Engineering/_workitems/edit/531899/)

Link: preview-[358](https://design.docs.microsoft.com/pulls/358)

This PR adds a radio component.

Note: radio with images is not being included. It has been depreciated from assessment. Report Feedback still uses it, but verified with PMs that the images can be removed as the pattern is not good for accessibility.

## Testing

1. Visit [the radio page](https://design.docs.microsoft.com/pulls/358/components/radio.html).
2. Check all themes and screensize.


css/src/components/form/radio.scss
```diff
@@ -0,0 +1,96 @@
+$radio-circle-size: 1.125rem !default;
```
wibjorn, line 1: Should the radio circle contain a background color of its own? I believe this would ensure the control would work regardless of where its placed

css/src/components/form/radio.scss
```diff
@@ -0,0 +1,96 @@
...
+$radio-dot-size: 0.75rem !default;
```
wibjorn, line 2: have we used `em` sizing appropriately throughout the component?

site/src/components/radio.md
```diff
@@ -0,0 +1,159 @@
...
+                            checked
+                    	/>
+                    	<span class="radio-label-text">50%</span>
+                    </label>
+				</td>
+				<td>
+					<label class="radio" title="">
+                    	<input
+                    		type="radio"
+                    		id="assessment-choice-b16fe0d7-9dfa-460d-a8c4-8b39cad04240-4"
```
wibjorn, line 69: perhaps we can reduce the assessment related ids in this markup?

css/src/components/form/radio.scss
```diff
@@ -0,0 +1,96 @@
...
+$radio-dot-checked-color: $primary !default;
+$radio-dot-color: $text-subtle !default;
+$radio-timing-function: $input-timing-function !default;
+$radio-duration: $input-transition-duration !default;
+$radio-animation: boop !default;
+$radios-spacing: 0.5rem !default;
```
wibjorn, line 8: `radios` spacing applied to the old `radios` (group) component. Are we sure we don't want to bring that over?

css/src/components/form/radio.scss
```diff
@@ -0,0 +1,96 @@
...
+
+@mixin radio-dot {
+	position: absolute;
+	inset: 0;
+	width: $radio-dot-size;
```
wibjorn, line 13: Could just be my screen, but I see varying degrees of dot wonkiness on different zooms. Not entirely sure it's avoid able, but could you double check?

image

site/src/components/radio.md
```diff
@@ -0,0 +1,159 @@
...
+```
+
+## Modifiers
+
+### Styles
+
+`is-checked` can be used with `.radio-dot` to create the appearance of a radio button with a checked state, without actually setting the state.
+
+```html
+<fieldset class="field border-none">
```
wibjorn, line 113: Not sure the text in this example is correctly aligned with the dot:
image

css/src/components/form/radio.scss
homing1: Originally, I included the `radios` component for the `is-vertical` modifier, but I tried wrestling with its related linting issues to no avail and I didn't want to just disable them. Open to suggestions!

image

css/src/components/form/radio.scss
homing1: I thought it went away for me for a while, but today it seems to have returned. I drew inspiration from a codepen that our colleague Scott O'Hara created, which doesn't seem to have this problem but haven't figured out why. 

https://codepen.io/scottohara/pen/MWoEGZe

css/src/components/form/radio.scss
homing1: Actually, I might have found a way to make it work.

css/src/components/form/radio.scss
wibjorn: Looks to me as well that the Figma has larger dots relative to circle size than what's implemented, although it's pretty close

https://www.figma.com/file/uVA2amRR71yJZ0GS6RI6zL/%F0%9F%8C%9E-Atlas-Design-Library?node-id=838%3A1096

css/src/components/form/radio.scss
homing1: Added `background-color: $body-background;`

css/src/components/form/radio.scss
homing1: Originally when I used `em`, I thought that the radio was too small at `font-size-xs` so there's not a lot to click on. Upon second thought, you can also click on the label so perhaps it's a non-issue. Updated to em.

site/src/components/radio.md
homing1: Remove the id

css/src/components/form/radio.scss
homing1: Added radios.

css/src/components/form/radio.scss
homing1: Updated the dot size

site/src/components/radio.md
homing1: Added `inset-block-start` to `radio-dot` to lower the radio down a bit, without making it look odd for single line text.

css/src/components/form/radio.scss
```diff
@@ -0,0 +1,117 @@
...
+		&::before {
+			@include radio-dot;
+		}
+	}
+
+	.radio-label-text {
+		margin: 0 $radios-spacing;
+	}
+
+	legend + & {
```
ollips, line 54: I think in design the word "Label" in the screenshot below was incorrectly used, more likely it matched the legend. So the question is should `legend` be styled like that? 

image

site/src/components/radio.md
```diff
@@ -0,0 +1,159 @@
...
+	<legend>This legend describes the nature of the selection.</legend>
+	<div class="radios">
+		<label class="radio">
+			<input name="question-2" type="radio" class="is-checked radio-dot" />
+			<span class="radio-label-text"
+				>Reliability metrics are built into organizational objectives for teams managing mission
+				critical workloads</span
+			>
```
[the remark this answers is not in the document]
wibjorn: I'd recommend changing the text as well.

css/src/components/form/radio.scss
wibjorn: I think we should remove this entirely if there are no major repercussions for updates in docs-ui

css/src/components/form/radio.scss
```diff
@@ -0,0 +1,117 @@
...
+		margin-block-start: $radios-spacing;
+	}
+
+	.radios & {
```
wibjorn, line 58: Should `radios` related code be placed in a separate file, similar to `.buttons`?

css/src/core/animations.scss
```diff
@@ -10,3 +10,23 @@
...
+	50% {
+		transform: scale(1.1);
+	}
+
+	100% {
+		transform: scale(1);
+	}
+}
+
+@mixin reduceMotion {
```
wibjorn, line 28: Should be in mixins, if included at this point at all

css/src/components/form/radio.scss
```diff
@@ -0,0 +1,117 @@
...
+	.radio-dot.is-checked {
+		animation: $radio-animation $radio-duration $radio-timing-function 1;
+		border-color: $radio-dot-checked-color;
+
+		&::before {
+			@include radio-dot-checked;
+		}
+	}
+
+	@include reduceMotion {
```
wibjorn, line 102: Personally, I think that this can be removed from this PR. The `boop` animation already falls under the reduced motion category IMO.

css/src/components/form/radio.scss
```diff
@@ -0,0 +1,117 @@
...
+
+		+ .radio {
+			margin-top: $radios-spacing;
+			margin-left: 0;
+		}
+	}
+
+	/* stylelint-disable selector-no-qualifying-type */
+
+	input {
```
wibjorn, line 84: Nit: unify code style with the rules starting on line 85:
```
input.is-checked
input.focus-visible

File: site/src/components/radio.md [hunk 13]
```diff
@@ -0,0 +1,159 @@
...
+description: The radio component in the Atlas Design System
+template: standard
+figmaEmbed: https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FuVA2amRR71yJZ0GS6RI6zL%2F%25F0%259F%258C%259E-Atlas-Design-Library%3Fnode-id%3D838%253A1096"
+---
+
+# Radio
+
+A radio button allows users to select a single option. It is usually associated with form submissions.
+
+<div class="table-wrapper margin-top-xs">
```
wibjorn, line 12: I think we should be making documentation tables in markdown

site/src/components/radio.md
```diff
@@ -0,0 +1,159 @@
...
+	</table>
+</div>
+
+## Usage
+
+Here is an example of a group of radio buttons. You can group multiple ratio buttons together within a `fieldset` and `radios` container.
+Make sure to use the same value for the `name` attribute within a group of radio buttons.
+
+```html
+<fieldset class="field border-none">
```
wibjorn, line 85: I think we should reduce the example here to include `radios` and `radio` only. Form patterns should be listed separately and will require further design guidance.

site/src/components/radio.md
```diff
@@ -0,0 +1,159 @@
...
+		</label>
+	</div>
+</fieldset>
+```
+
+## Modifiers
+
+### Styles
+
+`is-checked` can be used with `.radio-dot` to create the appearance of a radio button with a checked state, without actually setting the state.
```
wibjorn, line 108: This isn't really recommended - it's more for documentation purposes than functional. I wonder if we need to call it out

css/src/components/form/radio.scss
BenDMyers: That makes sense to me!

site/src/components/radio.md
```diff
@@ -0,0 +1,159 @@
...
+			</tr>
+		</thead>
+		<tbody>
+			<tr>
+				<td>Unchecked</td>
+				<td>
+					<label class="radio" title="">
+                    	<input
+                    		type="radio"
+                    		name="b16fe0d7-9dfa-460d-a8c4-8b39cad04240"
```
wibjorn, line 28: Do we need complex hashes here?

site/src/components/radio.md
wibjorn: is what I'm looking at not updated? seems like they're still around

css/src/components/form/radio.scss
```diff
@@ -0,0 +1,117 @@
...
+$radio-animation: boop !default;
+$radios-spacing: 0.5em !default;
+
+@mixin radio-dot {
+	position: absolute;
+	inset: 0;
+	width: $radio-dot-size;
+	height: $radio-dot-size;
+	margin: auto;
+	border-radius: 50%;
```
wibjorn, line 16: nit: rounded radius variable courtesy of @ollips's recent PR?

site/src/components/radio.md
homing1: Hmm maybe just refresh again? I removed it a few commits ago https://github.com/microsoft/atlas-design/pull/358/commits/d2577784fe6c9f09bd6196c8eb9a477d4a154684

css/src/components/form/radio.scss
```diff
@@ -0,0 +1,117 @@
...
+	cursor: pointer;
+
+	.radio-dot {
+		display: inline-block;
+		position: relative;
+		inset-block-start: 0.0625em;
+		flex-shrink: 0;
+		width: $radio-circle-size;
+		height: $radio-circle-size;
+		border: 1px solid $text;
```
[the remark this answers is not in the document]
homing1: Updated

site/src/components/radio.md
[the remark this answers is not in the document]
homing1: The format is due to Prettier. I updated the text to Lorem Ipsum though.

css/src/components/form/radio.scss
homing1: Removed, but it means we'll need to add `margin-top-xxs` in numerous places where we use the fieldset.

css/src/components/form/radio.scss
homing1: Moved to a new file `radios.scss`

css/src/core/animations.scss
homing1: reverted

css/src/components/form/radio.scss
homing1: reverted

css/src/components/form/radio.scss
homing1: Updated

site/src/components/radio.md
homing1: Updated

site/src/components/radio.md
homing1: Removed fieldset and legend

site/src/components/radio.md
homing1: Removed.

site/src/components/radio.md
homing1: updated

css/src/components/form/radio.scss
```diff
@@ -0,0 +1,117 @@
...
+		border-radius: $radius-rounded;
```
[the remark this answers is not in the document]
homing1: updated

css/src/components/form/radio.scss
homing1: Updated

site/src/components/radio.md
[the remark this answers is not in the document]
ollips: Just put the inner span's text and prettier would do it's thing correctly. Was able to update it locally just fine:

image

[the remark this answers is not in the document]
homing1: weird, I tried it again and it worked. Updated

css/src/components/form/radios.scss
```diff
@@ -0,0 +1,16 @@
+.radios {
+	.radio {
+		margin-top: $radios-spacing;
+	}
+
+	&.radios-vertical {
```
ollips, line 6: It's used to be an example on a doc page for that, isn't it? Why was it removed?

site/src/components/radio.md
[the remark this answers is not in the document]
ollips: Thanks!

css/src/components/form/radio.scss
ollips: That's what I've done for a bunch of `labels` after migrating to docs 😁

css/src/components/form/radios.scss
ollips: Oh, just realized you used it on a docs page. However, I don't think it was a good example, because the text in the examples is way to long so they would stack anyway. What about shortening it to `Lorem ipsum dolor sit amet` so the difference is visible? Or apply the same `Yes/No/Maybe` as in the example above.

image

css/src/components/form/radios.scss
```diff
@@ -0,0 +1,16 @@
+.radios {
```
jdanyow, line 1: Let's remove `.radios`. Duplicative of `display-flex gap-xxxs`. The `.checkbox` component doesn't have `.checkboxes`.

css/src/components/form/radios.scss
homing1: My original intention was to include a multi-line example on the page, but since the number of examples has been reduced to only two, it may no longer be practical. Updated to a shorter example.

css/src/components/form/radios.scss
homing1: Reverted

site/src/components/radio.md
```diff
@@ -0,0 +1,53 @@
...
+<label class="radio">
+	<input name="question-1" type="radio" class="radio-dot" value="2" />
+	<span class="radio-label-text">Maybe</span>
+</label>
+```
+
+You can wrap the radio button group with a container and use flex classes to display a vertical alignment.
+
+```html
+<div class="display-flex flex-direction-column margin-top-xxs">
```
ollips, line 39: nit: `margin-top-xxs` is not needed here

site/src/components/radio.md
```diff
@@ -0,0 +1,53 @@
...
+	<label class="radio margin-bottom-xxs">
+		<input name="question-3" type="radio" class="radio-dot" value="0" />
+		<span class="radio-label-text">0%</span>
+	</label>
+	<label class="radio margin-bottom-xxs">
+		<input name="question-3" type="radio" class="radio-dot" value="1" />
+		<span class="radio-label-text">50%</span>
+	</label>
+	<label class="radio margin-bottom-xxs">
```
ollips, line 48: nit: `margin-bottom-xxs` is not needed here

site/src/components/radio.md
```diff
@@ -0,0 +1,53 @@
...
+| Unchecked | <label class="radio" title=""><input type="radio" name="question-0" value="0" class="radio-dot"> <span class="radio-label-text">50%</span></label>           | <label class="radio" title=""><input type="radio" name="question-0" value="1" class="radio-dot" disabled> <span class="radio-label-text">70%</span></label>           |
+| Checked   | <label class="radio" title=""><input type="radio" name="question-0-1" value="0" class="radio-dot" checked> <span class="radio-label-text">50%</span></label> | <label class="radio" title=""><input type="radio" name="question-0-2" value="1" class="radio-dot" checked disabled> <span class="radio-label-text">70%</span></label> |
+
+## Usage
+
+Here is an example of radio buttons. Make sure to use the same value for the `name` attribute within a group of radio buttons.
+
+```html
+<label class="radio">
+	<input checked name="question-1" type="radio" class="radio-dot" value="0" />
```
ollips, line 23: nit: Shouldn't the `value` match the label's text? I know it doesn't matter much in these examples but I think it would be nice to follow the standards. Same for all examples

site/src/components/radio.md
homing1: removed

site/src/components/radio.md
homing1: removed

site/src/components/radio.md
ollips: I see, still, I don't think we should rely on assessments here, and better to follow the general html standards in case someone outside the docs world would land on a page and check the examples.

homing1: updated

ollips (approved):

jdanyow (approved):

css/src/components/form/radio.scss
```diff
@@ -0,0 +1,92 @@
...
+		.radio-dot:not(:checked)::before {
+			@include radio-dot-checked;
+		}
+	}
+
+	/* stylelint-enable */
+
+	/* stylelint-disable selector-no-qualifying-type */
+
+	input.focus-visible,
```
wibjorn, line 69: Sorry I forgot this in previous suggestions, but we need to use the mixin to target `.focus-visible` to ensure we can easily drop the polyfill when it comes time

mixin called `focus-visible` lives here:
css\src\mixins\focus.scss

wibjorn (approved):

PR #367, opened 2022-03-14
Fix `[disabled].button.button-filled` buttons in all semantic colors.

wibjorn:
Task: task-531865

Link: preview-367

I accidentally implemented some incorrect logic for disabled buttons. Luckily removing certain lines of code fixes the issue. This PR also updates the documentation to more closely match the Figma.

## Testing

1. Visit: preview-367
2. Visit buttons page. Check all disabled examples.

### DIFF
--- a/site/src/components/button.md
+++ b/site/src/components/button.md
@@ -23,11 +23,11 @@ All buttons, by default, are secondary buttons. There are three gradutating seco
 2. Clear
 3. Filled
 
-| Type     | Class                      | Default State                                         | Hover                                                            |
-| -------- | -------------------------- | ----------------------------------------------------- | ---------------------------------------------------------------- |
-| Outlined | `.button`                  | <button class="button">Default</button>               | <button class="button is-hovered">Default</button>               |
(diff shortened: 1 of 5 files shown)


homing1:
I just noticed a typo on line 20 
image

site/src/components/button.md
```diff
@@ -126,28 +126,11 @@ Make a button take up the full width of a container.
 
 ### Loading buttons
 
-Ensure the user knows they need to wait for some event (like a fetch request) in order to interact with a button.
+Ensure the user knows they need to wait for some event (like a fetch request) in order to interact with a button. Can be used with the `disabled` attribute, as in the second button example below.
```
homing1, line 129: I'm wondering if `is-disabled` should also be mentioned somewhere? Not just for buttons, since we have that modifier available on other components too.

wibjorn: Updated this with a list of commonly used state-related classes

homing1 (approved):

BenDMyers (approved):

PR #371, opened 2022-03-16
Updates uses of focus-visible class instances with mixin that can piv…

wibjorn:
Task: task-572109

Link: preview-371

Near the beginning of this project, we tried to be friendly to the forthcoming support of `:focus-visible` by using a mixin to allow builds to pivot between the polyfill (which uses `.focus-visible`) and the psuedo-selector by setting a feature flag. At some point we all collectively forgot about that best practice. Luckily, we've not come very far away from it - and with Safari now supporting the psuedo-selector, it's important that we get the mixin up and running to ensure a smooth transition.

## Testing

1. Tab through links in next section
2. Read code

## Additional information

[Button](https://design.docs.microsoft.com/pulls/371/components/button.html)
[Link Button](https://design.docs.microsoft.com/pulls/371/components/link-button.html)
[Radio](https://design.docs.microsoft.com/pulls/371/components/radio.html)

### DIFF
--- a/css/src/components/buttons.scss
+++ b/css/src/components/buttons.scss
@@ -64,11 +64,18 @@
 				z-index: $zindex-hover;
 			}
 
-			&.focus-visible,
 			&.is-focused,
 			&:active,
 			&.is-active,
 			&.is-selected {
+				z-index: $zindex-focus;
+
+				&:hover {
+					z-index: $zindex-multi;
+				}
+			}
+
+			@include focus-visible {
 				z-index: $zindex-focus;
 
 				&:hover {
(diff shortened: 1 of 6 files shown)


BenDMyers (approved):

ollips (approved):

css/src/components/buttons.scss
```diff
@@ -76,6 +75,14 @@
 				}
 			}
 
+			@include focus-visible {
```
homing1, line 78: Comment: not related to the changes here, but it seems like the filled button on this page is missing a visible focus ring https://design.docs.microsoft.com/pulls/371/components/button.html

wibjorn: Ah good catch, you and @BenDMyers have found the same thing

wibjorn: Addressing in a new one

wibjorn: https://ceapex.visualstudio.com/Engineering/_workitems/edit/565910

PR #388, opened 2022-03-29
Fix off centered radio

homing1:
Task: task-[580494](https://dev.azure.com/ceapex/Engineering/_workitems/edit/580494)

Link: preview-[388](https://design.docs.microsoft.com/pulls/388/components/radio.html)

The current radio dot appears to be off-centered at times. Adding `background-clip` with padding seems to help. The use of `appearance: none` and applying custom styling eliminates the need for the `:before` pseudo element.

## Testing

1. Visit https://design.docs.microsoft.com/pulls/388/components/radio.html
2. Verify that the checked radio button dot is centered. It should stay centered even if you adjust screen size.
3. Add different font sizes to the label to test. It still looks a bit off-centered with `font-size-xs`.

### DIFF
--- a/css/src/components/form/radio.scss
+++ b/css/src/components/form/radio.scss
@@ -37,7 +37,7 @@ $radio-spacing: 0.5em !default;
 
 	&.is-hovered,
 	&:hover:not([disabled]) {
-		.radio-dot:not(:checked)::before {
+		.radio-dot:not(:checked) {
 			background: $radio-dot-hover-unchecked-color;
 		}
 	}
--- a/css/src/components/form/radio.scss
+++ b/css/src/components/form/radio.scss
@@ -17,7 +17,6 @@ $radio-spacing: 0.5em !default;
 	cursor: pointer;
 
 	.radio-dot {
-		inset-block-start: 0.0625em;
 		width: $radio-circle-size;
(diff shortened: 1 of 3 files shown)


css/src/components/form/radio.scss
```diff
@@ -5,67 +5,33 @@ $radio-border-width: $control-border-width !default;
...
-		display: inline-block;
-		position: relative;
-		inset-block-start: 0.0625em;
-		flex-shrink: 0;
 		width: $radio-circle-size;
 		height: $radio-circle-size;
+		padding: $radio-dot-whitespace;
 		border: $radio-border-width solid $radio-border-color;
 		border-radius: $border-radius-rounded;
+		background-clip: content-box !important;
```
wibjorn, line 25: For whatever reason, we're getting different results on Safari on iOS. (I've had issues with background clip on Safari and iOS chrome, so I knew to double check). Doesn't look to bad but unfortunately it's not consistent.

image

wibjorn (changes requested):

css/src/components/form/radio.scss
homing1: Thanks for letting me know. I'll close the PR for now until I find an alternative solution.

PR #394, opened 2022-04-07
Fix radio off-centered dot and background colors

homing1:
Task: task-[580494](https://dev.azure.com/ceapex/Engineering/_workitems/edit/580494/)

Link: preview-[394](https://design.docs.microsoft.com/pulls/394/components/radio.html)

This PR addresses three issues for radios:
- Sometimes the radio dot is off-centered when checked, depending on screen size and zoom level. This approach fills the entire radio, then uses box shadow to add the white space. 
- The body background color wasn't filling the entire unchecked radio when paired with a container with `background-color-body-medium`. The entire radio should be white when unchecked.
- The color on hover on `is-checked` was not matching the checked color. The hover color should be the same.

## Testing

1. Visit https://design.docs.microsoft.com/pulls/394/components/radio.html .
2. Try different screen sizes and zoom levels. The dot should stay in the center. 
3. Add `background-color-body-medium` on a radio's parent container. The entire circle should have a white background when unchecked.
4. Add `is-checked` to an unchecked radio. There should be no color change on hover.
5. Test with OS high contrast or browser forced colors

### DIFF
--- a/css/src/components/form/radio.scss
+++ b/css/src/components/form/radio.scss
@@ -1,4 +1,5 @@
 $radio-circle-size: 1.25em !default;
+$radio-dot-size: 0.625em !default;
 $radio-border-color: $control-border !default;
 $radio-border-width: $control-border-width !default;
 $radio-dot-checked-color: $primary !default;
@@ -8,6 +9,17 @@ $radio-timing-function: $input-timing-function !default;
 $radio-duration: $input-transition-duration !default;
 $radio-animation: boop !default;
 $radio-spacing: 0.5em !default;
+
+@mixin radio-dot {
+	position: absolute;
+	inset: 0;
+	width: $radio-dot-size;
+	height: $radio-dot-size;
+	margin: auto;
+	border-radius: $border-radius-rounded;
+	content: '';
+	overflow: hidden;
+}
 
 @mixin radio-dot-checked {
 	background: $radio-dot-checked-color;
@@ -21,6 +33,7 @@ $radio-spacing: 0.5em !default;
 
 	.radio-dot {
 		display: inline-block;
+		position: relative;
 		inset-block-start: 0.0625em;
 		flex-shrink: 0;
 		width: $radio-circle-size;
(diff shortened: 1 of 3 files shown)


wibjorn (approved):

css/src/components/form/radio.scss
```diff
@@ -10,62 +9,35 @@ $radio-duration: $input-transition-duration !default;
...
-	margin: auto;
-	border-radius: $border-radius-rounded;
-	background-color: $body-background;
-	content: '';
-	overflow: hidden;
-}
-
 @mixin radio-dot-checked {
 	background: $radio-dot-checked-color;
+	box-shadow: inset 0 0 0 0.25em $body-background;
```
jdanyow, line 14: The dots disappear in forced colors mode

jdanyow (changes requested):

css/src/components/form/radio.scss
jdanyow: might be able to use this technique https://github.com/microsoft/atlas-design/blob/main/css/src/components/toggle.scss#L25

homing1: yup looking at that right now

homing1: Forced colors sets box shadow to none, so I think we can use the previous strategy of `:before` for forced colors only. This means the dot can still look weird in forced colors mode though. 

image

wibjorn: That will work, just verifying `forced-color-adjust: none;` was used inside the mixin body during testing though before we go that route.

```scss
@include forced-colors {
  forced-color-adjust: none;
  [... etc]
}
```

homing1: Tested using both with and without `forced-color-adjust: none;` using all four OS high contrast settings.

css/src/components/form/radio.scss
```diff
@@ -42,30 +40,26 @@ $radio-spacing: 0.5em !default;
 		height: $radio-circle-size;
 		border: $radio-border-width solid $radio-border-color;
 		border-radius: $border-radius-rounded;
+		background-color: $body-background;
 		color: $radio-dot-color;
 		appearance: none;
+		content: '';
```
ollips, line 46: nit: can you add here `cursor: pointer;`? For some reason, I get the default arrow, would be nice to see the pointer :)

homing1: Added!

ollips (approved):

jdanyow (approved):

PR #397, opened 2022-04-11
initial work for playwright visual diffing and user flow testing

wibjorn:
Task: task-588379
Link: preview-397

This PR adds a new integration package and install playwright.  There's no diffing going on, but screenshots should be captured on runs of the [`visual-diff` action](https://github.com/microsoft/atlas-design/actions/workflows/visual-diff.yml) and tests should be run on the [`flows` action](https://github.com/microsoft/atlas-design/actions/workflows/flows.yml)

run a few local commands and read through code.

`npm run screenshots:light` see light screenshots populated in `integration/screenshots`
`npm run screenshots:all` see light,  dark, high contrast screenshots populated in `integration/screenshots`

`npm run flows`, runs one unit test in two device projects

`npm run codegen`, run process to get some code for a test


.github/workflows/flows.yml
```diff
@@ -0,0 +1,58 @@
...
+  # Triggers the workflow on push or pull request events but only for the main branch
+  # push:
+  #   branches:
+  #     - main
+  pull_request:
+    branches:
+      - main
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
BenDMyers, line 15: I'm fairly new to GitHub Actions, so forgive me if this is wrong, but does this property need a value after it?

wibjorn: Nope, there's an optional inputs property though; we don't need it now. Can't manually dispatch until this file is in `main` though.

https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

BenDMyers: YAML is even weirder than I thought! Works for me 👍🏻

integration/config/page.ts
```diff
@@ -0,0 +1,95 @@
+import { devices, PlaywrightTestOptions, PlaywrightWorkerOptions, Project } from '@playwright/test';
```
wibjorn, line 1: renames to pages, remove `themes` from configuration interface

homing1:
Some observations:

- After running `npm i` and `npm run screenshots:light`, I see the following error. It worked after I ran `npx playwright install`
image

- It mentions about a report, but the command fails as it's not pointing to the integration folder
image

- `npm run codegen` failed for me 
image

integration/README.md
```diff
@@ -0,0 +1,18 @@
+# Atlas Integration Testing
+
+This folder provides the basis for all integration testing against Atlas. Atlas is primarily a CSS framework, and we are not seeking to test CSS itself. We use two methods of integration testing here.
+
+1. Visual regression testing (visual diff) against our site's component pages - a work in progress. [See config](./visual-diff.config.ts) and [entry point](./visual-diff/visual-diff.spec.ts).
+2. Assertion style testing of interactions (flows) for elements and behaviors in the `atlas-js` package. [See config](./visual-diff.config.ts) and [example](./flows/popover.spec.ts).
+
+## Using Playwright
+
+Both of the above use [`@playwright/test`](https://playwright.dev/), the documentaton for should be the primary source of information on how to write tests here. Because of our monorepo setup, we've taken the approach of referencing playwright commands via npm. They the primary commands available at the root of the project are:
```
homing1, line 10: nit: `the documentaton for should be` -> `their documentation should be`?

wibjorn:
@homing1 

1. I'll see what I can do about ensuring playwright is installed, but right now that's as I'd expect. Perhaps at least a lint in the readme about it.
2. Because of our monorepo, you'll need to use `npm run show-resport` instead of using playwright directly from the root folder.
3. See docs in integration/Readme.md, but you need the local server running before you run `codegen`

integration/README.md
homing1: `They the primary ` -> `The primary`?

homing1:
> @homing1
> 
> 1. I'll see what I can do about ensuring playwright is installed, but right now that's as I'd expect. Perhaps at least a hint in the readme about it.
> 2. Because of our monorepo, you'll need to use `npm run show-resport` instead of using playwright directly from the root folder. Or you can CD into /integration if you so desire :D
> 3. See docs in integration/Readme.md, but you need the local server running before you run `codegen`

This was just me following the test steps before reading the code :D  

`npm run show-report` works and the codegen works after running the local server.

tests/example.spec.ts
```diff
@@ -0,0 +1,381 @@
+import { test, expect, Page } from '@playwright/test';
```
jdanyow, line 1: is this file/folder needed?

integration/flows/popover.spec.ts
```diff
@@ -0,0 +1,19 @@
+import { test, expect } from '@playwright/test';
```
jdanyow, line 1: Rather than "flows" how about "tests". I expect this will contain tests for many components, behaviors, etc.

integration/flows/popover.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+
+test('popover closes .popover-content when an outer element is clicked', async ({ page }) => {
+	// Go to http://localhost:1111/components/popover.html
+	await page.goto('http://localhost:1111/components/popover.html');
```
jdanyow, line 5: nit- relative url to conform with:

>It is also recommended to specify [testOptions.baseURL](https://playwright.dev/docs/api/class-testoptions#test-options-base-url) in the config, so that tests could use relative urls.
https://playwright.dev/docs/test-advanced#launching-a-development-web-server-during-the-tests

jdanyow (changes requested):

integration/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+You must [install Playwright and its dependencies on your machine](https://playwright.dev/docs/intro#manually) before using any commands.
+
+```sh
+npx playwright install
+```
+
+## Using Playwright
+
+Both of the above use [`@playwright/test`](https://playwright.dev/). Their documentaton for should be the primary source of information on how to write tests here. Because of our monorepo setup, we've taken the approach of referencing playwright commands via npm. The primary commands available at the root of the project are:
```
homing1, line 18: nit: documentaton ~~for~~ should be

integration/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+| Command                           | Referant                                                                  | Description                                                                                                 |
```
homing1, line 20: nit:  Referent ?

integration/README.md
wibjorn: done!

integration/README.md
wibjorn: done!

homing1 (approved):

jdanyow (approved):

PR #402, opened 2022-04-25
[wip]

homing1:
Task: task-[work-item-number]

## Testing

1. Step one
2. Step two
   Expected result:

## Additional information

[Optional]


css/src/core/bare-elements.scss
```diff
@@ -11,3 +11,15 @@ details > * {
...
+	color: $hyperlink;
+}
+
+:where(a:visited) {
+	color: $visited;
+}
+
+::part(visually-hidden) {
+	@include visually-hidden();
+}
```
jdanyow, line 25: thinking this should go in /atomics/visually-hidden.scss so that when we eventually move .visually-hidden from docs to atlas they're combined into 
```scss
.visually-hidden,
::part(visually-hidden) {
	@include visually-hidden();
}
```

js/package-lock.json
```diff
@@ -1,34 +1,33 @@
 {
```
jdanyow, line 1: should this file be included in the PR?

js/src/elements/star-rating.ts
```diff
@@ -0,0 +1,283 @@
...
+
+	get disabled() {
+		return this.hasAttribute('disabled');
+	}
+
+	set disabled(val) {
+		this.setAttribute('disabled', val.toString());
+		const fieldset = this.shadowRoot?.querySelector('fieldset') as HTMLFieldSetElement;
+		fieldset.disabled = val;
+		fieldset.querySelectorAll('input').forEach(input => (input.disabled = val));
```
jdanyow, line 189: Thoughts on making `disabled` work like `name` where the prop sets the attribute and the attributeChangedCallback handles synchronization of the child elements?

js/src/elements/star-rating.ts
```diff
@@ -0,0 +1,283 @@
...
+		}
+		if (name === 'disabled') {
+			this.disabled = this.hasAttribute('disabled');
+		}
+	}
+
+	handleEvent(event: Event) {
+		switch (event.type) {
+			case 'change':
+				this.setAttribute('value', (event.target as HTMLInputElement).value);
```
jdanyow, line 259: needed?

js/src/elements/star-rating.ts
```diff
@@ -0,0 +1,283 @@
...
+				break;
+			case 'formdata':
+				// https://web.dev/more-capable-form-controls/
+				(event as FormDataEvent).formData.append(this.name, this.value.toString());
+				break;
+		}
+	}
+
+	updateStarFill(newValue: number) {
+		const stars = this.shadowRoot?.querySelectorAll('input');
```
jdanyow, line 270: think you can use `this.shadowRoot!.query...` here right? To avoid the `if (stars)` below?

js/src/elements/star-rating.ts
```diff
@@ -0,0 +1,283 @@
...
+
+	/** TODO: add form related properties */
+
+	connectedCallback() {
+		this.shadowRoot?.addEventListener('change', this);
+		// arrow function to bind `this` value to star rating in handleFormData
+		this.closest('form')?.addEventListener('formdata', this);
+		this.shadowRoot?.querySelectorAll('input[type="radio"]').forEach(input => {
+			/* Upon initial focus on a brand new star rating, set value to match the currently focused star.
+			Otherwise, the value remains at 0 until user selects the focused star. */
```
jdanyow, line 216: I think it would be best to align with standard radio behavior and not force selection here.

js/src/elements/star-rating.ts
```diff
@@ -0,0 +1,283 @@
...
+		this.setAttribute('name', value);
+	}
+
+	get value() {
+		return Math.max(0, Math.min(parseInt(this.getAttribute('value') || '0'), 5));
+	}
+
+	set value(val) {
+		this.setAttribute('value', val.toString());
+	}
```
jdanyow, line 206: For consistency with native input/select/textarea I think it would be good to make the value property be of type string with a default value of `''`. 

Inputs have a property getter for `valueAsNumber` which we could add here for convenience:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement

The other thing to note is unlike the `name`, `disabled`, and `required` properties, the `value` property is not "reflected", meaning when you set a value property on an input, it's associated attribute is not updated automatically. This is useful for form "reset" functionality should we choose to implement support for the reset event in the future.

Initially I had fooled around with the floor/numeric value property as you have here but it seems like it would be better to align with other native controls especially for validation purposes.

js/src/elements/star-rating.ts
```diff
@@ -0,0 +1,283 @@
...
+		});
+	}
+
+	disconnectedCallback() {
+		this.shadowRoot?.removeEventListener('change', this);
+		this.closest('form')?.removeEventListener('formdata', this);
+	}
+
+	attributeChangedCallback(name: string, oldValue: string, newValue: string) {
+		if (oldValue !== newValue) {
```
jdanyow, line 229: nice optimization 👍

js/src/elements/star-rating.ts
```diff
@@ -0,0 +1,283 @@
...
+			if (stars) {
+				if (i < newValue) {
+					stars[i].classList.add('is-selected');
+				} else {
+					stars[i].classList.remove('is-selected');
+				}
+			}
+		}
+	}
+}
```
jdanyow, line 281: support for required, validity, and type properties in future pr?

jdanyow (changes requested):

js/src/elements/star-rating.ts
jdanyow: note that explicitly disabling the inputs should not be required- disabling the fieldset will automatically disable it's children

jdanyow:
as you hover across the stars the description text seems to move to the right slightly- minor css issue?

css/src/core/bare-elements.scss
jdanyow: could have it's own page and maybe we link to it from display.md, docs could be a followup work item

homing1: Added to atomics with a comment to add the actual atomic class later

js/package-lock.json
homing1: removed

js/src/elements/star-rating.ts
homing1: updated

js/src/elements/star-rating.ts
homing1: removed

js/src/elements/star-rating.ts
homing1: updated

js/src/elements/star-rating.ts
homing1: removed

js/src/elements/star-rating.ts
homing1: updated

js/src/elements/star-rating.ts
homing1: Added required and type, will add validity in the next pr

homing1:
> as you hover across the stars the description text seems to move to the right slightly- minor css issue?

Updated css to fix

js/src/elements/star-rating.ts
homing1: Actually, not having the value reflect immediately seems to be problematic for star filling. I'll look for another mechanism to trigger the filling of the stars.

homing1: nevermind, fixed it

css/src/core/bare-elements.scss
jdanyow, line 25: ✂️

homing1: ack removed

js/src/elements/star-rating.ts
```diff
@@ -0,0 +1,289 @@
...
+	}
+	
+	fieldset {
+		display: flex;
+        margin: 0;
+        border: none;
+        padding-block: 0.75rem;
+	}
+	
+	#input-container {
```
BenDMyers, line 36: Is there a possibility that a page could have more than one rating widget? If so, should this be a class or other reusable selector?

homing1: Updated. I tried using a class in the past and it wasn't filling the stars properly for whatever reason, but it seems to be fine now.

jdanyow (approved):

js/src/elements/star-rating.ts
jdanyow: This stylesheet and the element with the `#input-container` id are scoped to the element's shadow dom so in theory this is a fine practice and there should not be a collision risk.

BenDMyers: Ah, good deal - just my own misunderstanding of shadow DOM then! No worries!

js/src/elements/star-rating.ts
```diff
@@ -0,0 +1,289 @@
...
+		cursor: not-allowed;
+		pointer-events: none;
+	}
+  </style>
+
+<fieldset>
+	<slot name="legend" part="legend">Enter rating</slot>
+
+	<div class="star-container">
+		<input part="visually-hidden" type="radio" value="1" id="radio-1" />
```
BenDMyers, line 122: I opened up the accessibility tree display to see what the inputs' accessible name was that screenreaders would announce, and the accessible names are coming up empty. See:
Screenshot of accessibility tree display for an input in the star rating widget. Although multiple labels are associated with the input, the accessible name is coming up as an empty string.

Each individual star input has two labels associated with it -- the star and the "Poor"/"Fair"/etc. rating, but I think the star label is winning out and it doesn't have any accessible text inside of it. This means that as screenreader users navigate between the inputs, they might not be told any information like which star they're on or what that rating means.

BenDMyers: (For anyone unfamiliar with the accessibility tree: I'm looking under the Computed Properties pane on the right. The computed name is `""`, which it's getting from two `label` elements)

BenDMyers: Might be worth disregarding - this works as intended in VoiceOver, so maybe fieldset works differently than I expected 😅

js/src/elements/star-rating.ts
```diff
@@ -0,0 +1,289 @@
...
+		<label for="radio-1">
+			<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" style="overflow: visible" >
```
jdanyow, line 124: aria-hidden="true" on the svgs ?

homing1: Added

js/src/elements/star-rating.ts
BenDMyers: Resolving this particular thread as the radios' legend names are now coming up in the accessibility tree.

jdanyow (approved):

BenDMyers (approved):

PR #405, opened 2022-05-07
[wip]

homing1:
Task: task-[work-item-number]

Link: preview-[405](https://design.docs.microsoft.com/pulls/405/patterns/form-validation.html)

## Testing

1. Step one
2. Step two
   Expected result:

## Additional information

[Optional]


js/src/elements/star-rating.ts
```diff
@@ -0,0 +1,355 @@
...
+			}
+		}
+	}
+
+	get valueAsNumber() {
+		return parseInt(this.getAttribute('value') ?? '');
+	}
+
+	get validity() {
+		// borrow input's validity, https://caniuse.com/?search=validity
```
homing1, line 251: Internals is not supported for usage yet, so can't hook into internals/form control API. `Validity` refers to a read-only `ValidityState` interface. Rather than trying to create our own `ValidityState` interface, it may be simpler to just sync with the active input's validity, or the first input if not active.

https://web.dev/more-capable-form-controls/
https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/setCustomValidity

site/src/patterns/form-validation.md
```diff
@@ -0,0 +1,145 @@
+---
+title: Form Validation
+description: The Form validation pattern in the Atlas Design System
+template: standard
+---
+
+Use the form behavior component within a form to ensure accessible form validation. It expands on HTML5 validation.
+
+```html
+<form id="question-ask" data-form-type="question" action="#" method="POST" new="" novalidate="">
```
homing1, line 10: Currently using this as a POC to verify things will still work for QnA, so the examples here are taken from QnA ask question form template.

js/src/behaviors/tag-inputs.ts
```diff
@@ -0,0 +1,70 @@
+export function initTagInputs(container: HTMLElement = document.body) {
```
homing1, line 1: This is added to replicate adding/removing tags.

PR #407, opened 2022-05-20
Add form field control

homing1:
Task: task-[606276](https://dev.azure.com/ceapex/Engineering/_workitems/edit/606276/), task-[606275](https://dev.azure.com/ceapex/Engineering/_workitems/edit/606275/)

Link: preview-[407](https://design.docs.microsoft.com/pulls/407/components/form.html)

This PR adds form field and control components. 

## Testing

1. Review https://design.docs.microsoft.com/pulls/407/components/form.html.
2. Compare with https://docsstyleguide.z13.web.core.windows.net/form.html if desired.


css/src/components/form/form.scss
```diff
@@ -0,0 +1,241 @@
...
+
+			.input,
+			.select select {
+				padding-inline-end: $icon-size;
+			}
+
+			.input[type='date'] {
+				position: relative;
+
+				&::-webkit-calendar-picker-indicator {
```
homing1, line 94: This indicator is used here https://review.docs.microsoft.com/en-us/test/learn/certifications/courses/az-203t01?branch=main&preview=ilt

site/src/components/form.md
```diff
@@ -0,0 +1,195 @@
...
+For control elements that are required for form submission, you can add a span element with the `.required-indicator` class inside the label or adjacent to the element.
+
+```html
+<div class="field">
+	<div class="control">
+		<label class="field-label label margin-bottom-xxs" for="sample-form-input-2">
+			First name
+			<span class="required-indicator" aria-hidden="true"></span>
+		</label>
+		<input class="input" type="text" placeholder="First name" id="sample-form-input-2" />
```
jdanyow, line 192: The label's required indicator is aria-hidden, which might make sense assuming a required attribute is present on the input and provides the same info via assistive technology. 

This article suggests adding alt/aria-label for the required indicator asterisk. Might be worth getting the a11y team's input if we're not sure here. It could be annoyingly redundant to have the label and the input both announce "required".
https://www.deque.com/blog/anatomy-of-accessible-forms-required-form-fields/

site/src/components/form.md
```diff
@@ -0,0 +1,195 @@
...
+```html
+<div class="field">
+	<div class="control control-icons control-icon-right">
+		<input class="input input-sm" placeholder="www.microsoft.com" />
+		<span class="icon border" aria-hidden="true"> </span>
+	</div>
+</div>
+```
+
+`.is-loading` is used to add a spinner icon.
```
jdanyow, line 171: can we remove this modifier?

site/src/components/form.md
```diff
@@ -0,0 +1,195 @@
...
+
+The `.field` component is a container for form control elements to provide consistency.
+It can be used to hold `.control` containers, which holds the individual control elements.
+
+`.field-label` is used on the label. Add `field-label-sm`/`field-label-lg` makes label a little bigger or smaller than the default.
+
+```html
+<div class="field">
+	<label class="field-label field-label-sm label" for="sample-form-input-sm"> Input </label>
+	<div class="control">
```
jdanyow, line 31: can we rename the control class to something `field-`? eg `field-body`? I think something like that would align better with the naming standards and disambiguate control from the [control mixin](https://github.com/microsoft/atlas-design/blob/main/css/src/components/form/input.scss#L20)

site/src/components/form.md
```diff
@@ -0,0 +1,195 @@
...
+		<input
+			id="sample-form-input-sm"
+			name="title"
+			class="input"
+			type="text"
+			placeholder="e.g., Why is the sky blue?"
+		/>
+	</div>
+</div>
```
jdanyow, line 40: Can this pr standardize where to insert the field note/help text/described-by?
Example:
https://primer.style/css/components/forms#form-group-validation

site/src/components/form.md
```diff
@@ -0,0 +1,195 @@
...
+		<a class="button button-primary"> Transfer </a>
+	</div>
+</div>
+```
+
+`.field-horizontal` is used for displaying form label and field side by side.
+`field-body` is used on the control element container.
+
+```html
+<div class="field field-horizontal">
```
jdanyow, line 102: Do you know if field-horizontal is used in docs? Thinking it would be nice to remove this modifier for a couple reasons:
1. Consistent form layout across the site.
2. In practice, using this modifier to stack horizontal fields would result in alignment issues due to varying label widths.

site/src/components/form.md
```diff
@@ -0,0 +1,195 @@
...
+- [`.select`](./select.md)
+- [`.textarea`](./textarea.md)
+- custom form elements
+
+## Form Field
+
+The `.field` component is a container for form control elements to provide consistency.
+It can be used to hold `.control` containers, which holds the individual control elements.
+
+`.field-label` is used on the label. Add `field-label-sm`/`field-label-lg` makes label a little bigger or smaller than the default.
```
BenDMyers, line 26: I've got a slight rewording recommendation:
```suggestion
`.field-label` is used on the label. Adding `field-label-lg` or `field-label-sm` makes the label a little bigger or smaller than the default.
```

site/src/components/form.md
```diff
@@ -0,0 +1,195 @@
...
+				placeholder="www.microsoft.com"
+			/>
+		</div>
+	</div>
+</div>
+```
+
+## Form Control
+
+Form control element are wrapped with a `.control` container.
```
BenDMyers, line 136: Nit: this should probably be `elements`

css/src/components/form/form.scss
```diff
@@ -0,0 +1,238 @@
...
+			}
+
+			&:first-child {
+				.button,
+				.input,
+				.select select {
+					border-top-right-radius: 0;
+					border-bottom-right-radius: 0;
+				}
+			}
```
BenDMyers, line 150: Could this be an opportunity to use logical border radius properties?

```suggestion
			&:first-child {
				.button,
				.input,
				.select select {
					border-start-end-radius: 0;
					border-end-end-radius: 0;
				}
			}
```

More info: [`border-start-end-radius`](https://developer.mozilla.org/en-US/docs/Web/CSS/border-start-end-radius) and [`border-end-end-radius`](https://developer.mozilla.org/en-US/docs/Web/CSS/border-end-end-radius)

css/src/components/form/form.scss
```diff
@@ -0,0 +1,238 @@
...
+
+			&:last-child {
+				.button,
+				.input,
+				.select select {
+					border-top-left-radius: 0;
+					border-bottom-left-radius: 0;
+				}
+			}
```
BenDMyers, line 159: Similarly, can we use logical border radius properties here?

```suggestion
			&:last-child {
				.button,
				.input,
				.select select {
					border-start-start-radius: 0;
					border-end-start-radius: 0;
				}
			}
```

css/src/components/form/form.scss
BenDMyers: (If this shouldn't change for RTL languages, please feel free to disregard)

css/src/components/form/form.scss
```diff
@@ -0,0 +1,238 @@
...
+	&.field-horizontal {
+		@include tablet {
+			display: flex;
+		}
+
+		.field-label {
+			@include tablet {
+				margin-inline-end: $spacer-7;
+				align-self: center;
+				text-align: right;
```
BenDMyers, line 196: Is this an opportunity to use the logical value `end`?

```suggestion
				text-align: end;
```

css/src/components/form/form.scss
```diff
@@ -0,0 +1,238 @@
...
+
+$icon-size: 2.25em;
+
+$icon-color: $secondary-background !default;
+$icon-active-color: $secondary !default;
+
+.control {
+	position: relative;
+	font-size: $form-font-size;
+	text-align: left;
```
BenDMyers, line 15: Could we use the logical value `start` here?

```suggestion
	text-align: start;
```

BenDMyers (changes requested):
I've got a few wording nits, as well as some ideas for logical properties/values. If the design doesn't work for the logical props/values, please feel free to disregard!

site/src/components/form.md
homing1: I essentially combined the `.field is-grouped` and `.field is-horizontal` into one. `.field is-grouped` is still being used in Docs, but looks like it's mostly for grouping submit + cancel buttons so it can be replaced by flex atomics. REST API Try It has horizontal label and input field, but it's using columns to bypass the second issue you mentioned above. All to say that yes it can be removed.

css/src/components/form/form.scss
homing1: Updated

site/src/components/form.md
homing1: Removed. Looks like the autocomplete's loader is separate from this form field loader. If we are planning to add autocomplete (eventually), then perhaps we can add it at that time.

site/src/components/form.md
homing1: How about `field-section`? I'm thinking a field can have multiple sections, with each section containing an element and its label. If a field can only hold a single element, then `field-body` would be appropriate as it would be weird to have multiple bodies.

site/src/components/form.md
homing1: I was going to save it for the form validation document, but I added a section about it. Once the form validation document is available, I can add a link to it.

site/src/components/form.md
homing1: Removed

site/src/components/form.md
homing1: Updated

css/src/components/form/form.scss
homing1: Updated

css/src/components/form/form.scss
homing1: Updated

css/src/components/form/form.scss
homing1: Updated

site/src/components/form.md
homing1: In the asterisk example, it seems like they just put the asterisk as a part of the label text. How about removing the aria-hidden and let it read "First name star/asterisk" + add the text explaining that the asterisk represents a required field (although may be impractical to add onto every form)? I'm thinking the most important piece is having the required attribute on the element if it's a truly required field. The asterisk would be backup. 

I can try to squeeze into Tuesday's a11y office hours to get another opinion.

homing1: I was curious about how Fluent handles it. According to the example, it seems like they also skip aria-label when the asterisk is paired with a visible label. https://developer.microsoft.com/en-us/fluentui#/controls/web/textfield

site/src/components/form.md
jdanyow: I think validation should be considered part of this document since most/all forms will use validation such as required fields.

Can we simplify the markup a bit? Maybe something like

```html
<div class="field">
     <label class="field-label">
     <div class="field-body">
         ... input and decorations ...
     </div>
</div>
```

Would be good to have a radio/checkbox example.

site/src/components/form.md
```diff
@@ -0,0 +1,176 @@
...
+	<div class="field-section">
+		<label class="field-section-label field-section-label-lg label" for="sample-form-input-note">
+			Input
+		</label>
+		<input
+			id="sample-form-input-note"
+			name="title"
+			class="input"
+			type="text"
+			placeholder="e.g., Why is the sky blue?"
```
jdanyow, line 117: Placeholders result in poor accessibility for folks with low vision because they make fields appear to be filled in. Let's continue to discourage the use of placeholders on Docs.

site/src/components/form.md
```diff
@@ -0,0 +1,176 @@
...
+### Required Indicator
+
+For control elements that are required for form submission, you can add a span element with the `.required-indicator` class inside the label or adjacent to the element.
+
+```html
+<div class="field">
+	<p class="margin-bottom-xs">
+		Required fields are marked with an
+		<span class="required-indicator" aria-label="required"></span>
+	</p>
```
jdanyow, line 167: I know you're reviewing this with the accessibility team, just wanted to mention that this reads as "Required fields are marked with an required".

site/src/components/form.md
```diff
@@ -0,0 +1,176 @@
...
+			type="text"
+			placeholder="e.g., Why is the sky blue?"
+		/>
+	</div>
+</div>
+<div class="field display-flex">
+	<div class="field-section">
+		<button class="button button-primary margin-right-xxs">Submit</button>
+		<button class="button button-danger">Cancel</button>
+	</div>
```
jdanyow, line 73: Thinking the form buttons should use flex/utilities or a dedicated class rather than merging with field.

```html
<div class="display-flex gap-xxs">
  <button class="button button-primary">Submit</button>
  <button class="button button-danger">Cancel</button>
</div>
```

site/src/components/form.md
```diff
@@ -0,0 +1,176 @@
...
+`.field-section-icon-right` is used to display the icon on the right side of the element.
+
+```html
+<div class="field">
+	<div class="field-section field-section-icons field-section-icon-right">
+		<input class="input input-sm" placeholder="www.microsoft.com" />
+		<span class="icon border" aria-hidden="true"> </span>
+	</div>
+</div>
+```
```
jdanyow, line 156: Related to the comment above about reducing the markup to field, field-label, field-body:

Can we decouple input groups / input addons from "field". That seems like a separate thing.

site/src/components/form.md
BenDMyers: Ooh, this one is an interesting case.

Jeremy, is there a particular browser/screenreader combo you're seeing this announcement in? `aria-label` is only intended for interactive elements and landmarks, so while some browser/screenreader pairs might announce the label, that's definitely nonstandard, and I believe _most_ browser/screenreader combos would announce the span's contents.

The accessibility team recommended not doing anything special for the asterisks' markup -- just `<span class="required-indicator">` -- since the fields' `required` attributes will convey the required status. But in this case, where we have some help text disconnected from any inputs, I do think we might actually want to consider doing some markup trickery. I say this because [while JAWS and VoiceOver both announce "star" for asterisks by default, NVDA omits them entirely by default](https://www.deque.com/blog/dont-screen-readers-read-whats-screen-part-1-punctuation-typographic-symbols/#grammatical). Things get even trickier when folks have tweaked their verbosity settings. Because of this, when NVDA users and anyone with low-verbosity screenreader settings navigate to this help text, they'll likely be told `Required fields are marked with an`.

I might recommend something like:

```html
<p class="margin-bottom-xs">
	Required fields are marked with an
	<span class="required-indicator" aria-hidden="true"></span>
	<span class="visually-hidden">asterisk</span>
</p>
```

in this particular case.

homing1: I think it was reading that way because of the aria-label. For me, NVDA does read "star" as expected after removing the aria-label, but I've never changed my verbosity settings so I suppose I should explore that.

site/src/components/form.md
homing1: There's only one use case of addon in Docs and it seems to be misused, Perhaps it can just be removed.

site/src/components/form.md
homing1: Updated, but moved further down to the form validation section.

site/src/components/form.md
homing1: Removed

site/src/components/form.md
```diff
@@ -0,0 +1,130 @@
...
+	<div class="field-body field-body-icons field-body-icon-right">
+		<input class="input input-sm" />
+		<span class="icon border" aria-hidden="true"> </span>
+	</div>
+</div>
+```
+
+## Form Validation
+
+`.field-section-message` is placed after the element for form validation messages.
```
homing1, line 83: I'm thinking to update this section in the following form behavior PR since that component does not exist in this PR yet.

site/src/components/form.md
homing1: Changed it to just checkbox as there are other input examples below.

site/src/components/form.md
homing1: removed

BenDMyers (approved):

site/src/components/form.md
jdanyow: ok let's remove it here, I don't think input addons is part of the form field markup / css.

site/src/components/form.md
```diff
@@ -0,0 +1,130 @@
...
+
+```html
+<fieldset class="field border-none">
+	<legend class="field-label label margin-bottom-xs">Radio group</legend>
+	<p class="margin-bottom-xs">
+		Required fields are marked with
+		<span class="required-indicator"></span>
+	</p>
+	<div class="field-body">
+		<label class="radio field-label">
```
jdanyow, line 119: I don't think field-label belongs here. For checkboxes and radios the field-body element should contain the markup described on the [checkbox](https://design.docs.microsoft.com/components/checkbox.html) and [radio](https://design.docs.microsoft.com/components/radio.html) pages.

css/src/components/form/form.scss
```diff
@@ -0,0 +1,127 @@
...
+	.field-label {
+		margin-block-end: $spacer-3;
+
+		@include tablet {
+			flex: 1 0 0;
+		}
+	}
+
+	.field-body {
+		&.field-body-icons {
```
jdanyow, line 30: related to the other comment- let's remove field-body-icons. Input addons/icons can be part of the input styles if we need them in the future.

I think the hierarchy here is:

- .field: a `<div>` or a `<fieldset>`, depending on what the .field-body content is.
   - .field-label: a `<label>` or `<legend>`, depending on what the .field-body content is.
   - .field-body
      - anything... an [input](https://design.docs.microsoft.com/components/input.html), [radios](https://design.docs.microsoft.com/components/radio.html), [checkboxes](https://design.docs.microsoft.com/components/checkbox.html)

site/src/components/form.md
```diff
@@ -0,0 +1,130 @@
...
+`field-label` is used for the element's label.
+
+```html
+<fieldset class="field border-none">
+	<legend class="field-label label margin-bottom-xs" aria-describedby="drink-description">
+		Favorite drink
+	</legend>
+	<p class="field-description margin-bottom-xxs" id="drink-description">Which do you prefer?</p>
+	<div class="field-body">
+		<label class="checkbox margin-bottom-xxs">
```
jdanyow, line 34: Thinking `.checkbox:not(:last-child)` add the margin rather than relying on folks to add utilities.

site/src/components/form.md
```diff
@@ -0,0 +1,130 @@
...
+- custom form elements
+
+## Form Field
+
+The `.field` component is a container for form control elements to provide consistency.
+It can be used to hold `.field-body` containers, which holds the individual elements.
+`field-label` is used for the element's label.
+
+```html
+<fieldset class="field border-none">
```
jdanyow, line 28: lets remove border-none and move this over from docs-ui
https://dev.azure.com/ceapex/Engineering/_git/docs-ui?path=/packages/styles/src/base/_generic.scss&version=GBdevelop&_a=contents
```
fieldset {
	border: none;
	min-width: 0; // as a replaced element min-width:0 is required to enable it to work in responsive layouts
}
```

site/src/components/form.md
homing1: Removed

site/src/components/form.md
homing1: Removed

css/src/components/form/form.scss
homing1: We do have several use cases in docs so we do need some sort of replacement. I moved icons to inputs.

site/src/components/form.md
homing1: Added checkbox rule

site/src/components/form.md
```diff
@@ -0,0 +1,130 @@
...
+	<legend class="field-label label margin-bottom-xs" aria-describedby="drink-description">
```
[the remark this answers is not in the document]
homing1: I don't think we've implemented label in Docs yet. Updated `field-label` to use the same label mixin as `label`.

site/src/components/form.md
homing1: Added to bare-elements

site/src/components/form.md
```diff
@@ -0,0 +1,130 @@
...
+	<div class="field-body">
+		<label class="checkbox margin-bottom-xxs">
+			<input type="checkbox" />
+			<span class="checkbox-check" role="presentation" aria-hidden="true"></span>
+			<span class="checkbox-text" aria-hidden="true">Coca-cola</span>
+		</label>
+		<label class="checkbox">
+			<input type="checkbox" />
+			<span class="checkbox-check" role="presentation" aria-hidden="true"></span>
+			<span class="checkbox-text" aria-hidden="true">Pepsi</span>
```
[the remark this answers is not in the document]
homing1: Removed

css/src/components/form/form.scss
homing1: Removed icon changes from this PR and created a separate PR https://github.com/microsoft/atlas-design/pull/411

jdanyow (changes requested):

site/src/components/form.md
```diff
@@ -0,0 +1,97 @@
...
+	<div class="field">
+		<label class="field-label" for="sample-form-validation-input">Input</label>
+		<div class="field-body">
+			<input
+				id="sample-form-validation-input"
+				name="title"
+				class="input"
+				type="text"
+				aria-describedby="section-note"
+			/>
```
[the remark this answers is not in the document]
homing1: Updated

site/src/components/form.md
```diff
@@ -0,0 +1,97 @@
...
+		<div class="field-body">
+			<label class="radio">
+				<input checked name="question-1" type="radio" class="radio-dot" value="Yes" required />
+				<span class="radio-label-text">Yes</span>
+			</label>
+			<label class="radio">
+				<input checked name="question-1" type="radio" class="radio-dot" value="No" />
+				<span class="radio-label-text">No</span>
+			</label>
+			<span class="required-indicator"></span>
```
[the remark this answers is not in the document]
homing1: Updated

site/src/components/form.md
```diff
@@ -0,0 +1,97 @@
...
+For control elements that are required for form submission, you can add a span element with the `.required-indicator` class inside the label or adjacent to the element.
+
+```html
+<form>
+	<fieldset class="field">
+		<legend class="field-label">Radio group</legend>
+		<p class="margin-bottom-xs">
+			Required fields are marked with
+			<span class="required-indicator"></span>
+		</p>
```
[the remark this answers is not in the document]
homing1: Updated. Confirmed that NVDA reads "star" and Narrator reads "asterisk" no matter where the required indicator is placed , so I added both to the text.

site/src/components/checkbox.md
```diff
@@ -94,7 +51,7 @@ Here is an example of a group of checkboxes. Default spacing is added between ea
 ```html
 <label class="checkbox">
```
jdanyow, line 52: missing the input

site/src/components/form.md
```diff
@@ -0,0 +1,100 @@
...
+## Form Field
+
+The `.field` component is a container for form control elements to provide consistency.
+It can be used to hold `.field-body` containers, which holds the individual elements.
+`field-label` is used for the element's label.
+
+```html
+<fieldset class="field">
+	<legend class="field-label">Field label</legend>
+	<p class="field-description" id="drink-description">Field description</p>
```
jdanyow, line 30: nit id not needed

site/src/components/form.md
```diff
@@ -0,0 +1,100 @@
...
+</fieldset>
+```
+
+## Form Validation
+
+`.field-error` is placed after the element for form validation error messages.
+
+```html
+<form>
+	<p class="margin-bottom-xs visually-hidden">Required fields are marked with asterisk/star</p>
```
jdanyow, line 52: nit margin not needed

site/src/components/form.md
```diff
@@ -0,0 +1,100 @@
...
+</form>
+```
+
+### Required Indicator
+
+For control elements that are required for form submission, you can add a span element with the `.required-indicator` class inside the label or adjacent to the element.
+
+```html
+<form>
+	<p class="margin-bottom-xs visually-hidden">Required fields are marked with asterisk/star</p>
```
jdanyow, line 85: nit margin not needed

jdanyow (dismissed):

site/src/components/form.md
homing1: removed

site/src/components/form.md
homing1: removed

site/src/components/form.md
homing1: removed

jdanyow (approved):

BenDMyers (approved):

PR #411, opened 2022-06-08
Add input icons

homing1:
Task: task-[632162](https://dev.azure.com/ceapex/Engineering/_sprints/taskboard/Docs/Engineering/?workitem=632162)

Link: preview-[411](https://design.docs.microsoft.com/pulls/411/components/input.html)

This PR adds icons to inputs.

## Testing

1. Visit https://design.docs.microsoft.com/pulls/411/components/input.html to review example.

### DIFF
--- a/css/src/components/form/input.scss
+++ b/css/src/components/form/input.scss
@@ -16,6 +16,14 @@ $input-success-border-color: $success !default;
 
 $input-focus-box-shadow-size: 0 0 0 0.125rem !default;
 
+$input-icon-size: 2.25em !default;
+$input-icon-color: $secondary-background !default;
+$input-icon-active-color: $secondary !default;
+
+$input-icon-font-size: $font-size-7 !default;
+$input-icon-font-size-sm: $font-size-8 !default;
+$input-icon-font-size-lg: $font-size-6 !default;
+
 .input {
 	@include control;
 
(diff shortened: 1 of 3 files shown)


site/src/components/input.md
```diff
@@ -45,6 +45,19 @@ Here is an example of a standard usage of input paired with a label.
...
+<input class="input input-danger" type="text" />
+```
+
+### Modifiers
+
+Use `input-icon` on the input in combination with a `.icon` sibling to add an icon to the input. If the icon needs to displayed at the end of the input, use `input-icon-right`.
+
+```html
+<div class="position-relative margin-bottom-xxs">
+	<input class="input input-icon" type="text" /> <span class="icon border"></span>
```
wibjorn, line 58: If possible, would love to see an example with an inlined svg here as well, since that is how we do this on the site. I was messing in the dev tools with this and wasn't able to get the result I wanted - but that could be paternity leave brain 😆 ...

css/src/components/form/input.scss
```diff
@@ -72,4 +80,65 @@ $input-focus-box-shadow-size: 0 0 0 0.125rem !default;
 			box-shadow: $input-focus-box-shadow-size $input-success-border-color;
 		}
 	}
+
+	&.input-icon-right {
```
wibjorn, line 84: image

In the example, I'm still seeing an indented cursor when the icon is on the right.

homing1: Should be removed now

site/src/components/input.md
homing1: I ended up directly taking the svg from docons to show the icon.

css/src/components/form/input.scss
```diff
@@ -72,4 +80,69 @@ $input-focus-box-shadow-size: 0 0 0 0.125rem !default;
...
+
+		&.input-icon-right {
+			padding-inline-start: unset;
+		}
+
+		+ .icon {
+			position: absolute;
+			inset-block-start: 0;
+			inset-inline-start: 0;
+			width: $input-icon-size;
```
wibjorn, line 119: Should sizes be done in `em` to ensure scaling? i.e. should we remove the variable declarations on 23-25 in favor of em to ensure this always scales, regardless of an input's size?

homing1: Removed the icon font size variables and updated the `.icon` container font size instead. The svg defaults to 1em already. I think it's better to not force the user to use another font size helper class in this scenario to avoid weird accidental pairings (like input-sm with font-size-xxl). Adding `input sm/lg` should take care of the rest.

css/src/components/form/input.scss
```diff
@@ -72,4 +76,66 @@ $input-focus-box-shadow-size: 0 0 0 0.125rem !default;
...
+		}
+
+		@include focus-visible {
+			+ .icon {
+				color: $input-icon-active-color;
+			}
+		}
+
+		&.input-sm {
+			~ .icon {
```
wibjorn, line 130: applies font size?

css/src/components/form/input.scss
```diff
@@ -72,4 +76,66 @@ $input-focus-box-shadow-size: 0 0 0 0.125rem !default;
...
+				@include control-sm;
+			}
+		}
+
+		&.input-lg {
+			~ .icon {
+				@include control-lg;
```
wibjorn, line 137: applies font size?

css/src/components/form/input.scss
homing1: updated

css/src/components/form/input.scss
homing1: updated

wibjorn (approved):

BenDMyers (approved):

PR #418, opened 2022-07-06
Add hash reload

homing1:
Task: task-[654492](https://dev.azure.com/ceapex/Engineering/_sprints/taskboard/Docs/Engineering/2022-06-26?workitem=654492)

Link: preview-[418](https://design.docs.microsoft.com/pulls/418/components/form.html)

This PR adds hash reload to form behavior. The goal is to add hash, then reload the page showing the updated info.

## Testing

1. Review code changes.
2. Visit https://design.docs.microsoft.com/pulls/418/components/input.html to verify form behavior still works.

### DIFF
--- a/js/src/elements/form-behavior.ts
+++ b/js/src/elements/form-behavior.ts
@@ -124,7 +124,7 @@ class FormBehaviorElement extends HTMLElement {
 
 	navigate(href: string | null) {
 		if (href) {
-			navigateAfterSubmit(href, this.getAttribute('navigation'));
+			navigateAfterSubmit(href, this.getAttribute('navigation') as NavigationSteps);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


js/src/elements/form-behavior.ts
```diff
@@ -691,6 +680,16 @@ export function navigateAfterSubmit(href: string, navigationStep: string | null)
 				location.href = href;
 			}
 			break;
+		case 'hash-reload':
```
BenDMyers, line 683: Would you be open to setting the valid strings explicitly in this function's signature?

I'm thinking something like:
```ts
export function navigateAfterSubmit(
    href: string,
    navigationStep: 'follow' | 'hash-reload' | 'replace' | 'reload' | null
) {
```

This way, as we're invoking `navigateAfterSubmit`, we can get intellisense as to which options are available.

BenDMyers: (Feel free to disregard or table this for later)

homing1: I think that's fine, had to add it as a type though.

BenDMyers (approved):

ollips (approved):

PR #422, opened 2022-07-20
Add form behavior custom event

homing1:
Task: task-[662991](https://dev.azure.com/ceapex/Engineering/_workitems/edit/662991/)

Link: preview-[422](https://design.docs.microsoft.com/pulls/422/components/form.html)

This PR adds two custom events for custom validation and error handling. Users can subscribe to the `form-validating` custom event and implement custom validation and validation error handling (such as tags). User can subscribe to the `submission-error` custom event to receive data about failed responses and implement custom error handling.

## Testing

1. Review code.
2. Visit https://design.docs.microsoft.com/pulls/422/components/form.html . There should be no tags in the form. Click the submit button. The existing validation for other elements should still work.


js/src/elements/form-behavior.ts
```diff
@@ -168,23 +170,25 @@ class FormBehaviorElement extends HTMLElement {
...
-			this.dispatchEvent(validationErrorEvent);
 			return;
 		}
 
 		try {
 			this.submitting = true;
 			setBusySubmitButton(form, this.submitting);
-			const valid = await this.validateForm(form);
-			if (!valid.valid) {
+			const result = await this.validateForm(form);
```
jdanyow, line 183: I think we need to remove the attachment validation logic entirely, including the loc string and use an event to externalize the custom validation.

Inside `this.validateForm(form)` we could fire an event:

```ts
validateForm(form) {
  ... current logic to get list of errors ...
  
  // new event:
  this.dispatchEvent(new CustomEvent('form-validating', { detail: { form, errors }, bubbles: true });

  // current logic to return result
  return ...
}
```

This would enable subscribers of the event to push their own custom errors into the errors array to augment the standard validation, removing the need for us to handle tag or attachment internals.

jdanyow (changes requested):

js/src/elements/form-behavior.ts
homing1: Removed everything tag related.

homing1: Also added another custom event for clearing validation errors so that the subscriber will know when to remove the custom errors.

js/src/elements/form-behavior.ts
```diff
@@ -396,14 +366,7 @@ class FormBehaviorElement extends HTMLElement {
...
-			this.runBasicValidation(
-				input,
-				displayValidity,
-				errors,
-				errorList,
-				isTagSelector,
-				isCustomElement
-			);
+			this.runBasicValidation(input, displayValidity, errors, errorList, isCustomElement);
 		}
```
jdanyow, line 371: I think you should fire the form-validating event here, this way the external code has an opportunity to insert validation errors into the errors array prior to us checking `if (errors.length === 0)` on the next line. This should also remove the need to dispatch a clear-validation-errors event.

jdanyow (changes requested):

js/src/elements/form-behavior.ts
homing1: Moved the event firing. I think we still need a way to signal when to clear the custom errors though. For normal inputs, we're clearing the errors on input and change events, so it's before resubmission of the form, hence before the `form-validating` event fires again. I think we should keep that the same even for elements with custom validation. I think it would become confusing to show a stale error message if we wait to clear it until resubmission and revalidation for these custom validation inputs, but have a different behavior for normal inputs.

Tags involve hidden inputs and sibling elements, so in the external code, I'll need to re-implement the tag specific code that I'm removing in this PR.

jdanyow: totally makes sense- would it be resolved if the tag control fired change events when it's value changed?

jdanyow (approved):

js/src/elements/form-behavior.ts
```diff
@@ -379,8 +359,7 @@ class FormBehaviorElement extends HTMLElement {
 				continue;
 			}
 
-			const isTagSelector = input.classList.contains('tag-input');
-			if (input.hasAttribute('aria-hidden') && !isTagSelector) {
+			if (input.hasAttribute('aria-hidden')) {
```
BenDMyers, line 362: I feel like this is work for another time, but should we explicitly check if `aria-hidden` is `"true"`? It seems to me that `aria-hidden="false"` would trip this up.

homing1: Good point. Let's table this for a follow up PR so it won't block the integration into Docs.

BenDMyers: Makes sense to me!

BenDMyers (approved):

js/src/elements/form-behavior.ts
homing1: The existing tag control in Docs does fire change events on value change so it will trigger the clearing function. The clearing function no longer handles tag, but with the custom event, I can listen for it and clear the error for tags or any other custom input. It didn't make sense to keep the tag specific error handling around when we're removing the tag validation altogether.

PR #426, opened 2022-07-26
wip [nomerge]

wibjorn:
Task: task-[work-item-number]

Link: preview-426 > Patterns > Side scrolling container

No need to merge this. Just an ugly proof of concept at this point.

https://design.docs.microsoft.com/pulls/426/patterns/side-scrolling-container.html

## Testing

1. Step one
2. Step two
   Expected result:

## Additional information

[Optional]

### DIFF
--- a/site/src/patterns/side-scrolling-container.md
+++ b/site/src/patterns/side-scrolling-container.md
@@ -1,0 +1,48 @@
+---
+title: Side scrolling container
+description: Side scrolling container pattern in the Atlas Design System
+template: standard
+---
+
+# Side scrolling container
+
+WIP
+
+```html
+<div class="scroll-horizontal scroll-snap-container display-flex">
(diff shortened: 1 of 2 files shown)


wibjorn:
Notes from discussion: 
- if cards were to have focusable elements, especially multiple focusable elements, it would likely be ideal if other cards not in view were marked as inert. desired keyboard focus behavior would be to tab through one card only, then hit navigation.
- it's a bummer to have horizontal scroll when you click a jump link, maybe we need to prevent default and use `scrollTo` with smooth option instead
- navigation should be well labeled for screenreader users; we probably shouldn't mark it as area hidden because it's possible low vision focus might want items scrolled. This is debatable to me, but worth trying for now. Aria label could be `scroll card <n> into view` or something like that.
- didn't address high lighting of active card

PR #427, opened 2022-07-28
Hero component

ollips:
Task: task-659504

Link: preview-427

This pull request adds a hero and several pieces of documentation.

- Adds a hero component
-  Adds a "Full screen" content section button to the site, so that heroes can properly be tested
- Creates a layout gap placeholder class
- Updates parcel transformer to include an html code block that can be a full bleed when the content section is full screens
- Adds page for background-image pattern atomics

## Testing

https://design.docs.microsoft.com/pulls/427/components/hero.html
https://design.docs.microsoft.com/pulls/427/patterns/hero.html
https://design.docs.microsoft.com/pulls/427/atomics/background-pattern.html

1. Step one
2. Step two
   Expected result:


css/src/components/hero.scss
```diff
@@ -0,0 +1,153 @@
...
+
+		@include desktop {
+			width: $hero-content-fluid-width;
+			min-height: $hero-min-height-default;
+			padding-block: $hero-padding-default;
+			padding-inline-end: $hero-padding-sm;
+		}
+	}
+
+	&:not(.hero-image) {
```
ollips, line 37: _Reviewers note_: I've realized it might be very useful to fix the content width to the max reading width as it initially was if there is no image on the right side. Any thoughts?

homing1: I think that's fine.

css/src/components/hero.scss
```diff
@@ -0,0 +1,153 @@
...
+$hero-padding-sm: $layout-5 !default;
+$hero-padding-default: $layout-6 !default;
+$hero-padding-lg: $layout-7 !default;
+
+$hero-min-height-xs: 200px !default;
+$hero-min-height-sm: 300px !default;
+$hero-min-height-default: 400px !default;
+$hero-min-height-lg: 500px !default;
+
+$hero-content-width: 45% !default;
```
homing1, line 11: this variable seems unused?

homing1, line 11: `$hero-content-width` seems unused?

css/src/components/hero.scss
```diff
@@ -0,0 +1,153 @@
...
+.hero {
+	@extend %layout-gap;
+
+	display: flex;
+	position: relative;
+	flex-direction: column;
+	align-items: stretch;
+	justify-content: space-between;
+
+	.hero-content {
```
homing1, line 25: Any thoughts on vertically center the content of `hero-content` by default and the user can change it if they want to?  
image

ollips: That's what bothers me now after the bottom gradient is gone. Processing

css/src/components/hero.scss
ollips: Fixed!

site/src/patterns/hero.md
```diff
@@ -0,0 +1,53 @@
...
+<!--
+## Accent hero
+
+Hero with a theme-aware companion image.
+
+```html-no-indent
+<section
+    class="hero hero-image hero-image-gradient-accent-right background-color-body-accent"
+    style="
+        --hero-background-image-light: url('./scaffold/media/hero-400-light.jpg');
```
ollips, line 40: _Reviewers note:_ commented out this demo because need the full path URL to the images.

css/src/components/hero.scss
ollips: @wibjorn any thoughts on this?

wibjorn: Looks like that is not intended to me

image

wibjorn: Looks like that is not intended based on the Figma

css/src/components/hero.scss
```diff
@@ -0,0 +1,153 @@
...
+			object-fit: cover;
+		}
+	}
+}
+
+.hero-image {
+	&::before {
+		display: none;
+		position: absolute;
+		top: 0;
```
wibjorn, line 112: `inset-block-start`?

css/src/components/hero.scss
```diff
@@ -0,0 +1,153 @@
...
+		.theme-dark & {
+			--hero-background-image: var(--hero-background-image-dark);
+		}
+
+		.theme-high-contrast & {
+			--hero-background-image: none;
+		}
+	}
+
+	&.hero-image-gradient-accent-right {
```
wibjorn, line 147: Yikes on the name, can put this in a new file or something?

css/src/components/hero.scss
```diff
@@ -0,0 +1,153 @@
...
+		img {
+			width: 100%;
+			height: 100%;
+			object-fit: cover;
+		}
+	}
+}
+
+.hero-image {
+	&::before {
```
wibjorn, line 109: reduce level of nesting by doing `.hero-image::before`?

css/src/components/hero.scss
ollips: Keeping as is for now

css/src/components/hero.scss
wibjorn: Or maybe we need to do some more CSS variable trickery here? Would require some other extensions, but at least we shouldn't be tied necessarily to this one color.

```scss
background-image: background-image: linear-gradient(90deg, transparent 90%, var(--hero-gradient-color), 100%),
				var(--hero-background-image);
```

Then you'd have

```scss
.hero-gradient-color-accent {
   $body-background-accent
}
```

ollips: Any suggestions on a better name? I also feel like it has to live in this file because it is tightened to the hero image

wibjorn: Will have to spend some time thinking this one over, seems like we've got a single use class in here, which isn't good. It only works with one background color and doesn't seem extensible.

site/src/scaffold/index.ts
```diff
@@ -1,6 +1,6 @@
 import { handleCodeFilters } from './scripts/code-filter';
 import { handleFigmaFullScreenRequest } from './scripts/figma-embed';
-import { initTheme } from './scripts/theming';
+import { initTheme, ThemeTypeMap } from './scripts/theming';
```
wibjorn, line 3: whoops, this one can be removed it looks like

css/src/components/hero.scss
wibjorn: I think we should likely revert this change. I would prefer to keep the widths shorter, and retain consistency with the view with the image included.

wibjorn: Can always make it later, after discussion with design team.

css/src/components/hero.scss
ollips: Is it really needed? 🤔

site/src/patterns/hero.md
wibjorn: Works fine with the correct file paths (note `../`):
```
 --hero-background-image-light: url('../scaffold/media/hero-400-light.jpg');
        --hero-background-image-dark: url('../scaffold/media/hero-400-dark.jpg');

File: site/src/scaffold/index.ts (hunk 8 above)
Comment (homing1) [reply site/src/scaffold/index.ts]: Removed

File: css/src/components/hero.scss (hunk 7 above)
Comment (homing1) [reply css/src/components/hero.scss]: Looking at the example, it seems like `hero-image` is meant as a subclass, so I nested it within `.hero`.

File: css/src/components/hero.scss (hunk 5 above)
Comment (homing1) [reply css/src/components/hero.scss]: Updated to `inset`

File: site/src/patterns/hero.md (hunk 4 above)
Comment (homing1) [reply site/src/patterns/hero.md]: Updated and uncommented the example.

**Review by wibjorn — APPROVED** [review]

**Review by wibjorn — DISMISSED** [review]

File: css/src/components/hero.scss [hunk 9]
```diff
@@ -0,0 +1,141 @@
...
+		@include tablet {
+			padding: $spacer-10;
+		}
+
+		@include desktop {
+			border-right: 1px solid $border;
+			border-bottom: 1px solid $border;
+			border-left: 1px solid $border;
+			border-bottom-left-radius: $border-radius-lg;
+			border-bottom-right-radius: $border-radius-lg;
```
BenDMyers, line 85: Should these use logical properties for the border?

plugins/parcel-transformer-markdown-html/lib/index.js
```diff
@@ -19,7 +19,8 @@ const languageDisplayNames = {
 	markdown: 'Markdown',
 	'atomics-filter': 'Atomics',
 	atomics: 'Atomics',
-	'abut-html': 'HTML'
+	'abut-html': 'HTML',
+	'html-no-indent': 'HTML'
```
BenDMyers, line 23: Totally okay if no, but is there a chance we could order the language name and modifier consistently? Maybe `{language}-{modifier}`, like `html-abut` and `html-no-indent`?

css/src/components/hero.scss
wibjorn: border-bottom-left-radius -> border-end-start-radius
border-bottom-right-radius -> border-end-end-radius

plugins/parcel-transformer-markdown-html/lib/index.js
wibjorn: You're definitely right @BenDMyers, I'd say for now why don't we consider updating `abut-html` in a follow up.

BenDMyers: Works for me!

wibjorn (approved):

css/src/components/hero.scss
```diff
@@ -0,0 +1,141 @@
...
+		z-index: 1;
+
+		@include tablet {
+			padding: $spacer-10;
+		}
+
+		@include desktop {
+			border-right: 1px solid $border;
+			border-bottom: 1px solid $border;
+			border-left: 1px solid $border;
```
BenDMyers, line 83: @wibjorn Do these three lines need logical props as well?

wibjorn (approved):

BenDMyers (approved):

wibjorn (approved):

PR #430, opened 2022-07-28
Export form behavior

homing1:
Task: task-[680557](https://dev.azure.com/ceapex/Engineering/_workitems/edit/680557), [680541](https://dev.azure.com/ceapex/Engineering/_workitems/edit/680541/), [680544](https://dev.azure.com/ceapex/Engineering/_workitems/edit/680544/)

Link: preview-430

This PR allows a few form behavior items for export and adds a test for when form is submitted without edits. It also removes star rating and two tests for form behavior. 

## Testing

1. Visit https://design.docs.microsoft.com/pulls/430/components/form.html . You should no longer see star rating in the form behavior section.


site/src/components/form.md
```diff
@@ -172,28 +172,11 @@ The form behavior component can accept certain HTML attributes.
 				minlength="10"
 				rows="4"
 				cols="30"
+				required
```
ollips, line 175: I see the `textarea` was marked as required but when I'm trying to submit a form it doesn't highlight it with red border or anything

image
 
And it passes the empty submission

image

ollips: Oh, I didn't notice that initially but it looks like the textarea contains some empty spaces by default

homing1: Made an update to remove the empty spaces

ollips (dismissed):

integration/tests/form-behavior.spec.ts
```diff
@@ -39,37 +39,6 @@ test.describe('form behavior validation', () => {
 		});
 	});
 
-	// The input value string is automatically reduced if input has maxlength
```
homing1, line 42: For a normal input, setting the `maxlength` attribute will automatically limit the number of characters to its value so this test is redundant.

jdanyow (approved):

js/src/elements/form-behavior.ts
```diff
@@ -358,7 +356,7 @@ class FormBehaviorElement extends HTMLElement {
 				continue;
 			}
 
-			if (input.hasAttribute('aria-hidden')) {
+			if (input.hasAttribute('aria-hidden') === true) {
```
wibjorn, line 359: curious why this change was necessary, seems like same with but with more characters

homing1: In an earlier PR, Ben made a comment on how `aria-hidden="false"` would trip this up. Perhaps it's an unlikely scenario to explicit set the attribute to false, but I suppose it can still happen.

.changeset/sweet-walls-sing.md
```diff
@@ -0,0 +1,7 @@
+---
+'@microsoft/atlas-integration': minor
```
wibjorn, line 2: patch patch patch

wibjorn: nvm, other minors changes are coming in so we can ignore

wibjorn (approved):

PR #435, opened 2022-08-07
Form behavior updates

homing1:
Task: task-692607, task-687169, task-691911

Link: preview-435

This PR ensures that there is bottom margin for the form behavior error container for all scenarios. Bottom spacing was accidentally removed recently, which affected the no edit scenario. Logic is also added to dispatch the `form-validating` event for the no edit scenario. `.form-error-container` is replaced with atomics and data attribute `data-form-error-container`. Updated form behavior logic to use `data-field-error` to make `.field-error` solely for styling. Added a small section on custom validation in the documentation.

## Testing

1. Visit https://design.docs.microsoft.com/pulls/435/components/form.html . Scroll down to form behavior section. Inspect the form to open dev tools, remove the `new` attribute from the form behavior and submit the form. There should be some spacing below the error summary at the top of the form. Type in an input and submit. The same bottom spacing should still exist.
2. There should be a warning icon in the error summary container. 
image
3. Make some inputs and submit again, should still work the same as before.


site/src/scaffold/styles/form.scss
```diff
@@ -3,12 +3,6 @@
 	height: 2.25rem;
 }
 
-.form-error-container {
```
homing1, line 6: Moving this class to form.scss

css/src/components/form/form.scss
```diff
@@ -52,7 +52,8 @@ $field-error-color: $danger !default;
 		}
 	}
 
-	.field-error {
+	.field-error,
+	.field-error-custom {
```
wibjorn, line 56: just for my edification, if `.field-error` and `.field-error-custom`  are the same thing, why do they both exist? Why not use only one?

css/src/components/form/form.scss
```diff
@@ -65,3 +66,9 @@ $field-error-color: $danger !default;
 	content: '*';
 	vertical-align: top;
 }
+
+.form-error-container {
```
wibjorn, line 70: Just musing here, resolve as you please, but part of me thinks that we should implement [banners](https://www.figma.com/file/uVA2amRR71yJZ0GS6RI6zL/%F0%9F%8C%9E-Atlas-Design-Library?node-id=1002%3A7106)

wibjorn: It's also worth nothing, this particular class can easily be made with atomics. Do we need the class?
```
color-danger border-color-danger border
```

css/src/components/form/form.scss
homing1: The basic validation in form behavior looks for the first `.field-error` class and updates its text content with a message when there's an error, or empties it if validation passes. That's ok when there's one input within a field as there can only be one message. 

For markdown editor, there's both a text area and an input so there's a potential of two messages, one for the text area and one for the input. The input requires custom validation outside of the basic validation provided by form behavior. The custom validation is run once `form-validating` event is dispatched, its message is applied to `.field-error`. Then, the basic validation runs for the text area, finds the same `.field-error` that already has a message, and overwrites it with the message for the text area. Using a different class for the custom validation error message would avoid this scenario.

homing1: It's possible for the custom validation error message to be something like this instead: 

`<p class="field-error">Textarea must have at least 10 characters."</p>`
`<p class="color-danger font-size-sm line-height-something">Maximum number of attachments exceeded. Please select up to 1 attachments.</p>`

vs.

`<p class="field-error">Textarea must have at least 10 characters."</p>`
`<p class="field-error-custom">Maximum number of attachments exceeded. Please select up to 1 attachments.</p>`

css/src/components/form/form.scss
homing1: You're absolutely right. Originally, I added this class because we haven't implemented banners yet. Replaced with atomics for now, as I'd rather not hastily try to implement banners, given the timeline.

css/src/components/form/form.scss
homing1: Chatted with Will, decided to completely decouple `.field-error` from the form behavior functionality, use data attributes instead. Reverted `.field-error-custom`

js/src/elements/form-behavior.ts
```diff
@@ -375,6 +384,11 @@ export class FormBehaviorElement extends HTMLElement {
 				continue;
 			}
 
+			// Don't check markdown editor attachment input
```
wibjorn, line 387: No need to hold this, but consider adding an attribute instead of a class. `data-skip-validation` or something similar.

wibjorn (dismissed):

js/src/elements/form-behavior.ts
homing1: Updated to check its id for now. I'll have to find the markdown editor codebase to add data attributes later.

wibjorn (approved):

wibjorn (approved):

ollips (approved):

PR #437, opened 2022-08-09
update snap scroll slide height

jim-benson:
Task: task-[688402](https://dev.azure.com/ceapex/Engineering/_sprints/taskboard/Assessments/Engineering/2022-07-24?workitem=688402)

Link: preview-[437]()

update snap scroll slide height

### DIFF
--- a/css/src/components/scroll.scss
+++ b/css/src/components/scroll.scss
@@ -21,6 +21,6 @@
 .scroll-snap-item {
 	flex-basis: 90%;
 	min-width: 80%;
-	height: auto;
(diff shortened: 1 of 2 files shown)


css/src/components/scroll.scss
```diff
@@ -21,6 +21,6 @@
 .scroll-snap-item {
 	flex-basis: 90%;
 	min-width: 80%;
-	height: auto;
+	height: auto !important; // make sure card height is full height of container
```
wibjorn, line 24: Think we can abandon this one, since we've figured a way around it

PR #451, opened 2022-08-23
add resource card pattern and image component

wibjorn:
Task: task-695887

Link: preview-451

https://design.docs.microsoft.com/pulls/451/components/image.html
https://design.docs.microsoft.com/pulls/451/patterns/card.html

## Testing

1. View first link, read through.
2. View second link, Resources card section. Read through.

## Additional information

image


site/src/components/image.md
```diff
@@ -0,0 +1,71 @@
...
+template: standard
+---
+
+# Image
+
+The image component parents an image tag or svg and provides width based classes to contrain its children.
+
+## Usage
+
+Here is an example of a standard usage of label paired with an input. The default image class takes up the full width of its container.
```
ollips, line 13: Remove the first sentence :)

.changeset/flat-hats-beam.md
```diff
@@ -0,0 +1,6 @@
+---
+'@microsoft/atlas-site': minor
+'@microsoft/atlas-css': patch
+---
+
+Add resource card pattern to site documenation.
```
homing1, line 6: "documentation"

site/src/components/image.md
```diff
@@ -0,0 +1,71 @@
...
+
+```html
+<div class="image">
+	<img src="~/src/scaffold/media/video-media.png" />
+</div>
+```
+
+## Sizes
+
+The following classes are available for resizing square images.
```
ollips, line 23: What do you think about including the small table and listing all the available sizes? I think it might be easier than checking the code example

site/src/components/image.md
wibjorn: Guess which file I copied this from ... :D

wibjorn: updated

site/src/components/image.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Image
+description: Image component in the Atlas Design System
+template: standard
+---
+
+# Image
+
+The image component parents an image tag or svg and provides width based classes to contrain its children.
```
homing1, line 9: "constrain"?

wibjorn: good catch, updated!

site/src/components/image.md
wibjorn: Added, see what you think

.changeset/flat-hats-beam.md
wibjorn: good catch, updated!

site/src/patterns/card.md
```diff
@@ -47,3 +47,29 @@ This card type contains a super title labelled by the card's type, a title, an i
 	</div>
 </div>
 ```
+
+## Resource card
+
+This card pattern makes use of the outer `.card` class and includes the use of the [`.media`](../components/media.md) and [`.image`](../components/image.md) components inside it. Also, because there is only one clickable element inside, it uses the [`.stretched-link` component](../components/stretched-link.md)
+
+```html
+<article class="card position-relative">
```
ollips, line 56: I think `position-relative` isn't needed here, is it?

wibjorn: See stretched link documenation

ollips: Oops, my bad, didn't notice it contains the stretched link 🤦🏻‍♀️

site/src/components/image.md
ollips: Do you think people might want to mix sizes and do smth like `16x32` by seeing this pattern? Would it be better to list sizes as `16x16, 24x24, etc`? Not insisting, just sharing thoughts.

css/src/components/image.scss
```diff
@@ -0,0 +1,44 @@
...
+.image {
+	display: block;
+	position: relative;
+
+	img,
+	svg {
+		display: block;
+		width: 100%;
+		height: auto;
+		max-height: 100%;
```
ollips, line 12: Is it needed? 🤔

site/src/components/image.md
```diff
@@ -0,0 +1,75 @@
...
+
+The default image class takes up the full width of its container, but various sizes and screensizes modifiers are available.
+
+| Class pattern                | Sizes                              | screensize          |
+| ---------------------------- | ---------------------------------- | ------------------- |
+| `.image.image-<size>x<size>` | 16 24 32 36 48 64 76 88 96 112 128 | `tablet`, `desktop` |
+
+```html
+<div class="image">
+	<img src="~/src/scaffold/media/video-media.png" />
```
ollips, line 21: Add `alt=""` to the examples so there are no red issues highlights in the browser

site/src/components/image.md
wibjorn: I don't think we should open pixel sizing up to non-square ratios, and the description mentions that they're square

css/src/components/image.scss
wibjorn: Likely a browser compat thing, I think we should keep it

site/src/components/image.md
wibjorn: You know what, nevermind. I like how explicit your idea is, going to update

css/src/components/image.scss
ollips: ok, just wanted to double check :)

ollips (approved):

homing1 (approved):

PR #454, opened 2022-08-24
Remove bonus classes / move margin from holder div to child

disimmon:
Task: task-[work-item-number]

Link: [preview-454](https://github.com/microsoft/atlas-design/pull/454)

The margin on the parent error container, which is always present, is missaligning elements in the UI.
- Removes margin and background classes from parent container
- Adds margin to alert, alert already has background / border applied

## Testing

Visit https://design.docs.microsoft.com/pulls/454/components/form.html and trigger an error. The spacing / background should be correct.

### DIFF
--- a/js/src/elements/form-behavior.ts
+++ b/js/src/elements/form-behavior.ts
@@ -65,7 +65,6 @@ export class FormBehaviorElement extends HTMLElement {
 
 		form.setAttribute('novalidate', '');
 		const errorSummaryContainer = document.createElement('div');
-		errorSummaryContainer.classList.add('margin-bottom-sm', 'background-color-danger-light');
 		errorSummaryContainer.setAttribute('data-form-error-container', '');
 		this.insertAdjacentElement('afterend', errorSummaryContainer);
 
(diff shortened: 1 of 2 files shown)


wibjorn (approved):

js/src/elements/form-behavior.ts
```diff
@@ -65,7 +65,6 @@ export class FormBehaviorElement extends HTMLElement {
 
 		form.setAttribute('novalidate', '');
 		const errorSummaryContainer = document.createElement('div');
-		errorSummaryContainer.classList.add('margin-bottom-sm', 'background-color-danger-light');
```
homing1, line 68: Here's the correct preview link https://design.docs.microsoft.com/pulls/454/components/form.html

Until we implement banners in Atlas, let's keep the background color class for now.

disimmon: Yeah, previewed it and realized we need it. Moved it to the `alert` class.

disimmon: I'm guessing alot fo this will change with banners so trying to do the minumum adjust here.

homing1: Correct, as the current `.alert` in Docs doesn't quite match the new banner designs anyway.

wibjorn (approved):

homing1 (approved):

PR #462, opened 2022-08-29
update .card-template to have flex-grow

wibjorn:
Link: preview-462

No diff in its original context: https://design.docs.microsoft.com/pulls/462/components/card.html

Recent bug was unearthed that shows if a card with .card-template is used in a auto grid type of setting, then the footer is not pushed down to the bottom of the card.


btelnes (approved):

BenDMyers (approved):
Looks great to me!

PR #481, opened 2022-10-27
Ensure forced colors mode doesn't interfere with fill: currentColor

BenDMyers:
Task: task-[555516](https://ceapex.visualstudio.com/Engineering/_workitems/edit/555516)

Link: preview-[481](https://design.learn.microsoft.com/pulls/481/components/input.html)

Sometimes, the operating system forced colors mode and `fill: currentColor;` will clash, where the browser will show an SVG path in its non-forced color. If that non-forced color doesn't stand out against the forced color background, that part of the SVG won't be visible.

Applying `forced-color-adjust: auto` to the SVG path makes the browser use the forced colors decided by the parent, ensuring the SVG displays in colors suitable for the user.

## Testing

1. Navigate to https://design.learn.microsoft.com/pulls/481/components/input.html
2. Cycle through Windows' contrast themes.
   Expected result: In every theme, the _Microsoft_ wordmark in the logo at the top of the page will be visible, using whichever color is used for hyperlinks. For inputs with icons, the icons continue to display.

## Additional information

[Optional]

### DIFF
--- a/site/src/scaffold/styles/form.scss
+++ b/site/src/scaffold/styles/form.scss
@@ -7,8 +7,8 @@
 	margin-inline-end: $spacer-3;
 
 	svg {
-		fill: currentColor;
-		forced-color-adjust: auto;
+		@include fill-current-color();
+
 		width: 1rem;
 		height: auto;
 	}
--- a/site/src/scaffold/styles/form.scss
+++ b/site/src/scaffold/styles/form.scss
@@ -8,6 +8,7 @@
 
 	svg {
 		fill: currentColor;
+		forced-color-adjust: auto;
 		width: 1rem;
 		height: auto;
 	}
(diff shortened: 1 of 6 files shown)


site/src/scaffold/styles/form.scss
```diff
@@ -8,6 +8,7 @@
 
 	svg {
 		fill: currentColor;
```
wibjorn, line 10: Should also use the mixin?

BenDMyers: Updated

site/src/scaffold/styles/form.scss
```diff
@@ -8,6 +8,7 @@
...
+		forced-color-adjust: auto;
 		width: 1rem;
 		height: auto;
 	}
```
jdanyow, line 14: should the mixin be used here too?

jdanyow (approved):

site/src/scaffold/styles/form.scss
BenDMyers: This has been updated

homing1 (approved):

PR #483, opened 2022-10-27
ensure if statement is treated as an expression not a string

wibjorn:
Must treat the release condition as an expression rather than a string. To do that you must surround the expression in `${{ }}`


BenDMyers (approved):
yolo

ollips (approved):

PR #487, opened 2022-11-10
add check to keep is-loading class through page reload

ncmarsh:
Task: task-[694873](https://dev.azure.com/ceapex/Engineering/_workitems/edit/694873)

Link: preview-487

This PR adds a check to keep the is-loading icon if response is ok on form submit and navigation is not null. 
These are the steps I was using to test locally

## Testing

1. git checkout main; git pull; git checkout nmarshall/qna-isloading-check; npm run build:js;
2. Copy dist folder
3. In docs-ui repo, git checkout develop; git pull
4. In node-modules/@microsoft/atlas-js, delete current dist folder and paste new dist folder
5. npm run develop + Fiddler 
6. Navigate to a test question in answersv2 
https://dev.learn.microsoft.com/en-us/answersv2/questions/42897/testing-question-post
7. Test posting an empty comment or answer (is-loading should not display/stop)
8. Test posting not enough characters comment or answer (is-loading should not display/stop)
9. Test posting a long enough comment or answer
10. Test clicking the Follow question button   
Expected result: When posting a long enough comment or answer or clicking the Follow question button, the is-loading icon should stay until the page reloads (current behavior is it will stop and then reload the page causing confusion for users on slower connections)

### DIFF
--- a/.changeset/empty-mayflies-argue.md
+++ b/.changeset/empty-mayflies-argue.md
@@ -1,0 +1,5 @@
+---
+'@microsoft/atlas-js': minor
+---
+
+Update form submit is-loading class based on navigation
--- a/.changeset/empty-mayflies-argue.md
+++ b/.changeset/empty-mayflies-argue.md
@@ -1,5 +1,5 @@
 ---
-'@microsoft/atlas-js': minor
+'@microsoft/atlas-js': patch
 ---
 
 Update form submit is-loading class based on navigation
--- a/js/src/elements/form-behavior.ts
+++ b/js/src/elements/form-behavior.ts
@@ -182,7 +182,7 @@ export class FormBehaviorElement extends HTMLElement {
 			this.dispatchEvent(validationErrorEvent);
 			return;
 		}
-
+		let isLoading = false;
 		try {
 			this.submitting = true;
 			setBusySubmitButton(form, this.submitting);
@@ -230,6 +230,7 @@ export class FormBehaviorElement extends HTMLElement {
 			const request = new Request(beforeSubmitEvent.detail.url, beforeSubmitEvent.detail.init);
 			const response = await fetch(request);
 			if (response.ok) {
+				isLoading = true;
 				this.removeAttribute('new');
 				this.initialData = formData;
 				this.setDirty();
@@ -268,8 +269,10 @@ export class FormBehaviorElement extends HTMLElement {
 				errorAlert.focus();
 			}
 		} finally {
-			this.submitting = false;
-			setBusySubmitButton(form, this.submitting);
+			if (this.getAttribute('navigation') === null || !isLoading) {
+				this.submitting = false;
+				setBusySubmitButton(form, this.submitting);
+			}
 		}
 	}
 
--- a/js/src/elements/form-behavior.ts
+++ b/js/src/elements/form-behavior.ts
@@ -269,7 +269,7 @@ export class FormBehaviorElement extends HTMLElement {
 				errorAlert.focus();
 			}
 		} finally {
-			if (this.getAttribute('navigation') === null || !isLoading) {
+			if (!this.getAttribute('navigation') || !isLoading) {
 				this.submitting = false;
 				setBusySubmitButton(form, this.submitting);
 			}
--- a/js/src/elements/form-behavior.ts
+++ b/js/src/elements/form-behavior.ts
@@ -693,8 +693,9 @@ export function navigateAfterSubmit(href: string | null, navigationStep: Navigat
 		case 'follow':
 			if (href) {
 				location.href = href;
+				return true;
 			}
-			return true;
+			return false;
 		case 'hash-reload':
 			if (href) {
 				const search = href.includes('?') ? '' : window.location.search;
@@ -709,8 +710,9 @@ export function navigateAfterSubmit(href: string | null, navigationStep: Navigat
 		case 'replace':
 			if (href) {
 				location.replace(href);
+				return true;
 			}
-			return true;
+			return false;
 		case 'reload':
 			location.reload();
 			return true;
--- a/js/src/elements/form-behavior.ts
+++ b/js/src/elements/form-behavior.ts
@@ -126,7 +126,8 @@ export class FormBehaviorElement extends HTMLElement {
 	}
 
 	navigate(href: string | null) {
-		navigateAfterSubmit(href, this.getAttribute('navigation') as NavigationSteps);
+		const navigate = navigateAfterSubmit(href, this.getAttribute('navigation') as NavigationSteps);
+		return navigate;
 	}
 
 	scheduleCommit(event: Event) {
@@ -182,7 +183,7 @@ export class FormBehaviorElement extends HTMLElement {
 			this.dispatchEvent(validationErrorEvent);
 			return;
 		}
-		let isLoading = false;
+		let isNavigating = false;
 		try {
 			this.submitting = true;
 			setBusySubmitButton(form, this.submitting);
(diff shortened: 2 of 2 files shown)


js/src/elements/form-behavior.ts
```diff
@@ -268,8 +269,10 @@ export class FormBehaviorElement extends HTMLElement {
 				errorAlert.focus();
 			}
 		} finally {
-			this.submitting = false;
-			setBusySubmitButton(form, this.submitting);
+			if (this.getAttribute('navigation') === null || !isLoading) {
```
disimmon, line 272: mega nit 
```
!this.getAttribute('naviagtion') || !isLoading
```

disimmon: Looks good otherwise.

ncmarsh: Updated! :)

homing1:
I think this change will need a changeset (patch).

js/src/elements/form-behavior.ts
```diff
@@ -182,7 +182,7 @@ export class FormBehaviorElement extends HTMLElement {
 			this.dispatchEvent(validationErrorEvent);
 			return;
 		}
-
+		let isLoading = false;
```
jdanyow, line 185: nit: isNavigating

js/src/elements/form-behavior.ts
```diff
@@ -230,6 +230,7 @@ export class FormBehaviorElement extends HTMLElement {
 			const request = new Request(beforeSubmitEvent.detail.url, beforeSubmitEvent.detail.init);
 			const response = await fetch(request);
 			if (response.ok) {
+				isLoading = true;
```
jdanyow, line 233: let's remove this line and use the navigate call as the source of truth as to whether a navigation is in progress. We might be able to use the beforeunload event but let's not complicate things too much right now.

```diff
- this.navigate(response.headers.get('location') ?? this.getAttribute('navigation-href'));
+ isNavigating = this.navigate(response.headers.get('location') ?? this.getAttribute('navigation-href'));
```
To make this work, two other changes would be needed:
1. this.navigate return the result of the navigateAfterSubmit invocation.
2. navigateAfterSubmit needs to return true when it navigates:

```diff
export function navigateAfterSubmit(href: string | null, navigationStep: NavigationSteps) {
	switch (navigationStep) {
		case null:
			// do nothing.
-			break;
+     return false;
		case 'follow':
			if (href) {
				location.href = href;
+       return true;
			}
-			break;
+     return false;
		case 'hash-reload':
			if (href) {
				const search = href.includes('?') ? '' : window.location.search;
				if (href !== search + window.location.hash) {
					const state = (history.state || {}) as Record<string, any>;
					window.history.pushState(state, document.title, window.location.pathname + search + href); // prevents scrolling to spot until reload
				}
				location.reload();
+       return true;
			}
-			break;
+     return false;
		case 'replace':
			if (href) {
				location.replace(href);
+       return true;
			}
-			break;
+     return false;
		case 'reload':
			location.reload();
-			break;
+     return true;
		default:
			throw new Error(`Unexpected navigation attribute value.`);
	}
}
```

js/src/elements/form-behavior.ts
```diff
@@ -268,8 +269,10 @@ export class FormBehaviorElement extends HTMLElement {
 				errorAlert.focus();
 			}
 		} finally {
-			this.submitting = false;
-			setBusySubmitButton(form, this.submitting);
+			if (!this.getAttribute('navigation') || !isLoading) {
+				this.submitting = false;
+				setBusySubmitButton(form, this.submitting);
+			}
 		}
```
jdanyow, line 276: with isNavigating I think the original code could be updated to:

```diff
  } finally {
    this.submitting = false;
-   setBusySubmitButton(form, this.submitting);
+   setBusySubmitButton(form, this.submitting || isNavigating);
  }
```

jdanyow (changes requested):

js/src/elements/form-behavior.ts
jdanyow: cc @disimmon @homing1 - suggesting this because it looks like navigateAfterSubmit is the only place we can really know whether a navigation is really going to occur at the moment.

js/src/elements/form-behavior.ts
ncmarsh: Thanks for the feedback!

I tested these changes out and everything seems to be working ok except for buttons that submit but don't reload the page. Example: on Q&A v2, the answer rating buttons don't reload the page, but they do prompt UI changes. With these updates, the button keeps the is-loading icon since this use case would return isNavigating = false.

ncmarsh: If I update line 271 to `this.submitting = this.submitting === isNavigating;`, all the buttons I was testing seem to be working

jdanyow: apologies- this should have been:
```diff
  } finally {
-   this.submitting = false;
-   setBusySubmitButton(form, this.submitting);
+   this.submitting = !isNavigating;
+   setBusySubmitButton(form, this.submitting);  // I mistakenly put `|| isNavigating` here in the previous comment
  }
```

ncmarsh: Should it be `this.submitting = isNavigating`? 

With `this.submitting = !isNavigating`, on answer rating where isNavigating is false, it's not removing the is-loading class.

With `this.submitting = isNavigating`, everything seems to be working except when posting a new question, which uses "follow"

ncmarsh: I have pushed some updates based on your feedback and some testing done in qna v2 :)

disimmon:
@ncmarsh - this looks good to me. Pulled into docs-ui and tested per steps. @jdanyow / @homing1 - please let Nicole know if there are any additional steps needed.

js/src/elements/form-behavior.ts
```diff
@@ -686,12 +689,12 @@ export function navigateAfterSubmit(href: string | null, navigationStep: Navigat
...
 		case null:
 			// do nothing.
-			break;
+			return false;
 		case 'follow':
 			if (href) {
 				location.href = href;
 			}
-			break;
+			return true;
```
jdanyow, line 697: does this need to return false when there's no href?

```
if (href) {
  location.href = href;
  return true;
}
return false;
```

js/src/elements/form-behavior.ts
```diff
@@ -700,16 +703,17 @@ export function navigateAfterSubmit(href: string | null, navigationStep: Navigat
...
+				return true;
 			}
-			break;
+			return false;
 		case 'replace':
 			if (href) {
 				location.replace(href);
 			}
-			break;
+			return true;
```
jdanyow, line 713: false when there's no href?

js/src/elements/form-behavior.ts
ncmarsh: Updated! I thought this would always reload the page with or without an href, but have learned otherwise

js/src/elements/form-behavior.ts
ncmarsh: Updated! I thought this would always reload the page with or without an href, but have learned otherwise

.changeset/empty-mayflies-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-js': minor
```
homing1, line 2: I think this is more of a patch rather than minor.

disimmon (approved):

jdanyow (approved):

PR #502, opened 2023-02-27
upgrade to use node 18

wibjorn:


BenDMyers (approved):
Node 18, here we come!

disimmon (approved):

PR #508, opened 2023-03-13
Updating checkbox component and form behavior

ollips:
Tasks:
- task-799787
- task-799788

Link: preview-508

- Updated the `checkbox` component with validation colors, and updated focus styles to match other form components' styles. Also, fixed an issue when on `disabled` checkbox the `hover` styles were applied.
- Updated form behavior validation with the possibility to hide the error message that appears on top of the form. Also, updated checkbox validation so it is visually clear if there is an error on the checkbox.

## Testing

1. Visit [checkbox](https://design.learn.microsoft.com/pulls/508/components/checkbox.html) component page. Review "Validation states" section.
2. Visit [form](https://design.learn.microsoft.com/pulls/508/components/form.html) page. Check the "Custom validation" section. The old demo was updated to include a checkbox component. Hit submit to see how it gets highlighted to indicate it is a required field. Another demo was added to demonstrate how to hide error banner.

### DIFF
--- a/.changeset/itchy-chefs-talk.md
+++ b/.changeset/itchy-chefs-talk.md
@@ -2,4 +2,4 @@
 '@microsoft/atlas-js': patch
 ---
 
-Updating form behavoir validation so it validates checkboxes correctly and provides an option to hide an error banner.
+Updating form behavior validation so it validates checkboxes correctly and provides an option to hide an error banner.
--- a/.changeset/itchy-chefs-talk.md
+++ b/.changeset/itchy-chefs-talk.md
@@ -1,0 +1,5 @@
+---
+'@microsoft/atlas-js': patch
+---
+
+Updating form behavoir validation so it validates checkboxes correctly and provides an option to hide an error banner.
--- a/css/src/components/form/checkbox.scss
+++ b/css/src/components/form/checkbox.scss
@@ -150,8 +150,10 @@ $checkbox-focus-box-shadow-size: 0 0 0 0.125rem !default;
 			border-color: $checkbox-danger-border-color;
 		}
 
-		&:focus-visible + .checkbox-check {
-			box-shadow: $checkbox-focus-box-shadow-size $checkbox-danger-border-color;
+		@include focus-visible() {
+			+ .checkbox-check {
+				box-shadow: $checkbox-focus-box-shadow-size $checkbox-danger-border-color;
+			}
 		}
 	}
 
@@ -160,8 +162,10 @@ $checkbox-focus-box-shadow-size: 0 0 0 0.125rem !default;
 			border-color: $checkbox-success-border-color;
 		}
 
-		&:focus-visible + .checkbox-check {
-			box-shadow: $checkbox-focus-box-shadow-size $checkbox-success-border-color;
+		@include focus-visible() {
+			+ .checkbox-check {
+				box-shadow: $checkbox-focus-box-shadow-size $checkbox-success-border-color;
+			}
 		}
 	}
 }
--- a/css/src/components/form/checkbox.scss
+++ b/css/src/components/form/checkbox.scss
@@ -143,8 +143,6 @@ $checkbox-focus-box-shadow-size: 0 0 0 0.125rem !default;
 		}
 	}
 
-	/* stylelint-disable no-descending-specificity */
-
 	&.is-invalid input[type='checkbox']:not(:checked) {
 		+ .checkbox-check {
 			border-color: $checkbox-danger-border-color;
--- a/css/src/components/form/checkbox.scss
+++ b/css/src/components/form/checkbox.scss
@@ -68,14 +68,6 @@ $checkbox-focus-box-shadow-size: 0 0 0 0.125rem !default;
 		}
 	}
 
-	&:hover .checkbox-check {
-		border-color: $checkbox-hover-border-color;
-
-		&::before {
-			@include checkmark;
-		}
-	}
-
 	&.checkbox-sm {
 		@include control-sm;
 	}
@@ -98,10 +90,16 @@ $checkbox-focus-box-shadow-size: 0 0 0 0.125rem !default;
 			}
 		}
 
-		&[disabled] + .checkbox-check {
-			border-color: $checkbox-unchecked-border-color;
-			opacity: 0.6;
-			cursor: not-allowed;
+		&[disabled] {
+			+ .checkbox-check {
+				border-color: $checkbox-unchecked-border-color;
+				opacity: 0.6;
+			}
+
+			+ .checkbox-check,
+			~ .checkbox-text {
+				cursor: not-allowed;
+			}
 		}
 
 		&:focus-visible {
@@ -116,6 +114,14 @@ $checkbox-focus-box-shadow-size: 0 0 0 0.125rem !default;
 				outline-style: solid;
 				box-shadow: $checkbox-focus-box-shadow-size $checkbox-focus-border-color;
 			}
+		}
+	}
+
+	&:hover input[type='checkbox']:not([disabled]) + .checkbox-check {
+		border-color: $checkbox-hover-border-color;
+
+		&::before {
+			@include checkmark;
 		}
 	}
 
--- a/css/src/components/form/checkbox.scss
+++ b/css/src/components/form/checkbox.scss
@@ -15,6 +15,13 @@ $checkbox-timing-function: $input-timing-function !default;
 $checkbox-duration: $input-transition-duration !default;
 $checkbox-spacing: $spacer-3 !default;
 
+$checkbox-danger-border-color: $danger !default;
+$checkbox-success-border-color: $success !default;
+
+$checkbox-hover-border-color: $primary !default;
+$checkbox-focus-border-color: $primary !default;
+$checkbox-focus-box-shadow-size: 0 0 0 0.125rem !default;
+
 @mixin checkmark {
 	display: block;
 	position: absolute;
@@ -61,20 +68,24 @@ $checkbox-spacing: $spacer-3 !default;
 		}
 	}
 
-	&:hover .checkbox-check::before {
-		@include checkmark;
+	&:hover .checkbox-check {
+		border-color: $checkbox-hover-border-color;
+
+		&::before {
+			@include checkmark;
+		}
 	}
 
 	&.checkbox-sm {
-		font-size: $font-size-8;
+		@include control-sm;
+	}
+
+	&.checkbox-lg {
+		@include control-lg;
 	}
 
 	input[type='checkbox'] {
 		@include visually-hidden();
-
-		&:focus-visible + .checkbox-check {
-			@extend %focus;
-		}
 
 		&:checked + .checkbox-check {
 			border-color: $checkbox-color;
@@ -88,8 +99,23 @@ $checkbox-spacing: $spacer-3 !default;
 		}
 
 		&[disabled] + .checkbox-check {
+			border-color: $checkbox-unchecked-border-color;
 			opacity: 0.6;
 			cursor: not-allowed;
+		}
+
+		&:focus-visible {
+			+ .checkbox-check {
+				@extend %focus;
+			}
+
+			&:not(:checked) + .checkbox-check {
+				border-color: $checkbox-focus-border-color;
+				outline-color: transparent;
+				outline-offset: 0;
+				outline-style: solid;
+				box-shadow: $checkbox-focus-box-shadow-size $checkbox-focus-border-color;
+			}
 		}
 	}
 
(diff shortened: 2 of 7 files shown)


.changeset/slimy-monkeys-love.md
```diff
@@ -0,0 +1,7 @@
+---
+'@microsoft/atlas-css': patch
+'@microsoft/atlas-js': patch
+'@microsoft/atlas-site': patch
+---
+
+Updating checkbox component with validation colors and sizes. Updating form behavior validation so it validates checkboxes correctly and provides an option to hide an error banner. Updating webiste documentation to reflect recent changes.
```
wibjorn, line 7: These should be different changesets I believe: 
Updating checkbox component with validation colors and sizes. - minor change
Updating form behavior validation so it validates checkboxes correctly and provides an option to hide an error banner. - patch
Updating **website** documentation to reflect recent changes. - minor change

wibjorn: 1,3 could be the same

css/src/components/form/checkbox.scss
```diff
@@ -88,9 +99,24 @@ $checkbox-spacing: $spacer-3 !default;
 		}
 
 		&[disabled] + .checkbox-check {
```
wibjorn, line 101: I can't really see why, but we've restarted that iOS checkbox bug, where on iOS if we hold our finger over a disabled checkbox is looks checked permanently because fo the hover style ...

css/src/components/form/checkbox.scss
```diff
@@ -116,4 +142,26 @@ $checkbox-spacing: $spacer-3 !default;
 			border-color: $checkbox-check-color;
 		}
 	}
+
+	/* stylelint-disable no-descending-specificity */
```
wibjorn, line 146: Fix specificity?

css/src/components/form/checkbox.scss
```diff
@@ -116,4 +142,26 @@ $checkbox-spacing: $spacer-3 !default;
...
+
+	&.checkbox-danger input[type='checkbox']:not(:checked) {
```
wibjorn, line 148: `.checkbox-danger` and `.checkbox-success` naming patterns don't match standard component naming conventions and style applications. If they did then the whole checkbox (including checked state) would be green or red in these cases. We also have a convention in atlas where states (focus, loading, etc) use `is-` classes. Should these modifiers be `is-invalid` and `is-valid` instead?

css/src/components/form/checkbox.scss
```diff
@@ -116,4 +142,26 @@ $checkbox-spacing: $spacer-3 !default;
...
+			box-shadow: $checkbox-focus-box-shadow-size $checkbox-danger-border-color;
+		}
+	}
+
+	&.checkbox-success input[type='checkbox']:not(:checked) {
+		+ .checkbox-check {
+			border-color: $checkbox-success-border-color;
+		}
+
+		&:focus-visible + .checkbox-check {
```
wibjorn, line 163: for focus visible, I believe we have a mixin.

.changeset/slimy-monkeys-love.md
ollips: Updated

css/src/components/form/checkbox.scss
ollips: Oh, I think it's a separate bug. I just checked on my iphone and looks like if you're clicking on checkbox it turns into "checked" state which is expected. And if you click again, checkbox is not blue but still shows the check icon. And it stays until you click on any other checkbox. It might be connected, so I'll create a bug and link it here shortly.

css/src/components/form/checkbox.scss
ollips: Fixed

css/src/components/form/checkbox.scss
ollips: I was debating about this as well, but decided to follow the same naming convention as other form components. But you're totally right - it's not fully meeting the expectation of turning the whole checkbox into red/green when selected. So updated per your suggestion.

css/src/components/form/checkbox.scss
ollips: updated

css/src/components/form/checkbox.scss
ollips: Actually I found what caused the issue so finally pushed the update! Please take a look again

css/src/components/form/checkbox.scss
```diff
@@ -116,4 +140,40 @@ $checkbox-spacing: $spacer-3 !default;
...
+		}
+
+		@include focus-visible() {
+			+ .checkbox-check {
+				box-shadow: $checkbox-focus-box-shadow-size $checkbox-danger-border-color;
+			}
+		}
+	}
+
+	&.is-valid input[type='checkbox']:not(:checked) {
```
homing1, line 156: I'm wondering if `is-valid` is necessary for checkbox or other form fields. 

For example, when an empty required input in a form is no longer empty, we revert the `is-invalid` color and remove any related messaging rather than making the input green.

site/src/components/form.md
```diff
@@ -188,21 +189,100 @@ If you want to skip the basic validation on an input, apply a `data-skip-validat
...
+				name="sample-skip-input"
 				class="input"
 				type="text"
 				value=""
+				required
 				data-skip-validation
 			/>
 		</div>
 	</div>
+	<div class="field">
```
homing1, line 208: Perhaps a separate PR, but I wanted to point out that there will be more work required in order to validate a group of checkboxes with form behavior. Right now, it only accounts for one checkbox and the error message is using the checkbox text.

image


image

.changeset/itchy-chefs-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-js': patch
+---
+
+Updating form behavoir validation so it validates checkboxes correctly and provides an option to hide an error banner.
```
homing1, line 5: "behavior"

css/src/components/form/checkbox.scss
ollips: I added it for consistency with other form components which have both `success` and `danger` states.

site/src/components/form.md
ollips: I didn't mean to cover group selection at the moment.

.changeset/itchy-chefs-talk.md
ollips: updated

css/src/components/form/checkbox.scss
```diff
@@ -15,6 +15,13 @@ $checkbox-timing-function: $input-timing-function !default;
 $checkbox-duration: $input-transition-duration !default;
 $checkbox-spacing: $spacer-3 !default;
 
+$checkbox-danger-border-color: $danger !default;
+$checkbox-success-border-color: $success !default;
+
+$checkbox-hover-border-color: $primary !default;
+$checkbox-focus-border-color: $primary !default;
+$checkbox-focus-box-shadow-size: 0 0 0 0.125rem !default;
```
wibjorn, line 23: I'm not sure what the best guidance is here, but I find it a little strange that the keyboard focus rect is different between checked and uncheck states. Do we need this?

css/src/components/form/checkbox.scss
```diff
@@ -87,9 +90,30 @@ $checkbox-spacing: $spacer-3 !default;
...
+				border-color: $checkbox-unchecked-border-color;
+				opacity: 0.6;
+			}
+
+			+ .checkbox-check,
+			~ .checkbox-text {
+				cursor: not-allowed;
+			}
+		}
+
```
wibjorn, line 104: we missed the mixin here I think, it's used as a selector

@include focus-visible { ... rules }

css/src/components/form/checkbox.scss
```diff
@@ -116,4 +140,40 @@ $checkbox-spacing: $spacer-3 !default;
...
+		@include focus-visible() {
+			+ .checkbox-check {
+				box-shadow: $checkbox-focus-box-shadow-size $checkbox-success-border-color;
+			}
+		}
+	}
+
+	&:hover {
+		&:not(.is-valid):not(.is-invalid) {
```
wibjorn, line 169: Apologies for only just thinking of this, but what is the use case for `is-valid`? Could we remove it? Wouldn't the default state imply valid?

css/src/components/form/checkbox.scss
```diff
@@ -87,9 +90,30 @@ $checkbox-spacing: $spacer-3 !default;
...
+				cursor: not-allowed;
+			}
+		}
+
+		&:focus-visible {
+			+ .checkbox-check {
+				@extend %focus;
+			}
+
+			&:not(:checked) + .checkbox-check {
```
wibjorn, line 110: Do we need this rule for special box shadow focus? Seems like global focus styles would be more robust.

wibjorn (changes requested):

site/src/components/form.md
```diff
@@ -188,21 +189,100 @@ If you want to skip the basic validation on an input, apply a `data-skip-validat
...
+		</div>
+	</div>
+	<div class="display-flex">
+		<button type="submit" class="button button-primary button-filled">Submit form</button>
+	</div>
+	<div class="submitted-form-data-example"></div>
+</form>
+```
+
+If there is a need to hide the validation banner on top of the form apply a `data-hide-validation-banner` attribute to the `form`.
```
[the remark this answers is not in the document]
ollips: Added

site/src/components/checkbox.md
```diff
@@ -58,14 +58,41 @@ Here is an example of a group of checkboxes. Default spacing is added between ea
...
+```
+
+### Validation states
+
+`is-invalid`/`is-valid` helps with visual reflection of validation states.
+
+| State   | Class                   |                                                                                                                                                                                                                       |
+| ------- | ----------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Invalid | `.checkbox .is-invalid` | <label class="checkbox is-invalid"><input type="checkbox" name="checkbox-validation" /><span class="checkbox-check" role="presentation" aria-hidden="true"></span><span class="checkbox-text">Checkbox</span></label> |
+| Valid   | `.checkbox .is-valid`   | <label class="checkbox is-valid"><input type="checkbox" name="checkbox-validation" /><span class="checkbox-check" role="presentation" aria-hidden="true"></span><span class="checkbox-text">Checkbox</span></label>   |
```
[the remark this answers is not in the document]
ollips: Updated, `is-valid` was removed

css/src/components/form/checkbox.scss
ollips: Updated focus styles

css/src/components/form/checkbox.scss
ollips: `is-valid` was removed

css/src/components/form/checkbox.scss
ollips: updated

css/src/components/form/checkbox.scss
ollips: updated

css/src/components/form/checkbox.scss
```diff
@@ -3,6 +3,7 @@
 $checkbox-border-radius: $border-radius-sm !default;
 $checkbox-border-width: 0.0625em !default;
 $checkbox-unchecked-border-color: $control-border !default;
+$checkbox-danger-border-color: $danger !default;
```
wibjorn, line 6: should var name reference `invalid`?

site/src/components/checkbox.md
```diff
@@ -58,14 +58,40 @@ Here is an example of a group of checkboxes. Default spacing is added between ea
 </label>
 ```
 
-### Size
+### Sizes
 
-`.checkbox-sm` is used with `.checkbox` to display a smaller checkbox.
+`.checkbox-sm`/`.checkbox-lg` used in conjunction with `.checkbox` makes checkbox a little bigger or smaller than the default.
```
wibjorn, line 64: `.checkbox-sm`/`.checkbox-lg` can be used to make a checkbox bigger or smaller than the default.

site/src/components/checkbox.md
```diff
@@ -58,14 +58,40 @@ Here is an example of a group of checkboxes. Default spacing is added between ea
...
+	<input type="checkbox" name="checkbox-size" />
+	<span class="checkbox-check" role="presentation" aria-hidden="true"></span>
+	<span class="checkbox-text">Checkbox large</span>
+</label>
+```
+
+### Validation states
+
+`is-invalid` helps with visual reflection of invalid validation state.
+
```
wibjorn, line 86: We can use `.is-invalid` to reflect an invalid validation state. Note that the state should also be display to the user in another way, since by itself color is not enough.

site/src/components/form.md
```diff
@@ -188,21 +189,102 @@ If you want to skip the basic validation on an input, apply a `data-skip-validat
...
+	<div class="display-flex">
+		<button type="submit" class="button button-primary button-filled">Submit form</button>
+	</div>
+	<div class="submitted-form-data-example"></div>
+</form>
+```
+
+##### Hiding the validation banner
+
+If there is a need to hide the validation banner on top of the form apply a `data-hide-validation-banner` attribute to the `form`.
```
wibjorn, line 229: If there is a need to hide the validation banner on top of the form, we can apply the `data-hide-validation-banner` attribute to the `form`.

ollips: Updated

site/src/components/checkbox.md
ollips: Updated

site/src/components/checkbox.md
ollips: Updated

css/src/components/form/checkbox.scss
ollips: Updated

wibjorn (approved):

homing1 (approved):

PR #525, opened 2023-04-04
Banner and Notification components

ollips:
Task: task-808933

Link: preview-525

Adding Banner and Notification components to Atlas library.

## Testing

1. Review [banner](https://design.learn.microsoft.com/pulls/525/components/banner.html) and [notification](https://design.learn.microsoft.com/pulls/525/components/notification.html) components pages.
2. Provide feedback :)


site/src/components/banner.md
```diff
@@ -0,0 +1,130 @@
...
+
+Ensure the user knows they need to wait for some event in order to interact with a banner.
+
+```html
+<section class="banner is-loading">
+	<p>Loading banner.</p>
+</section>
+```
+
+## Semantic colors
```
ollips, line 117: @wibjorn Do you think we need different colors for banners? I usually see only purple ones, and leaning towards the idea of creating just one and if more would be needed - adding those in the future.

site/src/components/banner.md
```diff
@@ -0,0 +1,130 @@
...
+	data-uid="example-banner-uid"
+	data-dismissable="disappearing"
+>
+	<div class="banner-content">
+		<button
+			type="button"
+			data-bi-name="close"
+            data-dismiss
+			data-banner-dismiss="example-banner-uid"
+			class="button-reset button-close line-height-1 border-radius-rounded position-absolute top-0 right-0"
```
jdanyow, line 63: Back in docs-ui we used data-dismiss to indicate the button was a dismiss button and data-banner-dismiss to indicate the id of the banner being removed from the DOM. The banner had data-uid to indicate it's id.

Thinking we should replace data-uid with id and data-banner-dismiss with aria-owns for better a11y.

Question: is `section` the right element for a banner?

cc @BenDMyers @wibjorn

BenDMyers: Checking into the semantics; will report back

BenDMyers: **TL;DR:** `section` as it's used here is *probably* fine, but we *probably* want to consider encouraging the use of `aria-label` or `aria-labelledby`, *especially* for titled banners.

***

I wanted to see what accessibility practitioners were saying and doing about notification banners like these, so I checked on other design systems and component libraries that I knew were design-conscious, and I checked for prior conversations in an accessibility Slack I'm in.

## Slack Says Probably Not a Big Deal

The Slack conversations I saw didn't seem to have strong opinions on which roles were best for this, or whether a non-`generic` landmark role was even necessary for this. The mood seemed to be that a `div` (or an unlabeled `section` like this, which behaves identically to a `div`) could probably be fine, or even desirable in cases where we'd want to avoid overwhelming users with a clutter of landmarks. In some contexts, folks suggested using an `aside`, and this was also met with a "that's probably fine" shrug.

## GOV.UK Opts For `role=region`

The GOV.UK Design System, which is generally considered accessibility-conscious, uses `role=region` for [their Notification Banner component](https://design-system.service.gov.uk/components/notification-banner/). Checking through their GitHub issues, it seems like they've brought this through user testing with screenreader, and it tested positively, as users were able to identify the content and backtrack to it easily if they needed. GOV.UK is also using this pattern for cookie consent banners.

`role=region` (or a `section` that we want to inherit the `region` role) needs to have an accessible name defined by either `aria-label` or `aria-labelledby`. If we were to recommend the `role=region` approach, we'd probably also need to include a suggested `aria-label` (maybe a localized "Notice"?) or suggestions for how to use `aria-labelledby` (pointing to the title if one is provided) in this case.

## My Recommendation

This is *probably* fine to leave as is - an unlabeled `section` that'll take a `generic` role much like a `div`.

If we determined we needed stronger semantics, maintaining the `section` would be best, and the way to go would be to provide guidance around `aria-label`/`aria-labelledby`, using GOV.UK Design System as prior art.

## Elements and Roles That Don't Make the Cut

<details>
<summary>header and role=banner, weirdly enough</summary>
The header element supplies a "banner" role, which sounds perfect! However, the "banner" role is meant for <a href="https://w3c.github.io/aria/#banner">"A landmark that contains mostly site-oriented content, rather than page-specific content."</a> There's also not meant to be more than one "banner" role on a page.
</details>
<details>
<summary>role=alert</summary>
This creates a live region on the page, which means the screenreader may likely immediately start yammering these contents once the page loads. Also, the "alert" role isn't meant to have interactive controls like close buttons in them.
</details>

site/src/components/banner.md
```diff
@@ -0,0 +1,130 @@
...
+template: standard
+figmaEmbed: https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FuVA2amRR71yJZ0GS6RI6zL%2F%25F0%259F%258C%259E-Atlas-Design-Library%3Fnode-id%3D1002%253A7106%26t%3DWU64bU98uaT1nU7A-1
+classType: Component
+classPrefixes:
+  - banner
+---
+
+# Banner
+
+Banner component allows adding a full width notification block to the page.
```
BenDMyers, line 13: Can we provide a paragraph here that briefly explains the difference between the Banner component and the Notification component, and then links out to the Notification component page?

These two components seem pretty similar to me, and I think it'd be good to make the difference between them evident up front.

site/src/components/notification.md
```diff
@@ -0,0 +1,129 @@
...
+template: standard
+figmaEmbed: https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FuVA2amRR71yJZ0GS6RI6zL%2F%25F0%259F%258C%259E-Atlas-Design-Library%3Fnode-id%3D1002%253A7106%26t%3DWU64bU98uaT1nU7A-1
+classType: Component
+classPrefixes:
+  - notification
+---
+
+# Notification
+
+Notification is a colored block meant to alert users of something.
```
BenDMyers, line 13: Similar feedback to what I said on the Banner page: can we provide a paragraph that explains the difference between Notifications and Banners, and links out to the Banner component page?

BenDMyers: Likewise, can we provide some sample intended use cases for Notifications?

css/src/components/banner.scss
```diff
@@ -0,0 +1,100 @@
...
+	display: block;
+	position: relative;
+	outline-color: $text;
+	background-color: $banner-background;
+	color: $banner-color;
+	font-size: $banner-font-size;
+	padding-block: $banner-padding;
+	word-wrap: break-word;
+	word-break: break-word;
+	border-block: $banner-border-width solid $banner-border-color;
```
BenDMyers, line 22: In the high contrast theme as well as in an operating system forced-colors contrast theme, I think the lack of left/right borders kind of stands out.

**High contrast theme:**
Banner in Usage section viewed in the high contrast theme. The banner has a black background, like the rest of the page. It has white borders on the top and bottom.

**Forced colors:**
Banner in Usage section viewed in the Dusk forced colors theme. The banner has a gray background, like the rest of the page. It has white borders on the top and bottom.

I think it might be nice to add left/right borders for the high contrast theme, as well as in forced colors mode. Thoughts?

css/src/components/banner.scss
```diff
@@ -0,0 +1,100 @@
...
+		color: currentColor;
+		font-weight: $weight-semibold;
+	}
+
+	.banner-title {
+		display: flex;
+		align-items: center;
+		justify-content: flex-start;
+
+		.icon {
```
BenDMyers, line 85: This icon doesn't seem to be theme-aware: here's what the sun icon shown in the docs looks like for me in dark mode:

A titled banner viewed in dark mode. The title has a sun icon to its left which is an unreadable black, whereas the surrounding text is a readable white.

Should we set icon colors in these styles, or is this meant to be left for the user to handle?

BenDMyers: That would work for me, @wibjorn

css/src/components/banner.scss
BenDMyers: On further thinking, this could just be an artifact of the way I'm viewing this not at fullscreen. A full-width banner wouldn't really need to show left/right borders.

That in mind, feel free to resolve this if you don't think we need this.

css/src/components/banner.scss
```diff
@@ -0,0 +1,100 @@
...
+				}
+			}
+
+			/* stylelint-enable max-nesting-depth */
+		}
+	}
+
+	.banner-title,
+	a:not(.button) {
+		color: currentColor;
```
BenDMyers, line 76: Could we tweak this rule so that in the high contrast theme, links inside banners can stay bright yellow?

Banner in high contrast theme with a black background, white text, and a bold white link that reads: Link Text

Feel free to resolve if we don't want/need that.

wibjorn:
Now that the accessibility automated testing tool has been added, please add the page to accessibility.spec.ts in integration

css/src/core/animations.scss
```diff
@@ -24,3 +24,18 @@
 		transform: scale(1);
 	}
 }
+
+// stylelint-disable-next-line keyframes-name-pattern
```
wibjorn, line 28: do we really need to disable lint here? seems like we could run rename the animation, same for below

css/src/tokens/themes.scss
```diff
@@ -21,6 +21,7 @@ $themes: (
 		'card-background': $palette-white,
 		'border': $palette-grey-30,
 		'border-accent': $palette-yellow-sand,
+		'border-dark': $palette-grey-100,
```
wibjorn, line 24: We already have border and control border. Do we really want another?

js/src/utilities/notify-content-updated.ts
```diff
@@ -0,0 +1,6 @@
+/**
```
wibjorn, line 1: Do we need this?

site/src/atomics/typography.md
```diff
@@ -26,6 +27,7 @@ The typography scale is designed for great readability across the platform. This
 | `letter-spacing`  | `wide`                                                           | N\A        |
 | `text-transform`  | `uppercase`                                                      | N\A        |
 | `text-align`      | `center`, `right`                                                | `tablet`   |
+| `line-height`     | `1`, `normal`                                                    | N\A        |
```
wibjorn, line 30: curious what the use case for `line-height-1` is. seems like we could do without it

css/src/components/banner.scss
```diff
@@ -0,0 +1,100 @@
...
+		font-weight: $weight-semibold;
+	}
+
+	.banner-title {
```
wibjorn, line 80: Can banner title be removed entirely? I thought this would mostly be an inline element

wibjorn (approved):

site/src/components/notification.md
```diff
@@ -0,0 +1,129 @@
...
+			>
+				<path
+					d="M8 0C8.73438 0 9.44271 0.09375 10.125 0.28125C10.8073 0.46875 11.4427 0.739583 12.0312 1.09375C12.6198 1.44792 13.1589 1.86458 13.6484 2.34375C14.138 2.82292 14.5573 3.36198 14.9062 3.96094C15.2552 4.5599 15.5234 5.19792 15.7109 5.875C15.8984 6.55208 15.9948 7.26042 16 8C16 8.73438 15.9062 9.44271 15.7188 10.125C15.5312 10.8073 15.2604 11.4427 14.9062 12.0312C14.5521 12.6198 14.1354 13.1589 13.6562 13.6484C13.1771 14.138 12.638 14.5573 12.0391 14.9062C11.4401 15.2552 10.8021 15.5234 10.125 15.7109C9.44792 15.8984 8.73958 15.9948 8 16C7.26562 16 6.55729 15.9062 5.875 15.7188C5.19271 15.5312 4.55729 15.2604 3.96875 14.9062C3.38021 14.5521 2.84115 14.1354 2.35156 13.6562C1.86198 13.1771 1.44271 12.638 1.09375 12.0391C0.744792 11.4401 0.476562 10.8021 0.289062 10.125C0.101562 9.44792 0.00520833 8.73958 0 8C0 7.26562 0.09375 6.55729 0.28125 5.875C0.46875 5.19271 0.739583 4.55729 1.09375 3.96875C1.44792 3.38021 1.86458 2.84115 2.34375 2.35156C2.82292 1.86198 3.36198 1.44271 3.96094 1.09375C4.5599 0.744792 5.19792 0.476562 5.875 0.289062C6.55208 0.101562 7.26042 0.00520833 8 0ZM9 12V10H7V12H9ZM9 9V4H7V9H9Z"
+					fill="currentColor"
```
wibjorn, line 42: should icon have styling for icon too?

wibjorn (changes requested):

css/src/components/notification.scss
```diff
@@ -0,0 +1,101 @@
...
+		}
+	}
+
+	.notification-title,
+	a:not(.button) {
+		color: currentColor;
+		font-weight: $weight-semibold;
+	}
+
+	.notification-title {
```
wibjorn, line 81: title seems too close to following text for me. Margin block end? image

ollips: Updated, does it look better?

image

css/src/components/banner.scss
ollips: Removed!

js/src/utilities/notify-content-updated.ts
ollips: I brought it here to safely replace dismiss in docs-ui. `notifyContentUpdated` called in dismissing code.

css/src/tokens/themes.scss
ollips: Good idea, updated!

css/src/core/animations.scss
ollips: Fixed 👍🏻

css/src/components/banner.scss
```diff
@@ -0,0 +1,100 @@
...
+		color: transparent !important;
+
+		> :first-child {
+			margin-inline-start: calc($banner-padding + 0.375em);
+		}
+
+		&::before {
+			@include loader;
+
+			position: absolute !important;
```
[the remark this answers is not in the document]
ollips: Oh, those `!important`'s were left there by accident, removed!

site/src/components/banner.md
```diff
@@ -0,0 +1,130 @@
...
+	</p>
+	<p>A banner about something happening in the world <a href="#">Link Text</a></p>
+</section>
+```
+
+### Dismissing banner
+
+Banners can be dismissed by adding the `data-dismissable` attribute to the `.banner` and `data-dismiss`to the closing button.
+
+If `data-dismissable` attribute is also assigned a `dissapearing` value - banner would slowly slide up.
```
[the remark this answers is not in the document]
ollips: Updated

site/src/components/banner.md
```diff
@@ -0,0 +1,130 @@
...
+
+You can add a title to the banner with an optional icon.
+
+```html-no-indent
+<section class="banner">
+	<p class="banner-title">
+		<span class="icon" aria-hidden="true">
+          <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 14 14"><path d="M7 10.3c-.6 0-1.2-.2-1.7-.5-.5-.2-.9-.6-1.2-1.1S3.7 7.6 3.7 7s.2-1.2.5-1.7.7-.9 1.2-1.2c1-.6 2.3-.6 3.4 0 .5.3.9.7 1.2 1.2s.5 1.1.5 1.7-.2 1.2-.5 1.7c-.4.5-.8.9-1.3 1.2-.5.3-1.1.4-1.7.4zm0-5.6c-.4 0-.8.1-1.1.3-.4.3-.6.5-.9.9-.2.3-.3.7-.3 1.1s.1.8.3 1.1c.3.4.5.6.9.9.7.4 1.6.4 2.2 0 .4-.3.6-.5.9-.9.2-.3.3-.7.3-1.1s-.1-.8-.3-1.1c-.3-.4-.5-.6-.9-.9-.3-.2-.7-.3-1.1-.3zM6.4 0h1.1v2.8H6.4zm0 11.2h1.1V14H6.4zm4.8-4.8H14v1.1h-2.8zm-1.561 3.606l.778-.778 1.98 1.98-.779.778zm.075-6.065l1.98-1.98.778.778-1.98 1.98zM0 6.4h2.8v1.1H0zm1.461-3.567l.778-.777 1.98 1.98-.778.777zm.067 8.429l1.98-1.981.777.777-1.978 1.98z"/></svg>
+        </span>
+		<span>Title</span>
```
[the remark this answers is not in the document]
ollips: The "banner with title" example was completely removed after realizing they normally won't have them.

css/src/components/banner.scss
ollips: Updated examples with `fill-current-color` atomic.

css/src/components/banner.scss
ollips: That was my thinking - since it's a full-width component that takes the whole screen left/right borders felt redundant to me.

site/src/components/banner.md
ollips: They are similar but for sure different ones. I've updated the description a bit on the banners page. let me know if you have any suggestions. Words are not my strongest side :D And I don't think these two need to be linked to each other, maybe only if we add a sentence like that - "For alerting users of something within the content we'd recommend using the notification component (link)".

site/src/components/banner.md
ollips: Updated examples to use `div`'s with `id`'s and `aria-owns` on the closing button.

Also, after doing some investigation and giving it a thought, the example with the title was removed - it turned out they normally don't have standing-out titles.

css/src/components/banner.scss
```diff
@@ -0,0 +1,53 @@
...
+.banner {
+	@extend %layout-gap;
+
+	display: block;
+	position: relative;
+	outline-color: $text;
+	background-color: $banner-background;
+	color: $banner-color;
+	font-size: $banner-font-size;
+	padding-block: $banner-padding;
```
homing1, line 19: Wondering if banners should also have some default inline padding.

wibjorn: The layout-gap mixin is doing that I believe

css/src/components/banner.scss
```diff
@@ -0,0 +1,53 @@
...
+	word-wrap: break-word;
+	word-break: break-word;
+	border-block: $banner-border-width solid $banner-border-color;
+
+	&.banner-dismiss {
```
wibjorn, line 24: Didn't we refactor so dismiss would be separate? Should it also contain separate styling? How does this jive with `button-dismiss`?

css/src/components/banner.scss
wibjorn: Might be worth one of the dreaded theme-selector rules

css/src/components/banner.scss
```diff
@@ -0,0 +1,53 @@
...
+			border-color: transparent transparent $banner-color $banner-color;
+		}
+	}
+
+	a:not(.button) {
+		color: currentColor;
+		font-weight: $weight-semibold;
+	}
+
+	&.disappearing {
```
wibjorn, line 50: Does this belong here? Seems like it should probably be refactored elsewhere

css/src/components/notification.scss
```diff
@@ -0,0 +1,85 @@
...
+		margin-block-end: $layout-1;
+
+		.icon {
+			flex-shrink: 0;
+			align-self: start;
+			margin-inline-end: 0.375em;
+		}
+	}
+
+	&.disappearing {
```
wibjorn, line 82: Duplicative? Move out?

js/src/elements/dismiss.ts
```diff
@@ -0,0 +1,34 @@
...
+
+			dismissElement(dismissTarget);
+		},
+		{ passive: true }
+	);
+}
+
+async function dismissElement(dismissTarget: Element) {
+	if (dismissTarget.getAttribute('data-dismissable') === 'disappearing') {
+		dismissTarget.classList.add('disappearing');
```
wibjorn, line 27: A third confirmation that I think we need to think about dismiss a bit and how we're going to use it as a selector.

We've got a few now:
[data-dismissable]
.disappearing
.notification-dismiss
.banner-dismiss

js/src/utilities/notify-content-updated.ts
wibjorn: I'm not sure this is something Atlas should do, two things:
- do we want to introduce a possible naming conflict with `notifyContentUpdated`?
- do we want to prefix `content-update` custom event name with something?

css/src/components/banner.scss
ollips: Yes, it is!

js/src/elements/dismiss.ts
ollips: After thinking about it for a while I don't think it's necessary to provide two options of how the element would be removed - slowly or immediately. I'm going to update so the default would "fade out" and if any component needs a different treatment - it can be adjusted within that's component css.

`.notification-dismiss` and `.banner-dismiss` classes were created to help with positioning the closing button correctly. I can update the selector to be `.banner[data-dismissable]` instead of `.banner.banner-dismiss`

css/src/components/notification.scss
ollips: Notification and banner have different animations so not really a duplicate.

css/src/components/banner.scss
ollips: I feel like it does. I'm setting up the animation styles only.

site/src/components/notification.md
ollips: Updated with `fill-current-color` class. Lmk if you meant something else

site/src/components/banner.md
```diff
@@ -0,0 +1,91 @@
...
+template: standard
+figmaEmbed: https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FuVA2amRR71yJZ0GS6RI6zL%2F%25F0%259F%258C%259E-Atlas-Design-Library%3Fnode-id%3D1002%253A7106%26t%3DWU64bU98uaT1nU7A-1
+classType: Component
+classPrefixes:
+  - banner
+---
+
+# Banner
+
+Banner component allows adding a full width notification block to the page to notify user about something that’s not directly related to the page content.
```
wibjorn, line 13: To decrease confusion, I think we should probably just discuss the purpose without using the word notification

site/src/components/notification.md
wibjorn: And maybe we don't use the word `.alert`?

site/src/components/banner.md
wibjorn: @jdanyow @BenDMyers @ollips I had to go and read up about `aria-owns`, and after doing so don't see why we should use it instead of maybe something like `aria-controls`.

> The only reason to include aria-owns is to expose a parent/child contextual relationship to assistive technology when the DOM's construction can't provide that relationship.

Quote from [aria-owns on mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-owns).

So is a dismissible button the parent of the item it dismisses? I would say no. Compare to `aria-controls`:

> The global aria-controls property identifies the element (or elements) whose contents or presence are controlled by the element on which this attribute is set.

site/src/components/banner.md
```diff
@@ -0,0 +1,55 @@
...
+</div>
+```
+
+### Dismissing banner
+
+Banners can be dismissed by adding the `data-dismissable` attribute to the `.banner` and `data-dismiss`to the closing button.
+
+```html-no-indent
+<div class="banner" id="example-banner-id" data-dismissable>
+	<p>Dismissable banner. It would slowly slide up after closing.</p>
```
wibjorn, line 33:
```suggestion
	<p>Dismissable banner. It will slowly slide up after closing.</p>
```

site/src/components/notification.md
```diff
@@ -0,0 +1,89 @@
...
+<div class="notification is-loading">
+	<p>Loading notification.</p>
+</div>
+```
+
+## Semantic colors
+
+| State               | Class                                 | Example                                                                                                                          |
+| ------------------- | ------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
+| Secondary (default) | `.notification`                       | <div class="notification"><p class="margin-top-none">Default notification <a href="#">Link</a></p></div>                         |
```
wibjorn, line 79: Maybe just say Default?

css/src/components/banner.scss
ollips: Updated

css/src/components/banner.scss
ollips: Updated. I moved `.dissapearing` styles to a separate file - `dismiss.scss`, but I still have it in `banner` because banner animates differently.

css/src/components/notification.scss
ollips: Moved out to `dismiss` file

site/src/components/notification.md
ollips: Ooops, updated

site/src/components/banner.md
ollips: Updated

site/src/components/banner.md
ollips: Updated

site/src/components/notification.md
ollips: Updated

css/src/core/dismiss.scss
```diff
@@ -0,0 +1,9 @@
+@mixin dismiss {
+	display: flex !important;
+	align-items: flex-start;
+	justify-content: space-between;
+}
+
+.disappearing {
```
wibjorn, line 7: atomics/animations.scss instead?

css/src/core/dismiss.scss
```diff
@@ -0,0 +1,9 @@
+@mixin dismiss {
```
wibjorn, line 1: should this be in /mixins? Due to import order, I think we want to ensure it's available early in the order.

css/src/core/animations.scss
```diff
@@ -24,3 +24,16 @@
 		transform: scale(1);
 	}
 }
+
+@keyframes slide-up-out-fade-out {
```
wibjorn, line 28: maybe disappearing should be up here?

css/src/core/dismiss.scss
ollips: Updated! Forget to move it after realized creating mixin is a good idea :D

css/src/core/dismiss.scss
ollips: Updated

css/src/components/banner.scss
ollips: Updated! :)

css/src/components/notification.scss
```diff
@@ -0,0 +1,101 @@
...
+				}
+			}
+
+			/* stylelint-enable max-nesting-depth */
+		}
+	}
+
+	.notification-title,
+	a:not(.button) {
+		color: currentColor;
```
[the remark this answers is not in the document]
ollips: Updated

js/src/utilities/notify-content-updated.ts
ollips: Refactored this section, removed completely `notifyContentUpdated`, and added `dismiss-content-update` custom even directly in dismiss code.

css/src/components/button-close.scss
```diff
@@ -0,0 +1,11 @@
+.button-close {
```
wibjorn, line 1: Should we rename this to something that doesn't sound like it's a subcomponent of `.button`?

site/src/components/banner.md
wibjorn: Resolving this

css/src/components/button-close.scss
ollips: Renamed to `.dismiss`

wibjorn (changes requested):

site/src/components/notification.md
```diff
@@ -0,0 +1,69 @@
...
+template: standard
+figmaEmbed: https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FuVA2amRR71yJZ0GS6RI6zL%2F%25F0%259F%258C%259E-Atlas-Design-Library%3Fnode-id%3D1002%253A7106%26t%3DWU64bU98uaT1nU7A-1
+classType: Component
+classPrefixes:
+  - notification
+---
+
+# Notification
+
+Notification is a colored block meant to inform users of something.
```
wibjorn, line 13:
```suggestion
Notification is a colored block meant to inform users of something. It is intended for use within another element, such as the body of an article.
```

site/src/components/banner.md
```diff
@@ -0,0 +1,35 @@
...
+template: standard
+figmaEmbed: https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FuVA2amRR71yJZ0GS6RI6zL%2F%25F0%259F%258C%259E-Atlas-Design-Library%3Fnode-id%3D35687%253A253899%26t%3DMgxlZCy8on3CKZ9O-1
+classType: Component
+classPrefixes:
+  - banner
+---
+
+# Banner
+
+Banner component allows adding a full width message block to the page to notify user about something that’s not directly related to the page content.
```
wibjorn, line 13:
```suggestion
Banner component allows adding a full width message block to notify user about something that’s not directly related to the page content.
```

ollips: Updated!

site/src/components/notification.md
ollips: Updated!

homing1 (approved):

wibjorn (approved):

BenDMyers (approved):

PR #528, opened 2023-04-10
add target-text color and url

ellynliu:
Task: [task-794609](https://ceapex.visualstudio.com/Engineering/_sprints/backlog/Docs/Engineering/2023-04-05?workitem=794609)

Link: [preview-528](https://design.learn.microsoft.com/pulls/528)

Change the color of the highlighted target text to meet accessibility requirements. Also, if a link is highlighted, make it clear to the reader that there is a link available (underline, on hover). 

## Testing

1. Highlight some text by adding a parameter to the URL:.
- For example: https://design.learn.microsoft.com/pulls/528#:~:text=Ensure%20git%20is%20installed
2. Ensure that text and links are visible in different color palettes and before/after clicking on a link.

### DIFF
--- a/.changeset/tricky-months-tie.md
+++ b/.changeset/tricky-months-tie.md
@@ -1,0 +1,5 @@
+---
+"@microsoft/atlas-css": patch
+---
+
+For browsers that support target-text, update the text color to make text and URLs more visible when highlighted
--- a/.changeset/tricky-months-tie.md
+++ b/.changeset/tricky-months-tie.md
@@ -2,4 +2,4 @@
 "@microsoft/atlas-css": minor
 ---
 
-For browsers that support target-text, update the text color to make text and URLs more visible when highlighted
+For browsers that support target-text, update the background color to make text and URLs more visible when highlighted
--- a/.changeset/tricky-months-tie.md
+++ b/.changeset/tricky-months-tie.md
@@ -1,5 +1,5 @@
 ---
-"@microsoft/atlas-css": patch
+"@microsoft/atlas-css": minor
 ---
 
 For browsers that support target-text, update the text color to make text and URLs more visible when highlighted
--- a/css/src/core/bare-elements.scss
+++ b/css/src/core/bare-elements.scss
@@ -38,6 +38,10 @@ fieldset {
 	color: $info-dark;
 }
 
-a > ::target-text {
-	color: $info-dark;
+:where(a)::target-text {
+	color: $hyperlink;
+}
+
+:where(a:visited)::target-text {
+	color: $visited;
 }
--- a/css/src/core/bare-elements.scss
+++ b/css/src/core/bare-elements.scss
@@ -32,3 +32,12 @@ fieldset {
 	min-width: 0; // as a replaced element min-width:0 is required to enable it to work in responsive layouts
 	border: none;
 }
+
+::target-text {
+	background-color: $info-background;
+	color: $info-dark;
+}
+
+a > ::target-text {
+	color: $info-dark;
+}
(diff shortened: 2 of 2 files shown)


homing1:
Looks like the PR is missing a changeset.

css/src/core/bare-elements.scss
```diff
@@ -32,3 +32,12 @@ fieldset {
...
 }
+
+::target-text {
+	background-color: $info-background;
+	color: $info-dark;
+}
+
+a > ::target-text {
+	color: $info-dark;
+}
```
BenDMyers, line 43: I'm a little nervous about making highlighted links match the overall text color, especially since we also remove underlines. Can you help me understand the direction for this?

ollips: @ellynliu you can actually use our standard link color, I just checked and it passes accessibility color contrast requirements with purple background:

```
color: $hyperlink;
```

css/src/core/bare-elements.scss
```diff
@@ -32,3 +32,12 @@ fieldset {
 	min-width: 0; // as a replaced element min-width:0 is required to enable it to work in responsive layouts
 	border: none;
 }
+
+::target-text {
```
homing1, line 36: Looks like it's not supported by Firefox and Safari. Is there a fallback for non-supported browsers? 
image

css/src/core/bare-elements.scss
homing1: We'll need another color for visited links.

ellynliu: Added $hyperlink and $visited as colors for links

.changeset/tricky-months-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+"@microsoft/atlas-css": patch
```
wibjorn, line 2: This is a minor update, because we're adding functionality.

css/src/core/bare-elements.scss
```diff
@@ -32,3 +32,16 @@ fieldset {
...
+	background-color: $info-background;
```
wibjorn, line 37: Wondering if `$code-highlight-background` might work as a better color to us, since it's already used for this purpose in code, and might not need to care about `color`. It might also ensure high contrast works a bit better.

```css
::target-text {
	background-color: $code-highlight-background;
}
```

image


Don't feel extremely strongly about this, but it might be worth experiementing.

.changeset/tricky-months-tie.md
ellynliu: Fixed!

css/src/core/bare-elements.scss
```diff
@@ -32,3 +32,15 @@ fieldset {
 	min-width: 0; // as a replaced element min-width:0 is required to enable it to work in responsive layouts
 	border: none;
 }
+
+::target-text {
+	background-color: $code-highlight-background;
+}
+
+:where(a)::target-text {
```
wibjorn, line 40: are latter two needed now with the new bg color?

ellynliu: Just tried it out and removed those sections

.changeset/tricky-months-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+"@microsoft/atlas-css": minor
+---
+
+For browsers that support target-text, update the text color to make text and URLs more visible when highlighted
```
wibjorn, line 5: Sorry one last little nit here, ` update the text color` should probably say ` update the background color`

wibjorn: (you can just edit this file directly instead of using the changesets cli)

BenDMyers (approved):

wibjorn (approved):

PR #541, opened 2023-05-17
[wip] Dismiss component

ollips:
Task: task-[work-item-number]

## Testing

1. Step one
2. Step two
   Expected result:

## Additional information

[Optional]

## Contributor checklist


css/src/components/dismiss.scss
```diff
@@ -0,0 +1,21 @@
...
+	font-size: 1em;
+	cursor: pointer;
+}
+
+.is-dismissed {
+	animation: fade-out 0.4s ease-in forwards;
+
+	&.animation-slide-up {
+		animation: slide-up-fade-out 0.4s ease-in forwards;
+	}
```
BenDMyers, line 20: Could/should we put this in a `prefers-reduced-motion: no-preference` media query? This disappearing animation isn't critical, so I think folks with vestibular disorders would benefit from not needing to see the animation.

css/src/components/banner.scss
```diff
@@ -22,6 +22,12 @@ $banner-border-width: $border-width !default;
 	border-block: $banner-border-width solid $banner-border-color;
 	line-height: $line-height-normal;
 
+	&[data-dismissable] {
```
wibjorn, line 25: classes for styling, and `data-` for behavior?

js/src/behaviors/dismiss.ts
```diff
@@ -0,0 +1,40 @@
...
+			}
+
+			dismissElement(dismissTarget);
+		},
+		{ passive: true }
+	);
+}
+
+async function dismissElement(dismissTarget: Element) {
+	if (dismissTarget.hasAttribute('disappearing')) {
```
BenDMyers, line 24: This should be `data-disappearing`, I think - otherwise browsers will give validation errors about this not being an existent HTML attribute name.

site/src/components/banner.md
```diff
@@ -33,3 +33,23 @@ Ensure the user knows they need to wait for some event in order to interact with
 	<p>Loading banner.</p>
 </div>
 ```
+
+### Dismissing banner
+
+Banner can be dismissed in conjunction usage with `dismiss` component. Read more about it [here](./dismiss.md).
```
BenDMyers, line 39: I'd like to avoid links with generic text like "here." Perhaps something like this could work?

```suggestion
Banners can be made dismissable with [the `dismiss` component](./dismiss.md).
```

site/src/components/banner.md
```diff
@@ -33,3 +33,23 @@ Ensure the user knows they need to wait for some event in order to interact with
...
+
+```html
+<div class="banner" id="example-banner-id" data-dismissable disappearing="slide-up">
```
BenDMyers, line 42: See comment above - I don't think we should use `disappearing` as an attribute name. This should be made into a data attribute.

css/src/components/banner.scss
```diff
@@ -22,6 +22,12 @@ $banner-border-width: $border-width !default;
...
+		display: flex !important;
```
wibjorn, line 26: Could we avoid `!important` here by reordering? It would be nice to keep `!important` reserved for atomics.

js/src/behaviors/dismiss.ts
```diff
@@ -0,0 +1,40 @@
...
+		dismissTarget.classList.add('is-dismissed');
+
+		const disappearing = dismissTarget.getAttribute('disappearing');
+		if (disappearing === 'slide-up') {
```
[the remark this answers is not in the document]
homing1: What about `animation`?

site/src/components/dismiss.md
```diff
@@ -0,0 +1,97 @@
...
+
+There are a few things to keep in mind when using dismiss component:
+
+- dismiss component has to have the `data-dismiss` attribute
+- the component it is used within has to have `data-dismissable` attribute
+
+Below is an example usage with `notification` component.
+
+```html
+<div class="notification" id="example-notification-01" data-dismissable data-testid="notification">
```
wibjorn, line 37: `data-test-id`?

css/src/components/banner.scss
wibjorn: perhaps this is a case when the `has` selector would work well? `*:has(.dismiss) {}`?

site/src/components/dismiss.md
```diff
@@ -0,0 +1,97 @@
...
+	</span>
+</button>
+```
+
+## Usage
+
+There are a few things to keep in mind when using dismiss component:
+
+- dismiss component has to have the `data-dismiss` attribute
+- the component it is used within has to have `data-dismissable` attribute
```
BenDMyers, line 32: Please feel to disregard this! But here's my take on this section.

```suggestion
Dismissal functionality requires two parts:

1. The dismiss button, which has the `data-dismiss` attribute
2. The parent element/container to dismiss, which has the `data-dismissable` attribute
```

css/src/components/banner.scss
BenDMyers: @wibjorn We're still [waiting on Firefox support for `:has`](https://developer.mozilla.org/en-US/docs/Web/CSS/:has#browser_compatibility), I think.

site/src/components/dismiss.md
BenDMyers: Do we even want to have `data-test-id` attributes at all in our Atlas docs? (I vote we remove it)

site/src/components/dismiss.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+### Animating disappearance
+
+There are two options available to animate component's disappearance. Adding `disappearing` attribute to the dismissable component will make it slowly fade out.
+
+Assigning `slide-up` value to the `disappearing` attribute will make component fading while sliding up.
+
+```html
+<div class="notification" id="example-notification-02" data-dismissable disappearing>
```
wibjorn, line 65: Could we unify the interface for animation here a bit? I think that maybe something like:

Couple very quick proposals:

```
on-dismiss-animation="fade" and
on-dismiss-animation="slide-up"
```

```
no property for default fade
on-dismiss="slide-up"
```

js/src/behaviors/dismiss.ts
```diff
@@ -0,0 +1,40 @@
...
+		},
+		{ passive: true }
+	);
+}
+
+async function dismissElement(dismissTarget: Element) {
+	if (dismissTarget.hasAttribute('disappearing')) {
+		dismissTarget.classList.add('is-dismissed');
+
+		const disappearing = dismissTarget.getAttribute('disappearing');
```
wibjorn, line 27: Something to noodle over, but do we really need/want these animation in our CSS? If the answer is no or maybe not, we might also be able to leverage the JS only API [element.animate](https://developer.mozilla.org/en-US/docs/Web/API/Element/animate), which would allow us to do the same. I don't feel strongly about this, feel free to resolve

css/src/components/dismiss.scss
ollips: Thanks for the suggestion, to eliminate extra CSS lines to target media query made some updates directly in `dismiss.ts`!

css/src/components/banner.scss
ollips: I don't have a better idea of how to target it, since as Ben mentioned, `:has` isn't supported in FF. I was trying to come up with some universal solution but it would really depend on the component's markup dismiss used within.

js/src/behaviors/dismiss.ts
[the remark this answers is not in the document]
ollips: Renamed to `data-dismiss-animation`

css/src/components/dismiss.scss
```diff
@@ -0,0 +1,21 @@
...
+	@include transparent-effects;
+
+	display: inline-block;
+	width: 1em;
+	height: 1em;
+	padding: 0;
+	border: 0;
+	border-radius: $border-radius-rounded;
+	color: currentColor;
+	font-size: 1em;
```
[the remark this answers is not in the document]
ollips: Good catch, updated!

js/src/behaviors/dismiss.ts
ollips: Renamed attribute to `data-dismiss-animation`

site/src/components/banner.md
ollips: Updated, thanks for recommendation :)

site/src/components/banner.md
ollips: Updated

css/src/components/banner.scss
ollips: Ooops it shouldn't be there in the first place, removed!

site/src/components/dismiss.md
ollips: Updated combining both versions 😁

site/src/components/dismiss.md
ollips: You read my mind, just earlier today I thought I probably should've not use any default option. Updated and renamed attribute to `data-dismiss-animation`.

site/src/components/dismiss.md
ollips: `data-testid` is used in the integration test.

Also, it is `data-testid` instead of `data-test-id` cause I read in the playwright documentation that's the default attribute name for targeting elements by id:

> By default, the data-testid attribute is used as a test id

https://playwright.dev/docs/api/class-locator#locator-get-by-test-id

js/src/behaviors/dismiss.ts
```diff
@@ -0,0 +1,46 @@
...
+
+			const dismissTarget = dismissButton.closest('[data-dismissable]');
+
+			if (!dismissTarget) {
+				return;
+			}
+
+			dismissElement(dismissTarget);
+		},
+		{ passive: true }
```
BenDMyers, line 19: I think `passive` is mostly meant for scroll events or similar. I'm not sure it provides any benefits for click events, though I could be wrong on that front.

ollips: You're right, it's not needed here. Updated

css/src/components/banner.scss
wibjorn: In keeping with classes for styling, we'd need a class: `.dismissable`

site/src/components/dismiss.md
wibjorn: Hey that is cool! I am starting to think that @BenDMyers is correct though. Maybe our use cases are simple enough that we don't require a specific testId selector in Atlas. The unfortunate thing is that they clutter the code examples.

If we really want to keep the ids (my vote is to remove), let's keep them the same as our other repos, and use in our playwright config.

```
use: {
  testIdAttribute: 'data-test-id'
}
```

css/src/components/dismiss.scss
```diff
@@ -0,0 +1,23 @@
+.dismiss {
+	@include transparent-effects;
+
+	display: inline-block;
+	width: 1em;
+	height: 1em;
```
BenDMyers, line 6: Should these also be `rem` 🤔

site/src/components/dismiss.md
ollips: Updated test to not use playwright's id at all, so examples look clean.

css/src/components/dismiss.scss
ollips: No, these need to adjust per possible atomic usage.

css/src/components/banner.scss
ollips: Updated. What do you think about using `<component>-dismissable` class instead? Might align better with practices.

wibjorn: yeah that sounds good to me, good idea!

BenDMyers (dismissed):

homing1 (approved):

css/src/core/animations.scss
```diff
@@ -24,3 +24,24 @@
...
+		opacity: 0;
+	}
+}
+
+@keyframes fade-out {
+	100% {
+		opacity: 0;
+	}
+}
+
```
wibjorn, line 40: Can these be wrapped in our reduced motion no preference? 
```scss
@media (prefers-reduced-motion: no-preference) {
.animation-fade {
	animation: fade-out 0.4s ease-in forwards;
}

.animation-slide-up {
	animation: slide-up-fade-out 0.4s ease-in forwards;
}
}
```
This should prevent these classes from having an effect if a user has set their preference to reduce.

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

js/src/behaviors/dismiss.ts
```diff
@@ -0,0 +1,42 @@
...
+		if (!dismissTarget) {
+			return;
+		}
+
+		dismissElement(dismissTarget);
+	});
+}
+
+async function dismissElement(dismissTarget: Element) {
+	const prefersReducedMotionReduceQuery = window.matchMedia('(prefers-reduced-motion: reduced)');
```
wibjorn, line 20: Thinking that this can be remove in favor of a media query, see `animations.scss`

ollips: The reason I did it here is to avoid 5s delay.

The alternative would be to use `animation-fade` for all cases, cause from my understanding `prefers-reduced-motion: reduced` doesn't necessarily mean no animation.

js/src/behaviors/dismiss.ts
```diff
@@ -0,0 +1,42 @@
...
+
+		if (!dismissTarget) {
+			return;
+		}
+
+		dismissElement(dismissTarget);
+	});
+}
+
+async function dismissElement(dismissTarget: Element) {
```
BenDMyers, line 19: I'd like to propose an alternative approach that lets us keep the `prefers-reduced-motion` logic in the CSS, avoid any delays in removing the element when the users prefers reduced motion, and lets us flex to any arbitrary animation timing: the [`animationend` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/animationend_event).

This would need two things:
1. In the CSS, use query `prefers-reduced-motion: reduce` media to set `animation-duration: 0s`.
2. Refactor this function to look like this:

```ts
async function dismissElement(dismissTarget: Element) {
	dismissTarget.addEventListener('animationend', () => {
		dismissTarget.remove();
	});

	const dismissAnimation = dismissTarget.getAttribute('data-dismiss-animation');

	if (dismissAnimation) {
		dismissTarget.classList.add('is-dismissed');
		switch (dismissAnimation) {
			case 'slide-up':
				dismissTarget.classList.add('animation-slide-up');
				break;
			case 'fade':
				dismissTarget.classList.add('animation-fade');
				break;
		}
	}

	// Notification about content update occurrence.
	window.dispatchEvent(new CustomEvent('dismiss-content-update'));
}
```

From my local testing, this should work.

BenDMyers: (Alternatively, the `addEventListener` call could/should probably go inside the `if (dismissAnimation)` block.

BenDMyers: An `await new Promise(resolve => setTimeout(resolve, 100)` could go inside the event listener to add extra buffer onto the end of the animation duration before the element is removed.

js/src/behaviors/dismiss.ts
```diff
@@ -0,0 +1,39 @@
...
+	if (dismissAnimation) {
+		switch (dismissAnimation) {
+			case 'slide-up':
+				dismissTarget.classList.add('animation-slide-up');
+				break;
+			case 'fade':
+				dismissTarget.classList.add('animation-fade');
+				break;
+		}
+	}
```
BenDMyers, line 35: I noticed a case I missed with my suggestion: when there's no specified animation. That's on me, sorry!

I think it might be worth an `else` check here.

```suggestion
	} else {
		dismissTarget.remove();
    }
```

BenDMyers (dismissed):

js/src/behaviors/dismiss.ts
ollips: Thanks for sharing your idea, I slightly modified and applied it :)

js/src/behaviors/dismiss.ts
ollips: Updated with Ben's suggestion, check the comment below

css/src/core/animations.scss
ollips: Updated with Ben's suggestion, check the comment below

BenDMyers (approved):

homing1 (approved):

site/src/components/dismiss.md
```diff
@@ -0,0 +1,90 @@
...
+description: The dismiss component in the Atlas Design System
+template: standard
+classType: Component
+classPrefixes:
+  - dismiss
+---
+
+# Dismiss
+
+Dismiss is a subcomponent can be used within any other component that can be removed from the page. For example, modal, notification, banner, etc.
```
wibjorn, line 12: `component`

site/src/components/dismiss.md
```diff
@@ -0,0 +1,90 @@
...
+
+Dismissal functionality requires two parts:
+
+1. The dismiss component, which has the `data-dismiss` attribute
+2. The parent component to dismiss, which has the `data-dismissable` attribute
+
+Below is an example usage with `notification` component.
+
+```html
+<div class="notification" id="example-notification-01" data-dismissable>
```
wibjorn, line 24: Can we declutter this example but removing all but the most basic parts?

site/src/components/dismiss.md
```diff
@@ -0,0 +1,90 @@
...
+### Animating disappearance
+
+To animate component's disappearance, add `data-dismiss-animation` attribute to the dismissable component with one of these values:
+
+- `fade` will make component slowly fade out.
+- `slide-up` will make component fading while sliding up.
+
+```html
+<div
+	class="notification"
```
wibjorn, line 50: Same thought, declutter examples?

site/src/components/banner.md
```diff
@@ -30,6 +32,35 @@ Ensure the user knows they need to wait for some event in order to interact with
...
-	<p>Loading banner.</p>
+    <div class="banner-content">
+	    <p>Loading banner.</p>
+    </div>
+</div>
+```
+
+### Dismissing banner
+
+Banner can be made dismissable with the [`dismiss` subcomponent](./dismiss.md).
```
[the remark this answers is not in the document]
ollips: Updated

site/src/components/dismiss.md
ollips: Updated

site/src/components/dismiss.md
ollips: Updated

site/src/components/dismiss.md
ollips: Updated

site/src/components/dismiss.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+**Note:** the dismiss styles/positioning can be adjusted within the component it is applied to.
+
+### Animating disappearance
+
+To animate component's disappearance, add `data-dismiss-animation` attribute to the dismissable component with one of these values:
+
+- `fade` will make component slowly fade out.
+- `slide-up` will make component fading while sliding up.
```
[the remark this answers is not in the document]
ollips: Updated!

BenDMyers (approved):

js/src/behaviors/dismiss.ts
```diff
@@ -0,0 +1,41 @@
...
+
+		dismissElement(dismissTarget);
+	});
+}
+
+function dismissElement(dismissTarget: Element) {
+	const dismissAnimation = dismissTarget.getAttribute('data-dismiss-animation');
+
+	if (dismissAnimation) {
+		dismissTarget.addEventListener('animationend', () => {
```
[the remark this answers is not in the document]
ollips: I don't think that's something needed right now. All possible values are locked in and are .4s. Let's get back to it if it will ever becomes a problem.

homing1 (approved):

PR #542, opened 2023-05-19
add check to set is-loading on submitted button

ncmarsh:
Task: [task-834886](https://dev.azure.com/ceapex/Engineering/_workitems/edit/834886)

Currently if there is more than one submit button on a form (Q&A answer rating and Q&A moderator edit forms for example), when one submit button is clicked, they both display `is-loading` class, which is confusing to users. This PR adds a check to `setBusySubmitButton` to check for the submitter button to add `is-loading` class to that button and `disabled` status to the remaining submit buttons.

## Testing

This is how I tested this on the Q&A site:
1. `git checkout main; git pull; git checkout nmarshall/qna-update-setbusysubmitbutton; npm run build:js`
2. Copy the `js/dist` folder
3. In `docs-ui`, delete the existing `node_modules/@microsoft/atlas-js/dist` folder
4. **Note**: you may need to run `git clean -fxd; npm i; then do the above step
5. `npm run dev` and go to Q&A https://learn-microsoft-com.local/en-us/answers/questions/
6. Test various forms, below I've listed Q&A form examples, but please test elsewhere too
- Ask a question/add answer/add comment/follow/etc. -- Should be no change
- Editing as a moderator, which has two buttons -- The clicked button should have `is-loading`, the other should be disabled
- On your own question with answers from others (feel free to ping me if you need answers), rate Yes or No on the answer rating form -- The clicked button should have `is-loading`, the other should be disabled

## Contributor checklist

- [X] Did you update the description of this pull request with a review link and test steps?
- [X] Did you submit a changeset? Changesets are required for all code changes.
- [X] Does your pull request implement complex UI logic (js/ts)? Consider adding an integration test to test your user flow.

### DIFF
--- a/.changeset/silly-turkeys-count.md
+++ b/.changeset/silly-turkeys-count.md
@@ -2,4 +2,4 @@
 '@microsoft/atlas-js': patch
 ---
 
-Update setBusySubmitButton to check for submitter button
+Add submitter check to ensure only one submit button displays loading state on forms with multiple submit buttons.
--- a/.changeset/silly-turkeys-count.md
+++ b/.changeset/silly-turkeys-count.md
@@ -1,0 +1,5 @@
+---
+'@microsoft/atlas-js': patch
+---
+
+Update setBusySubmitButton to check for submitter button
--- a/js/src/elements/form-behavior.ts
+++ b/js/src/elements/form-behavior.ts
@@ -626,13 +626,10 @@ function normalizeInputValue(target: EventTarget | null) {
 }
 
 function setBusySubmitButton(event: Event, form: HTMLFormElement, isLoading: boolean) {
-	const submitter =
-		(event as SubmitEvent).submitter instanceof HTMLButtonElement
-			? (event as SubmitEvent).submitter
-			: null;
+	const submitter = (event as SubmitEvent).submitter;
 	Array.from(form.elements).forEach(element => {
 		if (element instanceof HTMLButtonElement && element.type === 'submit') {
-			if (!submitter || submitter === element) {
+			if (submitter && submitter === element) {
 				element.classList.toggle('is-loading', isLoading);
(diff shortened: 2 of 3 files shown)


ncmarsh:
@disimmon @jdanyow @homing1 Could you take a look at this and let me know what you think :)

js/src/elements/form-behavior.ts
```diff
@@ -625,10 +625,18 @@ function normalizeInputValue(target: EventTarget | null) {
 	}
 }
 
-function setBusySubmitButton(form: HTMLFormElement, isLoading: boolean) {
+function setBusySubmitButton(event: Event, form: HTMLFormElement, isLoading: boolean) {
+	const submitter =
```
homing1, line 629: If there is no submitter, wouldn't it toggle every submit button in the form?

I wonder if this line can be simplified to `const submitter = (event as SubmitEvent).submitter;`

Then on line 635, it could be `if (submitter && submitter === element)`

If element has to be submit button element and submitter has to equal the element, then the submitter would also have to be a button element.

ncmarsh: This makes sense to me, I've updated accordingly

.changeset/silly-turkeys-count.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-js': patch
+---
+
+Update setBusySubmitButton to check for submitter button
```
homing1, line 5: How about something like "Add submitter check to ensure only one submit button displays loading state on forms with multiple submit buttons."

ncmarsh: Much better, thanks!

homing1 (approved):

disimmon (approved):

PR #543, opened 2023-05-22
Adding components.template file

ollips:
Link: preview-543

Adding `components.template` file which is a good starting point when writing a doc page for a new component. Similar to the `atomics.template`.

### DIFF
--- a/site/src/components.template
+++ b/site/src/components.template
@@ -14,10 +14,6 @@ The first thing is the general description.
 
 Consider using tables to organize information if there are a lot of nuances.
 
-| cssproperty | value                           | screensize |
-| ----------- | ------------------------------- | ---------- |
-| `<name>`    | `<value>`, `<value>`, `<value>` | `<size>`   |
-
 | State       | Class                  | Example        |
 | ----------- | ---------------------- | -------------- |
 | `<name>`    | `<value>`              | html```Demo``` |
--- a/site/src/components.template
+++ b/site/src/components.template
@@ -1,0 +1,43 @@
+---
+title: Title
+description: The <component-name> component in the Atlas Design System
+template: standard
+figmaEmbed: <figma-url> allowfullscreen
+classType: Component
+classPrefixes:
+  - <component-name>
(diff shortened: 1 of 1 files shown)


site/src/components.template
```diff
@@ -0,0 +1,43 @@
...
+  - <component-name>
+---
+
+# Page title (component name)
+
+The first thing is the general description.
+
+Consider using tables to organize information if there are a lot of nuances.
+
+| cssproperty | value                           | screensize |
```
wibjorn, line 17: This is for atomics, isn't it?

site/src/components.template
```diff
@@ -0,0 +1,43 @@
...
+
+```html
+<div>
+  <button class="button">Click me</button>
+</div>
+```
+
+### Modifier heading
+
+If there are components variations available such as different semantic colors on buttons, different hero sizes, loading states, etc., the next section(s) should be about them.
```
homing1, line 37:
```suggestion
If there are components variations available such as different semantic colors, sizes, loading states, etc., the next section(s) should be about them.
```

site/src/components.template
ollips: Tables can still be used in components docs, so I thought it worth mentioning

- https://design.learn.microsoft.com/components/button.html
- https://design.learn.microsoft.com/components/checkbox.html
- https://design.learn.microsoft.com/components/gradients.html
- https://design.learn.microsoft.com/components/help.html

site/src/components.template
ollips: Updated, thanks!

site/src/components.template
wibjorn: They can, but they won't have the headings `cssproperty | value | screensize`. I would suggest we remove it or at the very least update the headings.

ollips: Good point, kept the one with ` State | Class | Example ` titles

wibjorn (approved):

homing1 (approved):

PR #556, opened 2023-06-30
Adds Aspect ratio atomic

jwmunn:
Task: task-862482

Link: preview-556

When videos are added and the container is wider or taller than the aspect ratio of the video it creates black bars in the video player. These ratio atomics address this issue by allowing the player to scale according to the ratio when applied directly to the `iframe` or `video` element. It can also be used to crop images to a certain aspect ratio. This can also be applied to images but will stretch the img.

## Testing

1. Visit https://design.learn.microsoft.com/pulls/556/atomics/aspect-ratio.html
2. Open dev tools and verify classes are avialable.
   Expected result: Classes are avialable and scale element to different ratios.

## Contributor checklist


.changeset/tasty-fishes-tease.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-css': major
```
BenDMyers, line 2: Since this doesn't introduce a breaking change, this should either be a `minor` or a `patch`.

css/src/atomics/ratio.scss
```diff
@@ -0,0 +1,8 @@
+@each $name, $ratio in $ratios {
+	.ratio-#{$name} {
+		aspect-ratio: #{$ratio};
+		width: 100% !important; // must override width to fill ratio
+		height: 100% !important; // must override height to fill ratio
+		object-fit: cover;
```
BenDMyers, line 6: As a thought, would it be worth introducing an `object-fit-cover` atomic, so folks can opt into using that? That way, these ratio atomics can do somewhat fewer things at once.

.changeset/tasty-fishes-tease.md
jwmunn: I thought net new atomics, components, were a major change and extensions of existing atomics components were a minor change?

css/src/atomics/ratio.scss
jwmunn: We can do this but it will stretch the image (same behavior as the aspect ratio in mdn). Maybe we can just use video examples?

css/src/atomics/ratio.scss
```diff
@@ -0,0 +1,8 @@
+@each $name, $ratio in $ratios {
+	.ratio-#{$name} {
```
BenDMyers, line 2: Been thinking about this naming compared to our standard naming convention. We typically use the full property name, so this should probably be `.aspect-ratio-#{$name}` to keep with convention.

css/src/tokens/ratios.scss
```diff
@@ -0,0 +1,10 @@
+/**
+ * @sass-export-section="ratios"
+ */
+
+$ratios: (
+	'16-9': '16 / 9',
+	'4-3': '4 / 3',
+	'square': '1 / 1',
+	'vertical': '9 / 16'
```
BenDMyers, line 9: This should probably be `9-16` instead of `vertical`, seeing as a hypothetical `3 / 4` would also count as vertical.

I think `square` is the aspect ratio where a non-numeric name makes sense to me.

css/src/tokens/ratios.scss
```diff
@@ -0,0 +1,10 @@
+/**
+ * @sass-export-section="ratios"
+ */
+
+$ratios: (
```
BenDMyers, line 5: While we're adding these atomics, we should probably port over these ratios found in `packages\styles\src\elements\_image.scss` as well for full compatibility, so we can fully deprecate the older ratio utility classes:

- 5 / 4
- 3 / 2
- 2 / 1
- 3 / 1
- 4 / 5
- 3 / 4
- 2 / 3
- 1 / 2
- 1 / 3

I _think_ that's it.

site/src/atomics/ratio.md
```diff
@@ -0,0 +1,61 @@
+---
+title: Ratio Atomics
```
BenDMyers, line 2:
```suggestion
title: Aspect ratio
```

The other atomic pages' titles don't include "Atomics," so this helps maintain consistency in the lefthand nav.

site/src/atomics/ratio.md
```diff
@@ -0,0 +1,62 @@
...
+  	<video controls name="media" class="width-500 aspect-ratio-16-9" poster="~/src/scaffold/media/sample-learn-video-thumbnail.jpg" src="~/src/scaffold/media/sample-learn-video.mp4" type="video/mp4" />
+  	</video>
+  </div>
+```
+
+```html
+<div class="width-250 padding-sm">
+	<iframe
+		class="aspect-ratio-9-16"
+		src="~/src/scaffold/media/sample-learn-vertical-video.mp4"
```
BenDMyers, line 33: This video autoplays when I hit the page, audio and all, which is kind of startling. Is there a way we can avoid that?

BenDMyers: Weird. Can't seem to reproduce that now 🤔

site/src/atomics/ratio.md
```diff
@@ -0,0 +1,62 @@
+---
+title: Ratio
```
BenDMyers, line 2:
```suggestion
title: Aspect ratio
```
to align a little more with the property name?

site/src/atomics/ratio.md
jwmunn: Yea I pushed a fix for that using `sandbox` attribute on the iframe.

css/src/atomics/ratio.scss
BenDMyers: This hasn't been changed yet, despite the uses of it being updated.

css/src/atomics/ratio.scss
jwmunn: Makes sense for atomics, I'll do this in a separate PR.

css/src/atomics/ratio.scss
jwmunn: It's resolved

site/src/atomics/ratio.md
jwmunn: Went ahead and updated file names to match also.

site/src/scaffold/media/sample-learn-video.mp4
homing1, line 1: These videos are rather large. Are there shorter videos or generic short example videos that can be used instead?

homing1: Or perhaps do we even need a video example?

jwmunn: Good call, added smaller sample video files.

These classes are being added mainly for videos not images, images need additional classes to not be stretched. @BenDMyers suggested another atomic for `object-fit` and maybe `object-position` being added. The aspect atomic alone works on videos and prevents black bars from being added opn the tops or sides like on the current Learn homepage
image

jwmunn:
@BenDMyers @homing1 any idea why these tests are failing?

css/src/atomics/aspect-ratio.scss
```diff
@@ -0,0 +1,8 @@
+@each $name, $ratio in $aspect-ratios {
+	.aspect-ratio-#{$name} {
+		aspect-ratio: #{$ratio};
+		width: 100% !important; // must override width to fill ratio
+		height: 100% !important; // must override height to fill ratio
+		// object-fit: cover;
```
jwmunn, line 6: this needs to be removed

jwmunn:
> @BenDMyers @homing1 any idea why these tests are failing? It's timing out but the error messages aren't very helpful.

Figured it out, it was the `iframe`, which makes sense seeing as we had to make an exception for the `#figma-embedded-iframe` so I've updated that and everything is ready for another review.

css/src/tokens/aspect-ratios.scss
```diff
@@ -0,0 +1,19 @@
+/**
+ * @sass-export-section="aspect-ratios"
+ */
+$aspect-ratios: (
```
wibjorn, line 4: I'm in favor of keeping this closer to the atomics. Is this a useful token? Does it need sharing between components?

css/src/tokens/aspect-ratios.scss
```diff
@@ -0,0 +1,19 @@
...
+	'square': '1 / 1',
```
wibjorn, line 5: shall we keep 1-1 so that our atomic names continue to follow their conventions?

wibjorn (changes requested):

css/src/tokens/aspect-ratios.scss
```diff
@@ -0,0 +1,19 @@
...
+	'1-2': '1 / 2',
+	'1-3': '1 / 3',
+	'2-1': '2 / 1',
+	'2-3': '2 / 3',
+	'3-1': '3 / 1',
+	'3-2': '3 / 2',
+	'4-3': '4 / 3',
+	'4-5': '4 / 5',
+	'3-4': '3 / 4',
+	'5-4': '5 / 4',
```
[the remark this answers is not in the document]
BenDMyers: @wibjorn These were added in because we already have these aspect ratios in docs-ui's `image.scss` file, with the goal of making sure people could migrate any preexisting `is-[ratio]` class to its Atlas alternative.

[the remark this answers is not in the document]
BenDMyers: Might be worth seeing if any of those classes are cullable, if they're either not in use or could be replaced with a similar enough ratio. But as it stands, I'm not sure to what degree we want some, but not all, of the ratios to be available in Atlas.

site/src/atomics/aspect-ratio.md
```diff
@@ -0,0 +1,86 @@
...
+classType: Atomics
+classPrefixes:
+  - aspect-ratio
+---
+
+# Aspect ratio Atomics
+
+Ratio atomics are used to add a fluid aspect ratio to an element. The element will scale to fill 100% width of the container while matching the height to the defined aspect ratio.
+
+| modifiers | ratio                                                                                   |
```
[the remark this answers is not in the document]
jwmunn: Updated

css/src/tokens/aspect-ratios.scss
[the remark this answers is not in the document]
jwmunn: Did an audit of the image.scss classes not being used. Left 4-3 because it's a common video ratio, and added 4-1 because I found that many of the images on the site use that ratio even though it isn't included in the image.scss file.

css/src/tokens/aspect-ratios.scss
jwmunn: updated

css/src/tokens/aspect-ratios.scss
jwmunn: Deleted token, moved to atomic style file.

BenDMyers (approved):

site/src/atomics/aspect-ratio.md
```diff
@@ -0,0 +1,74 @@
...
+	></iframe>
+</div>
+```
+
+```html
+<div class="width-100 padding-xxs">
+	<img
+		class="aspect-ratio-1-1"
+		alt=""
+		role="presentation"
```
wibjorn, line 47: Wondering if keeping the image around might be better for unsighted developers.

```
alt="An example image use to show an aspect ratio of 1 to 1."
```
then remove `role="presentation"`?

site/src/atomics/aspect-ratio.md
```diff
@@ -0,0 +1,74 @@
+---
```
wibjorn, line 1: Given just how useful (very!) these will be for images. Perhaps it's worth a small section in the Image component documentations that mentions aspect-ratio atomics and links to this page?

https://design.learn.microsoft.com/components/image.html

css/src/tokens/aspect-ratios.scss
wibjorn: Yeah, we don't need to follow the older ways, more important to keep things consistent with the new conventions.

site/src/atomics/aspect-ratio.md
```diff
@@ -0,0 +1,74 @@
...
+  - aspect-ratio
+---
+
+# Aspect Ratio Atomics
+
+Ratio atomics are used to add a fluid aspect ratio to an element. The element will scale to fill 100% width of the container while matching the height to the defined aspect ratio.
+
+| cssproperty  | ratio                                             | screensize |
+| ------------ | ------------------------------------------------- | ---------- |
+| aspect-ratio | `1-1`, `2-1`, `3-2`, `4-3`, `5-3`, `9-16`, `16-9` | N/A        |
```
wibjorn, line 16: 4-3 and 5-3 definitely still needed?

css/src/atomics/aspect-ratio.scss
```diff
@@ -0,0 +1,17 @@
...
+	'4-3': '4 / 3',
+	'5-3': '5 / 3',
+	'9-16': '9 / 16',
+	'16-9': '16 / 9'
+) !default;
+
+@each $name, $ratio in $aspect-ratios {
+	.aspect-ratio-#{$name} {
+		aspect-ratio: #{$ratio};
+		width: 100% !important; // must override width to fill container
```
wibjorn, line 14: Just for my own edification, what happens if we don't include width and height here? My inclination is to not include these rules, but I could be convinced otherwise.

css/src/atomics/aspect-ratio.scss
```diff
@@ -0,0 +1,17 @@
...
+	'3-2': '3 / 2',
+	'4-3': '4 / 3',
+	'5-3': '5 / 3',
+	'9-16': '9 / 16',
+	'16-9': '16 / 9'
+) !default;
+
+@each $name, $ratio in $aspect-ratios {
+	.aspect-ratio-#{$name} {
+		aspect-ratio: #{$ratio};
```
wibjorn, line 13: Atomics rule declarations should use `!important`

integration/tests/accessibility.spec.ts
```diff
@@ -65,6 +66,7 @@ test.describe('axe should not find any automatically detectable accessibility is
 
 			const accessibilityScanResults = await new AxeBuilder({ page })
 				.exclude('#figma-embed-iframe')
+				.exclude('#embed-iframe')
```
wibjorn, line 69: How necessary is the `iframe` example? Given it isn't accessible by default, and that we can show outcomes with `video` or `img`, could we just mention it works with iframes instead of adding one?

css/src/atomics/aspect-ratio.scss
jwmunn: It covers some gaps and allows it to scale to the container. So it's mainly there to keep the style agnostic to other style/class dependencies. For instance docs-ui defines width 100% for iframes but no height, so even with the aspect ratio property you get the black bars. Similarly if an image is smaller than the container, it doesn't match the ratio. If its >= to the container width and doesn't have 100% height then it also doesn't match the aspect ratio (notice https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio sets width: auto, height: 100%)

integration/tests/accessibility.spec.ts
jwmunn: Good point, removed

css/src/atomics/aspect-ratio.scss
jwmunn: updated

site/src/atomics/aspect-ratio.md
jwmunn: Removed 3-2 and 5-3. Left 4-3 because it's a standard video and monitor ratio.

site/src/atomics/aspect-ratio.md
```diff
@@ -0,0 +1,74 @@
...
+<div class="width-100 padding-xxs">
+	<img
+		class="aspect-ratio-1-1"
+		alt=""
+		role="presentation"
+		src="~/src/scaffold/media/aspect-ratio-square.png"
+	/>
+</div>
+<div class="width-200 padding-xxs">
+	<img
```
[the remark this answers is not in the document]
jwmunn: done

site/src/atomics/aspect-ratio.md
jwmunn: done

BenDMyers (approved):

css/src/atomics/aspect-ratio.scss
wibjorn: After a little testing, I think that we cannot go the route of placing `height: 100%` on this, because this implies the height of a parent element. You can test the undesirable behavior by moving the video element out of its current spot (using dev tools) and putting it as a child of a tall parent element. In this case, I did this with the parent element being `#article`. Things get very tall very fast. 

image

Good news is I'm not sure we need it. Aspect ratio seems to work fine without it, if we apply an object-fit property as well. I think we _could remove height_ too because our reset applies `max-width: 100%` to images and videos. That said, I think we may want to consider apply `max-width: 100%` for safety, and so that other elements can get the same.

wibjorn: Cancel that thing about width, think we _do_ need width

jwmunn: @wibjorn after more testing and reflecting on the use case that brought this proposal up I think we should keep it a true atomic without a width or height. I've removed those from the style.

Thoughts behind this are:
1. The iframe video on the homepage had black bars because a width/height attribute were set on the iframe element. Without those attributes set the `aspect-ratio` style resolves the issue.
2. Images that are less than 100% width don't maintain the ratio but a majority of our images are larger than their containers, and we also have an `image` class that sets the img width to 100% which when used with this atomic would allow the ratio to be applied to smaller images.
3. This keeps it more inline with the principle of atomics and other classes that target width or object fit can be added for special use cases.
4. The width/height 100% !important was mainly to force the ratio on an element even if a different width or height are applied with attributes or inline styles, but that feels out of scope of an atomic.

BenDMyers (dismissed):

css/src/atomics/aspect-ratio.scss
wibjorn: Great! That was my initial inclination as well.

site/src/atomics/image.md
```diff
@@ -10,7 +10,7 @@ classPrefixes:
 
 # Image
 
-Image atomics are used to set the image behavior of an element.
+Image atomics are used to set the image behavior of an element. Image height and width ratios can be set with [aspect ratio atomics](./aspect-ratio.md).
```
wibjorn, line 13: Ah sorry @jwmunn I originally meant that this warranted mentioning in the image component. I think its definitely fine to keep this line here, but it needs to be added to this page as well: https://design.learn.microsoft.com/components/image.html

site/src/atomics/aspect-ratio.md
```diff
@@ -0,0 +1,61 @@
...
+description: Aspect ratio related atomic css classes for the Atlas Design System
+template: standard
+classType: Atomics
+classPrefixes:
+  - aspect-ratio
+---
+
+# Aspect Ratio Atomics
+
+Aspect ratio atomics are used to set the width-to-height ratio of an element, even if the parent container or viewport size changes.
```
wibjorn, line 12: Suggestion only: perhaps write up your general views on using width with this? Something like: 

> ... Note that in some scenarios, elements will need to have `width: 100%` and/or `height: 100%` set. This can be done with the use of atomics such as `width-full` or a component like [`image](./components.

I don't believe we have any height atomics at the moment, so depending on your thoughts, we can commit that part of the above statement.

site/src/atomics/aspect-ratio.md
```diff
@@ -0,0 +1,73 @@
...
+		class="aspect-ratio-16-9"
+		src="~/src/scaffold/media/aspect-ratio-video-16-9.mp4"
+		type="video/mp4"
+	></video>
+</div>
+```
+
+## Additional information
+
+In some scenarios an element's intrincis size is smaller than it's container size, so the ratio isn't calculated. These elements will need to have `width: 100%` or set to calculate the height. This can be done with the use of [width atomics](./width.md) or [ image components](../components/image.md).
```
homing1, line 36:
```suggestion
In some scenarios an element's intrinsic size is smaller than its container size, so the ratio isn't calculated. These elements will need to have its width set to `width: 100%` or set to calculate the height. This can be done with the use of [width atomics](./width.md) or [ image components](../components/image.md).
```

site/src/atomics/aspect-ratio.md
```diff
@@ -0,0 +1,73 @@
...
+
+Aspect ratio atomics are used to set the width-to-height ratio of an element, even if the parent container or viewport size changes.
+
+| cssproperty  | ratio                               | screensize |
+| ------------ | ----------------------------------- | ---------- |
+| aspect-ratio | `1-1`, `2-1`, `4-3`, `9-16`, `16-9` | N/A        |
+
+## Usage
+
+Here are examples of aspect ratio atomics being applied to `video` and `img` elements. When used on `video` or `iframe` elements with the matching aspect ratio, it will remove the black bars and scale responsively.
```
homing1, line 20: Can we add a little more context on what "the black bars" are and why they occur in the first place?

site/src/atomics/aspect-ratio.md
```diff
@@ -31,14 +31,26 @@ Here are examples of aspect ratio atomics being applied to `video` and `img` ele
...
+		<img
+			class="aspect-ratio-1-1"
+			alt="An example image with an aspect ratio of 1 by 1"
+			src="~/src/scaffold/media/aspect-ratio-square.png"
+		/>
+	</div>
 </div>
+```
+
+In other scenarios an element has a different intrinsic aspect ratio. Change the ratio without
```
wibjorn, line 50: Changing

homing1 (dismissed):

wibjorn (dismissed):

wibjorn (approved):

homing1 (approved):

PR #563, opened 2023-07-14
update form behavior locale strings

jim-benson:
Task: task-[866250](https://ceapex.visualstudio.com/Engineering/_sprints/taskboard/Q%20and%20A/Engineering/2023-07-12?workitem=866250)

Link: preview-[562](https://design.learn.microsoft.com/pulls/562)

We found that the form-behavior is bein initiated before the contents loads and when the locStrings tries to run there are no attributes yet.

So the thought is to have the locaStrings run in the `connectedCallback` instead of the constructor so the content can be loaded before the locStrings update.

### DIFF
--- a/.changeset/strange-goats-compare.md
+++ b/.changeset/strange-goats-compare.md
@@ -1,0 +1,5 @@
+---
+'@microsoft/atlas-js': patch
+---
+
+Update when the locale strings get updated during initization
--- a/js/src/elements/form-behavior.ts
+++ b/js/src/elements/form-behavior.ts
@@ -22,16 +22,7 @@ export class FormBehaviorElement extends HTMLElement {
 	toDispose: (() => void)[] = [];
 	isDirty = false;
 	commitTimeout = 0;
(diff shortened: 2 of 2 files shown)


disimmon (dismissed):

chrispydizzle (approved):

.changeset/strange-goats-compare.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-js': patch
+---
+
+Update when the locale strings get updated during initization
```
homing1, line 5: How about something like this?
```suggestion
Defer form behavior locale strings from loading until after being added to the DOM.
```

homing1 (changes requested):

.changeset/strange-goats-compare.md
jim-benson: updated

homing1 (approved):

BenDMyers (approved):

PR #571, opened 2023-09-05
Persona component

ollips:
Task: task-889937

Link: [preview](https://design.learn.microsoft.com/pulls/571/components/persona.html)

Adding persona component.

## Testing

1. Review code
2. Review doc page


css/src/components/persona.scss
```diff
@@ -0,0 +1,82 @@
...
+			display: block;
+			width: 100%;
+			height: auto;
+			max-height: 100%;
+		}
+
+		span {
+			line-height: 1;
+			text-transform: uppercase;
+		}
```
BenDMyers, line 40: I'd like to propose we don't support styled text for avatars at all. Instead, when we want to use initials as a fallback avatar, I think we should encourage [an approach we're already using in Q&A for fallback avatar generation](https://dev.azure.com/ceapex/Engineering/_git/docs-ui/pullrequest/757907?_a=files), where we generate an SVG as a data URI, and then use that data URI as an `img`'s `src`.

Possibly, a utility for generating that data URI could even exist in Atlas's scripts - I could go either way on that front.

Using styled text for fallback avatars...
- Can cause problems if the user has changed their browser or device-level base font size
- Means styled text avatars will be modified, probably unexpectedly, by things like operating system forced colors mode/contrast themes
- Introduces variation in markup, which complicates rendering logic and means different "avatars" announce differently to assistive technologies

ollips: I was looking into how we currently treat the initials, and saw that mostly those are generated SVGs (Yay!). Though I though it's good to have that option just in case, as a framework we should consider other cases besides learn.

But I was actually looking for your feedback on what would be best to do with initials markup and if we should assign it `aria-label="user name"` or something else so it announced the same way as the image avatar 🤔

BenDMyers: `aria-label` doesn't work on `span`s out of of the box, since `span`s have `generic` roles by default.

_If_ we did want to both support `aria-label` and ensure announcements are consistent with those of actual image avatars, we could use `role="img"`:

```html
<span role="img" aria-label="Avatar for Jane Doe">
    JD
</span>
```

css/src/components/persona.scss
```diff
@@ -0,0 +1,82 @@
...
+		display: flex;
+		align-items: center;
+		justify-content: center;
+		width: $persona-default-size;
+		height: $persona-default-size;
+		border-radius: $border-radius-rounded;
+		background-color: $persona-avatar-background-color;
+		color: $persona-avatar-font-color;
+		font-size: $persona-default-size * $persona-dimension-font-size-ratio;
+		overflow: hidden;
```
wibjorn, line 27: `overflow: hidden` required?

css/src/components/persona.scss
```diff
@@ -0,0 +1,82 @@
...
+$persona-metadata-font-color: $text-subtle !default;
+$persona-dimension-font-size-ratio: 0.45;
+
+.persona {
+	display: flex;
+	position: relative;
+	gap: $persona-gap-size;
+	font-size: $persona-font-size;
+
+	.persona-avatar {
```
wibjorn, line 17: I want to double check that we're able to put the avatar / circular image elsewhere if we need to. See @jim-benson' s PR.

css/src/components/persona.scss
```diff
@@ -0,0 +1,82 @@
...
+			max-height: 100%;
+		}
+
+		span {
+			line-height: 1;
+			text-transform: uppercase;
+		}
+	}
+
+	.persona-description :not(:first-child) {
```
wibjorn, line 43: Can you explain why `:not(:first-child)` is needed here for me?

css/src/components/persona.scss
```diff
@@ -0,0 +1,82 @@
...
+		color: $persona-metadata-font-color;
+		font-size: $persona-metadata-font-size;
+	}
+
+	// Sizes
+
+	/* stylelint-disable declaration-property-unit-allowed-list, max-nesting-depth */
+	@each $dimension in $persona-avatar-dimensions {
```
wibjorn, line 51: Rather than using the `.image` component convention (with pixels in the names), should we just use our normal component sizing convention of t-shirt sizes?

css/src/components/persona.scss
```diff
@@ -0,0 +1,82 @@
...
+				.persona-avatar {
+					width: $dimension * 1px;
+					height: $dimension * 1px;
+					font-size: $dimension * $persona-dimension-font-size-ratio * 1px;
+				}
+			}
+		}
+	}
+
+	/* stylelint-enable */
```
wibjorn, line 81: 🙈

css/src/components/persona.scss
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	.persona-description :not(:first-child) {
+		color: $persona-metadata-font-color;
+		font-size: $persona-metadata-font-size;
+	}
+
+	// Sizes
+
+	/* stylelint-disable declaration-property-unit-allowed-list, max-nesting-depth */
```
wibjorn, line 50: 🙈

css/src/components/persona.scss
wibjorn, line 43: Does `persona-description` need more subcomponents? I think we probably want to ensure we have defaults to avoid this situation I see in the document: 

image

css/src/components/persona.scss
```diff
@@ -0,0 +1,82 @@
+$persona-avatar-dimensions: 16 24 32 36 48 !default;
```
wibjorn, line 1: is 16 too small?

css/src/components/persona.scss
```diff
@@ -0,0 +1,82 @@
...
+$persona-avatar-background-color: $alternate-background-medium !default;
+$persona-avatar-font-color: $text-invert !default;
+$persona-default-size: 32px !default;
+$persona-font-size: $font-size-7 !default;
+$persona-gap-size: $layout-1 !default;
+$persona-metadata-font-size: $font-size-9 !default;
+$persona-metadata-font-color: $text-subtle !default;
+$persona-dimension-font-size-ratio: 0.45;
```
wibjorn, line 9: Should that component just use `em` sizing instead of a calculated font-size?

site/src/components/persona.md
```diff
@@ -0,0 +1,148 @@
...
+		<span>JD</span>
+	</div>
+	<div class="persona-description">
+		<a class="justify-self-stretch stretched-link" href="#">John Doe</a>
+		<p>Software Engineer</p>
+	</div>
+</div>
+```
+
+Persona might have both avatar and description presetned at the same time or just one of them.
```
homing1, line 59:
```suggestion
Persona might have both avatar and description presented at the same time or just one of them.
```

css/src/components/persona.scss
ollips: Yeah, it takes care of inner image. An alternative would be assigning the `border-radius` to the image itself.

css/src/components/persona.scss
ollips: It can be, the markup will look like this:

```html
<div class="persona">
    <figure class="persona-avatar">
        <img src="~/src/scaffold/media/avatar.png" alt="username" />
    </figure>
</div>
```

css/src/components/persona.scss
ollips: The idea behind it is that the first child will be the user's name and everything after it will be metadata with a bit different styling - font size is smaller and has a bit lighter text color. So I thought it's easier to target it this way rather than creating another class.

css/src/components/persona.scss
ollips: I don't know why I didn't think about t-shirt sizes right away, updated now!

css/src/components/persona.scss
ollips: Removed 😅

css/src/components/persona.scss
ollips: Same, removed!

css/src/components/persona.scss
ollips: Not sure what you mean here. Can you clarify please?

css/src/components/persona.scss
ollips: I updated the sizes to `28 / 32 / 36 / 42 / 48`, so the smallest font size for cases with initials doesn't get smaller than 12px. 

I also looked into what sized we use today and it's `32 / 36 / 48`. and we have `64 / 96` in profile, but I'm not feeling strongly about going that big.

css/src/components/persona.scss
ollips: I thought calculating was a good idea cause it would help keep the ratio perfectly the same for all sizes, but I can make another set of variables for font sizes

css/src/components/persona.scss
wibjorn: I think that's a better alternative - overflow is necessary at times, but it should still be a last resort. It affects focus outlines, popovers, etc, negatively.

css/src/components/persona.scss
wibjorn: Does it need `persona-description-name`? (I don't love the name) 

Does we really want bare elements in there? I'd think we should use classes to affect the anchor and paragraph.

css/src/components/persona.scss
wibjorn: If the ratios aren't kept, then we're probably missing a spot in a subcomponent where we should be using `em`.

.changeset/eleven-seals-nail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-site': minor
+---
+
+Adding doc page for persona component.
```
homing1, line 5:
```suggestion
Adding documentation page for persona component.
```

site/src/components/persona.md
```diff
@@ -0,0 +1,156 @@
...
+		<span>JD</span>
+	</div>
+	<div class="persona-description">
+		<p>John Doe</p>
+		<p>Software Engineer</p>
+	</div>
+</div>
+```
+
+Persona component might contain a link to user's profile, and with help of the [`.stretched-link` component](../components/stretched-link.md) the whole area persona takes becomes clickable:
```
homing1, line 45:
```suggestion

Persona description may contain a link to user's profile. With the help of [`.stretched-link` component](../components/stretched-link.md), the whole area persona becomes clickable:
```

css/src/components/persona.scss
ollips: updated!

css/src/components/persona.scss
ollips: Refactored this part

ollips: Updated! Actually used the `persona-name` class, and renamed `persona-description` to `persona-details`, for whatever reason today this sounded much better than description.

css/src/components/persona.scss
ollips: Refactored this, let me know what you think.

site/src/components/persona.md
ollips: Updated!

css/src/components/persona.scss
ollips: @wibjorn @homing1 any thoughts on this?

css/src/components/persona.scss
```diff
@@ -0,0 +1,203 @@
...
+			}
+		}
+	}
+
+	// Sizes
+
+	&.persona-xs {
+		font-size: $persona-avatar-xs;
+
+		.persona-avatar {
```
wibjorn, line 68: needs more `em` sizing?

css/src/components/persona.scss
```diff
@@ -0,0 +1,203 @@
...
+	&.persona-xl {
+		font-size: $persona-avatar-xl;
+
+		.persona-avatar {
+			width: $persona-avatar-xl;
+			height: $persona-avatar-xl;
+		}
+	}
+
+	@include tablet {
```
wibjorn, line 110: responsive sizes need initially? seems like we're adding a lot of cruft here?

css/src/components/persona.scss
```diff
@@ -0,0 +1,203 @@
...
+	display: flex;
+	position: relative;
+	gap: $persona-gap-size;
+	font-size: $persona-avatar-md;
+
+	.persona-avatar {
+		display: flex;
+		align-items: center;
+		justify-content: center;
+		width: $persona-avatar-md;
```
wibjorn, line 30:
```suggestion
		width: 1em;
```

css/src/components/persona.scss
```diff
@@ -0,0 +1,203 @@
...
+		height: $persona-avatar-md;
```
wibjorn, line 31:
```suggestion
		height: 1em;
```

css/src/components/persona.scss
```diff
@@ -0,0 +1,203 @@
+$persona-avatar-xs: 28px !default;
+$persona-avatar-sm: 32px !default;
+$persona-avatar-md: 36px !default;
+$persona-avatar-lg: 42px !default;
+$persona-avatar-xl: 48px !default;
+
+$persona-avatar-font-size: 0.45em;
```
wibjorn, line 7: audit for unused variables?

css/src/components/persona.scss
```diff
@@ -0,0 +1,203 @@
...
+
+	.persona-details {
+		align-self: center;
+		color: $persona-metadata-font-color;
+		font-size: $persona-metadata-font-size;
+
+		.persona-name {
+			font-size: $persona-name-font-size;
+
+			&:not(a) {
```
wibjorn, line 57: needed?

site/src/components/persona.md
```diff
@@ -0,0 +1,156 @@
...
+  - persona
+---
+
+# Persona
+
+Persona component comes in handy for user's representation. It consists of two parts - user's avatar and details. Details has their name and optional metadata.
+
+| Class pattern             | Sizes                              | screensize          |
+| ------------------------- | ---------------------------------- | ------------------- |
+| `.persona.persona-<size>` | `xs` `sm` `md (default)` `lg` `xl` | `tablet`, `desktop` |
```
wibjorn, line 17: how many sizes do we really need?

css/src/components/persona.scss
wibjorn: Fine to go with using SVGs only.

css/src/components/persona.scss
```diff
@@ -0,0 +1,203 @@
...
+	.persona-avatar {
+		display: flex;
+		align-items: center;
+		justify-content: center;
+		width: $persona-avatar-md;
+		height: $persona-avatar-md;
+		border-radius: $persona-avatar-border-radius;
+		background-color: $persona-avatar-background-color;
+		color: $persona-avatar-font-color;
+		font-size: $persona-avatar-font-size;
```
BenDMyers, line 35: I think the avatar height and the font size should be defined in similar units - otherwise, if the user has changed their base font size, things won't match.

site/src/components/persona.md
```diff
@@ -0,0 +1,156 @@
...
+		<figure class="persona-avatar">
+			<img src="~/src/scaffold/media/avatar.png" alt="username" />
+		</figure>
+	</div>
+</div>
+
+<div class="display-flex flex-wrap-wrap gap-xxs margin-top-md">
+	<div class="persona persona-xs">
+		<div class="persona-avatar">
+			<span>ww</span>
```
BenDMyers, line 120: If we're going to support styled-text avatars, we should ensure we demo accessible markup for these:

Could you please add `role=img` and an `aria-label`?

css/src/components/persona.scss
ollips: I was about to remove the letters version but realized we have places in learn where it is used.  Specifically in facepile component, which is gonna be replaced with persona.

So I applied the attributes Ben suggested so both versioned are announced the same way by screen readers.

css/src/components/persona.scss
ollips: Updated

css/src/components/persona.scss
ollips: Removed, I think I picked it up from the image and after looking around - we don't really need that much flexibility.

css/src/components/persona.scss
ollips: Updated

css/src/components/persona.scss
ollips: Updated

css/src/components/persona.scss
ollips: Updated

css/src/components/persona.scss
ollips: Yes, I think so. Won't it be an accessibility bug if the link doesn't stand out visually?

site/src/components/persona.md
ollips: Left only 5 sizes, per the conversation in Figma we can't cut to 3. And removed the responsive ones

css/src/components/persona.scss
ollips: Updated

site/src/components/persona.md
ollips: Updated

css/src/components/persona.scss
wibjorn: I don't think we need to support that old stuff. It should be relatively trivial to produce an SVG with initials anyway.

css/src/components/persona.scss
```diff
@@ -0,0 +1,82 @@
...
+			height: auto;
+			max-height: 100%;
+			border-radius: $persona-avatar-border-radius;
+		}
+	}
+
+	.persona-details {
+		align-self: center;
+		color: $persona-metadata-font-color;
+		font-size: $persona-metadata-font-size;
```
wibjorn, line 50: Shouldn't font-size be based on em here?

css/src/components/persona.scss
```diff
@@ -0,0 +1,82 @@
+$persona-font-size-xs: $font-size-9 !default;
```
wibjorn, line 1: I do not think we need a 12px avatar.

css/src/components/persona.scss
```diff
@@ -0,0 +1,82 @@
...
+$persona-font-size-sm: $font-size-8 !default;
+$persona-font-size-md: $font-size-7 !default;
+$persona-font-size-lg: $font-size-6 !default;
```
wibjorn, line 4: The difference between large and extra large is 2 pixels. Let's get rid of xl.

css/src/components/persona.scss
```diff
@@ -0,0 +1,82 @@
...
+	.persona-avatar {
+		display: flex;
+		align-items: center;
+		justify-content: center;
+		width: 2.35em;
+		height: 2.35em;
+		border-radius: $persona-avatar-border-radius;
+		background-color: $persona-avatar-background-color;
+		color: $persona-avatar-font-color;
+		font-size: $persona-font-size-md;
```
wibjorn, line 33: Reminder to once again go through and use em sizing with font-size applied at the component level.

site/src/components/persona.md
```diff
@@ -0,0 +1,136 @@
...
+template: standard
+figmaEmbed: https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FuVA2amRR71yJZ0GS6RI6zL%2F%25F0%259F%258C%259E-Atlas-Design-Library%3Ftype%3Ddesign%26node-id%3D1284%253A2163%26mode%3Ddesign%26t%3DklysUJ7ALWgcF1SQ-1 allowfullscreen
+classType: Component
+classPrefixes:
+  - persona
+---
+
+# Persona
+
+Persona component comes in handy for user's representation. It consists of two parts - user's avatar and details. Details has their name and optional metadata.
```
wibjorn, line 13:
```suggestion
We use the persona to represent users. It consists of two parts - user's avatar and details. Both subcomponents can be used independent of one another. Details has their name and optional metadata.
```

css/src/components/persona.scss
ollips: Sounds good, I will remove the initials version completely then

css/src/components/persona.scss
ollips: I finally got what the confusion was - the font variation I made was affecting the initials only to fit nicely in whatever avatar size is, and the description always had the same size. That's why I didn't use `em` here.

css/src/components/persona.scss
ollips: @wibjorn  So the avatar sizes we  use on Learn are `32`, `36`, and `48` with corresponding font sizes `14px`, `16px` and `20px`:

image

But per Figma's conversation, you were already asking to shrink it down to 3, but Courtney's feedback was to stick to 5, so that's why I made these - `28/12px`, `32/14px`, `36/16px`, `42/18px`, and `48/20px`, with 12px being the smallest to pass the accessibility reqs

image

Let me know if you feel like cutting those we don't need at the moment is okay and I'll do it.

Also, worth mentioning, in Figma avatar sizes are `16`, `24`, `28`, `32`, and `40` with the first two being tiny and initials won't be readable, so I cut those and added a few more to satisfy the request of having 5 sizes and included `36` and `48` we need.

PS I'll find a designer to update Figma to match what we'll end up with.

css/src/components/persona.scss
ollips: Please read the thread above about sizes.

css/src/components/persona.scss
```diff
@@ -0,0 +1,82 @@
...
+
+			&:not(a) {
+				color: $persona-name-font-color;
+			}
+		}
+	}
+
+	// Sizes
+
+	&.persona-xs .persona-avatar {
```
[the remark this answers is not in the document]
ollips: The reason it's done like that is that `persona-<size>` affects the avatar's size only, the description is fixed.

css/src/components/persona.scss
ollips: I can make description sizes change as well, will look like this:

image

ollips: After thinking about it a little more, for correct relative calculation the difference between sizes should be the same. So we should either proceed with keeping the persona's details part (sorry, not description as I said before) font the same for all sizes or update the `36` -> `40`. It's a small change and used only in one spot.

css/src/components/persona.scss
ollips: The initials version has been removed.

css/src/components/persona.scss
ollips: Updated again, since we have the `32` option used everywhere but global nav, I made its sizing perfect (avatar 32px, name 16px, and meta 12px), and the other two variants scale keeping the same proportions though a little different from the initial design: medium - avatar 37.8px, name 18.9px and meta 14px, large - avatar 43.19px, name 21.6 and meta 16px 😅 

image

css/src/components/persona.scss
ollips: Updated to 3 sizes.

css/src/components/persona.scss
ollips: Updated

css/src/components/persona.scss
ollips: Updated

css/src/components/persona.scss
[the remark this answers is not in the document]
ollips: Updated so the whole component scales.

site/src/components/persona.md
```diff
@@ -0,0 +1,100 @@
...
+template: standard
+figmaEmbed: https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FuVA2amRR71yJZ0GS6RI6zL%2F%25F0%259F%258C%259E-Atlas-Design-Library%3Ftype%3Ddesign%26node-id%3D1284%253A2163%26mode%3Ddesign%26t%3DklysUJ7ALWgcF1SQ-1 allowfullscreen
+classType: Component
+classPrefixes:
+  - persona
+---
+
+# Persona
+
+We use the persona to represent users. It consists of two parts - user's avatar and details. Both subcomponents can be used independent of one another. Details has their name and optional metadata.
```
wibjorn, line 13:
```suggestion
We use the persona to represent users. It consists of two parts - a user's avatar and details. Both subcomponents can be used independent of one another. Details has their name and optional metadata.
```

site/src/components/persona.md
```diff
@@ -0,0 +1,100 @@
...
+		<img src="~/src/scaffold/media/avatar.png" alt="Avatar for Jane Doe" />
+	</figure>
+	<div class="persona-details">
+		<a class="persona-name justify-self-stretch stretched-link" href="#">Jane Doe</a>
+		<p>Software Engineer</p>
+	</div>
+</div>
+```
+
+Persona might have both avatar and details parts presented at the same time or just one of them.
```
wibjorn, line 49:
```suggestion
Persona might have both avatar and details presented at the same time or just one of them.
```

ollips: Updated

site/src/components/persona.md
ollips: Updated

css/src/components/persona.scss
```diff
@@ -0,0 +1,65 @@
...
+
+.persona {
+	display: flex;
+	position: relative;
+	gap: $persona-gap-size;
+	font-size: $persona-font-size-md;
+
+	.persona-avatar {
+		display: flex;
+		flex-shrink: 0;
```
wibjorn, line 23: Can we remove `flex-shrink:0`, `display: flex`, `align-items: center`, `align-items: center`, and `line-height: 1;`? I'm not seeing that they have any noticeable effect

css/src/components/persona.scss
```diff
@@ -0,0 +1,65 @@
...
+		width: 2.6666em;
+		height: 2.6666em;
+		border-radius: $persona-avatar-border-radius;
+		background-color: $persona-avatar-background-color;
+		color: $persona-avatar-font-color;
+		line-height: 1;
+
+		img,
+		svg {
+			display: block;
```
wibjorn, line 35: Can this be reduced to: 

```
    border-radius:  $persona-avatar-border-radius;
    width: 100%;
    aspect-ratio: 1 / 1;
```
?

css/src/components/persona.scss
```diff
@@ -0,0 +1,65 @@
...
+			width: 100%;
+			height: auto;
+			max-height: 100%;
+			border-radius: $persona-avatar-border-radius;
+		}
+	}
+
+	.persona-details {
+		align-self: center;
```
wibjorn, line 44: can `align-self: center;` be removed?

wibjorn (changes requested):

css/src/components/persona.scss
ollips: Nope, it helps align the name nicely when there is no metadata:

image

Otherwise would be this:

image

css/src/components/persona.scss
ollips: `flex-shrink:0` needed so the avatar never shrinks, tested on small size:

image

And it actually fixes the issue other classes were helping with, demo below, so I'll remove them:

image

BenDMyers (approved):

css/src/components/persona.scss
ollips: Small addition - those classes were also needed to position initials perfectly in the center. 

Anyways, updated!

css/src/components/persona.scss
ollips: Updated

css/src/components/persona.scss
wibjorn: Huh! I missed that in my testing. Thanks for explaining.

wibjorn (approved):

BenDMyers (approved):

wibjorn (approved):

BenDMyers (approved):

PR #598, opened 2024-01-31
Set hero-image direction to avoid rtl flip

homing1:
Task: task-922683

Link: [preview-598](https://design.learn.microsoft.com/pulls/598/components/hero.html)

https://developer.mozilla.org/en-US/docs/Web/CSS/:dir

JsFiddle demonstrating how the `dir(rtl)` should work: https://jsfiddle.net/ua8ksr1z/

This PR sets the .hero-image direction as `ltr` to prevent the flipping the hero image on RTL pages. The hero content still respects the user's text direction.

Note: In the JSFiddle environment above, the matched rule is `div dir:rtl`, while the matched rule in Atlas and elsewhere is `:lang()` instead. 

image

image

## Testing

1. Visit https://design.learn.microsoft.com/pulls/598/components/hero.html . Scroll down to the hero image section. The image should be on the left side and the content text should be on the right side. 
2. Inspect the page with DevTools. For the top level `html` element, change the `lang` attribute to `ar-sa` and add `dir="rtl"` attribute. The hero content should remain on the left and the hero image should remain on the right.

## Contributor checklist

### DIFF
--- a/css/src/components/hero.scss
+++ b/css/src/components/hero.scss
@@ -100,6 +100,12 @@ $hero-image-width-widescreen: calc((100% - 2 * $layout-widescreen-gap) * 0.55) !
 		}
 	}
 
+	&.hero-image {
+		// Set direction to prevent image flip for rtl
+
+		direction: ltr;
+	}
+
 	&.hero-image::before {
 		display: none;
 		position: absolute;
--- a/css/src/components/hero.scss
+++ b/css/src/components/hero.scss
@@ -30,15 +30,12 @@ $hero-image-width-widescreen: calc((100% - 2 * $layout-widescreen-gap) * 0.55) !
 			width: $hero-content-fluid-width;
 			min-height: $hero-min-height-default;
 			padding-block: $hero-padding-default;
-			// not using logical properties here to avoid spacing issues for RTL
-			padding-right: $hero-padding-sm; /* stylelint-disable-line csstools/use-logical */
+			padding-inline-end: $hero-padding-sm;
 		}
 	}
 
 	&:not(.hero-image) {
 		.hero-content {
-			margin: auto;
-
 			@include desktop {
 				width: 100%;
 				max-width: $hero-content-width;
@@ -148,6 +145,11 @@ $hero-image-width-widescreen: calc((100% - 2 * $layout-widescreen-gap) * 0.55) !
 
 		.hero-content {
 			direction: rtl;
+
+			@include desktop {
+				padding-inline-end: unset;
+				padding-inline-start: $hero-padding-sm;
+			}
 		}
 	}
 }
--- a/css/src/components/hero.scss
+++ b/css/src/components/hero.scss
@@ -148,4 +148,12 @@ $hero-direction: ltr !default;
 			--hero-background-image: none;
 		}
 	}
+
+	&.hero-image:dir(rtl) {
+		direction: ltr;
+
+		.hero-content {
+			direction: rtl;
+		}
+	}
 }
--- a/css/src/components/hero.scss
+++ b/css/src/components/hero.scss
@@ -30,7 +30,8 @@ $hero-image-width-widescreen: calc((100% - 2 * $layout-widescreen-gap) * 0.55) !
 			width: $hero-content-fluid-width;
 			min-height: $hero-min-height-default;
 			padding-block: $hero-padding-default;
-			padding-inline-end: $hero-padding-sm;
+			// not using logical properties here to avoid spacing issues for RTL
+			padding-right: $hero-padding-sm; /* stylelint-disable-line csstools/use-logical */
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


BenDMyers (changes requested):
It's hard to tell if this is just because of the English content on this page, but I think that even if RTL text is shown in the left half, it should still be right-aligned:

## Expected:

Hero in RTL context, with text in the left half but still right-aligned

## Actual result:

Hero in RTL context, with the text on the left half, left-aligned

css/src/components/hero.scss
```diff
@@ -100,6 +101,16 @@ $hero-image-width-widescreen: calc((100% - 2 * $layout-widescreen-gap) * 0.55) !
 		}
 	}
 
+	&.hero-image {
+		// Set direction to prevent image flip for rtl
+
+		direction: $hero-direction;
+
+		.hero-content {
+			direction: $user-text-direction;
```
wibjorn, line 110: Agreed with Ben that we should continue to keep the text direction in line rtl if that's the preferred reading direction. Is the answer here to modify the hero image and any other relevant containers that are absolutely positioned to not use logical properties? If so, let's be sure we add comments next to them explaining why

wibjorn:
```css
	.hero-content {
		z-index: $zindex-active;
		padding-block: $hero-padding-sm;
		@include desktop {
			width: $hero-content-fluid-width;
			min-height: $hero-min-height-default;
			padding-block: $hero-padding-default;
			padding-inline-end: $hero-padding-sm; // padding-right
		}
	}
```
	and 
	
```css
		&:not(.hero-image) {
		.hero-content {
   			margin-right: auto; // addition

			@include desktop {
				width: 100%;
				max-width: $hero-content-width;
			}
		}
	}
```
as well as

```css
	&.hero-image::before {
		display: none;
		position: absolute;
		inset-block-start: 0;
		width: $hero-image-width;
		height: 100%;
		background-image: var(--hero-background-image);
		background-repeat: no-repeat;
		background-size: cover;
		content: ' ';
		right: $layout-gap; // here
```

homing1: I think it's more about DOM order rather than the logical properties. 

Example of replacing the logical properties: https://design.learn.microsoft.com/pulls/599/components/hero.html

It would require adding `position: relative` and `right: some %` to the hero content container, but only for rtl. The image position is also affected by `.gradient-border-right` so it would also require an override of its `border-inline-end-width` property.  

It seems more straightforward to set the correct direction property. I updated the PR to set `hero-content`'s direction to `$user-text-direction` to respect the correct text direction. Its value doesn't change within Atlas so there's no visual change for the Atlas site, but it is set to `rtl` in the rtl stylesheet in Docs-UI.

wibjorn: `$user-text-direction` is not something we can rely on the in the future. One goal we should have is to remove that property entirely.

css/src/components/hero.scss
```diff
@@ -30,12 +30,15 @@ $hero-image-width-widescreen: calc((100% - 2 * $layout-widescreen-gap) * 0.55) !
 			width: $hero-content-fluid-width;
 			min-height: $hero-min-height-default;
 			padding-block: $hero-padding-default;
-			padding-inline-end: $hero-padding-sm;
+			// not using logical properties here to avoid spacing issues for RTL
+			padding-right: $hero-padding-sm; /* stylelint-disable-line csstools/use-logical */
```
ollips, line 34: This works for sure, but I'm wondering if you considered instead of removing logical prop from here, adding these to line 150?

```
.hero-content {
	direction: rtl;
	padding-inline-end: 0;
	padding-inline-start: $hero-padding-sm;
}

```

image

css/src/components/hero.scss
```diff
@@ -30,12 +30,15 @@ $hero-image-width-widescreen: calc((100% - 2 * $layout-widescreen-gap) * 0.55) !
...
 		}
 	}
 
 	&:not(.hero-image) {
 		.hero-content {
+			margin: auto;
```
ollips, line 40: Can you explain the reasoning behind this update please? Looks like it changes the positioning of content:

image

ollips: Aka this:

image

homing1: It was originally suggested in an earlier comment. I removed it since there isn't any extra left spacing for LTR.

css/src/components/hero.scss
homing1: Yes, I also considered the suggested change above. Both would work. I updated to the above, but wrapped it in the desktop media query to match.

ollips (dismissed):

bbalentine87 (dismissed):

BenDMyers (approved):

css/src/components/hero.scss
wibjorn: Your solution looks great to me, seems to work as intended in a downstream package. Tested with npm link.

site/src/components/hero.md
```diff
@@ -98,6 +98,8 @@ Hero images are hidden in forced colors mode.
 
 **Images are currently not intended to scale or shrink. They are not shown on smaller or zoomed screens. For this reason, it's required that these images are presentational only. They should not contain words, subjects that might look naff if cut off, or anything essential to understanding the material of the page.**
 
+On RTL, the position of the hero image will remain on the right. The text within the hero content will be right aligned.
```
wibjorn, line 101: 1. Acronym, write out RTL
2. Describe why the hero image doesn't switch sides

wibjorn (approved):

BenDMyers (approved):

PR #620, opened 2024-04-29
Hero Patch for flex-direction vis diff items.

AaronAstle:
Task: task-953272

## Testing

1. Pull and Run the branch
2. Open the heros and inspect each hero.
3. Expect the hero WITH details to have the flex direction set to row.  Otherwise heros will not change their flex direction.
image

image

## Additional information

[Optional]

## Contributor checklist

### DIFF
--- a/.changeset/early-experts-wave.md
+++ b/.changeset/early-experts-wave.md
@@ -1,0 +1,5 @@
+---
+'@microsoft/atlas-site': patch
+---
+
+Cleaning up the hero markup. Removing an unneeded class and adding an atomic for the hero-details use case.
--- a/.changeset/early-experts-wave.md
+++ b/.changeset/early-experts-wave.md
@@ -2,4 +2,4 @@
 '@microsoft/atlas-site': patch
 ---
 
-Updated hero markup. Removing an unneeded class and adding an atomic for the hero-details use case.
+Updated markup for hero-details. Removing an unneeded class and adding an atomic for the hero-details use case.
--- a/.changeset/early-experts-wave.md
+++ b/.changeset/early-experts-wave.md
@@ -2,4 +2,4 @@
 '@microsoft/atlas-site': patch
 ---
 
-Cleaning up the hero markup. Removing an unneeded class and adding an atomic for the hero-details use case.
+Updated hero markup. Removing an unneeded class and adding an atomic for the hero-details use case.
--- a/.changeset/new-pets-speak.md
+++ b/.changeset/new-pets-speak.md
@@ -2,4 +2,4 @@
 '@microsoft/atlas-css': patch
 ---
 
-Updating the hero specific css to accomodate flex-direction prop when `.hero-details` is present. Otherwise the hero will not change flex-direction.
+Removing the flex-direction from the hero component and will apply the direction manually via templates and scripts
--- a/.changeset/new-pets-speak.md
+++ b/.changeset/new-pets-speak.md
@@ -2,4 +2,4 @@
 '@microsoft/atlas-css': patch
 ---
 
-Remove explicit flex-direction from the hero component in favor of atomics.
+Removed explicit flex-direction from the hero component in favor of atomics.
--- a/.changeset/new-pets-speak.md
+++ b/.changeset/new-pets-speak.md
@@ -2,4 +2,4 @@
 '@microsoft/atlas-css': patch
 ---
 
-Removing the flex-direction from the hero component and will apply the direction manually via templates and scripts
+Remove explicit flex-direction from the hero component in favor of atomics.
--- a/.changeset/new-pets-speak.md
+++ b/.changeset/new-pets-speak.md
@@ -1,0 +1,5 @@
+---
+'@microsoft/atlas-css': patch
+---
+
+Updating the hero specific css to accomodate flex-direction prop when `.hero-details` is present. Otherwise the hero will not change flex-direction.
--- a/css/src/components/hero.scss
+++ b/css/src/components/hero.scss
@@ -25,8 +25,10 @@ $hero-image-width-widescreen: calc((100% - 2 * $layout-widescreen-gap) * 0.55) !
 	align-items: stretch;
 	justify-content: space-between;
 
-	@include tablet {
-		flex-direction: row;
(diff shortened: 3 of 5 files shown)


css/src/components/hero.scss
```diff
@@ -25,8 +25,10 @@ $hero-image-width-widescreen: calc((100% - 2 * $layout-widescreen-gap) * 0.55) !
 	align-items: stretch;
 	justify-content: space-between;
 
-	@include tablet {
-		flex-direction: row;
+	&:has(.hero-details) {
```
BenDMyers, line 28: This would mark the first use of `:has` in Atlas (which I personally find exciting)

It's worth calling out that [support for `:has`](https://caniuse.com/css-has) is at ~92% adoption right now. That's pretty good - but it maybe could be better for something as prominent as a hero?

I wanted to volunteer a possible change, to see what we think of it:
1. Remove this `flex-direction: row` declaration altogether.
2. ...and then use our templates and scripts to apply the [`flex-direction-row-tablet` atomic](https://design.learn.microsoft.com/atomics/flex.html) to heroes with detail cards.

Thoughts?

sahiljohari: I, on the other hand, like Aaron's approach here. Let the CSS handle styling the hero elements instead of our templates/scripts. Thoughts?

ollips: what about just moving this rule to line 55?

BenDMyers: @ollips I don't think that would work, since that would apply to `.hero-details`, which is a child of the flex container.

.changeset/new-pets-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-css': patch
+---
+
+Removing the flex-direction from the hero component and will apply the direction manually via templates and scripts
```
BenDMyers, line 5: Given the function of this message here is to populate our changelog, what would you say to something like this?

```suggestion
Remove explicit flex-direction from the hero component in favor of atomics.
```

css/src/components/hero.scss
sahiljohari: I see it now! Yes, this makes sense.

.changeset/new-pets-speak.md
AaronAstle: should be updated now.

.changeset/early-experts-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-site': patch
+---
+
+Cleaning up the hero markup. Removing an unneeded class and adding an atomic for the hero-details use case.
```
homing1, line 5: What about "Updated markup for hero-details"?

AaronAstle: Done

.changeset/new-pets-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-css': patch
+---
+
+Remove explicit flex-direction from the hero component in favor of atomics.
```
homing1, line 5: Let's use past tense here too.

site/src/patterns/hero.md
```diff
@@ -53,11 +53,11 @@ Hero with a theme-aware companion image.
 
 ## Accent hero with details
 
-This component variant, known as 'Accent Hero with Details', overlays an element on top of the hero's image.
+This component variant, known as 'Accent Hero with Details', overlays an element on top of the hero's image. `.hero-details` will require the use of `.flex-direction-row-tablet` to be applied to the `.hero` markup.
```
BenDMyers, line 56: I might suggest a rewording here, for clarity, since at first blush I thought this meant the `.flex-direction-row-tablet` needed to be applied to `.hero-details`.

```suggestion
This component variant, known as 'Accent Hero with Details', overlays an element on top of the hero's image. Note that to place the details on the image correctly across screen sizes, the hero component should also have the `.flex-direction-row-tablet` atomic.
```

site/src/components/hero.md
```diff
@@ -116,11 +116,11 @@ On pages with right to left direction, the position of the hero image will remai
 
 Hero images with details
 
-This component variant overlays a container on the opposite side of the hero content. It can be used to accommodate more content within the hero section if necessary.
+This component variant overlays a container on the opposite side of the hero content. It can be used to accommodate more content within the hero section if necessary. `.hero-details` will require the use of `.flex-direction-row-tablet` to be applied to the `.hero` markup.
```
BenDMyers, line 119: Similar feedback as the patterns/hero.md change.

site/src/components/hero.md
```diff
@@ -116,11 +116,11 @@ On pages with right to left direction, the position of the hero image will remai
...
 
-This component variant overlays a container on the opposite side of the hero content. It can be used to accommodate more content within the hero section if necessary.
+This component variant overlays a container on the opposite side of the hero content. It can be used to accommodate more content within the hero section if necessary.  Note that to lay out the hero's children correctly across screen sizes, the hero component should also have the `.flex-direction-row-tablet` atomic.
 
 ```html-no-indent
 <section
-    class="hero hero-image hero-with-details border background-color-body-accent gradient-border-right gradient-border-body-accent"
+    class="hero hero-image flex-direction-row-tablet border background-color-body-accent gradient-border-right gradient-border-body-accent"
 	style="
         --hero-background-image-light: url('https://learn.microsoft.com/en-us/media/learn/plans/skilling_plan_hero.png?branch=main');
```
ollips, line 125: It's not related to current PRs changes, but something I noticed and wanted to point out - images used on hero and patterns/hero pages are not following the size requirements. they should be 1000x400px, but instead they're 1688x800px. Please update at your earliest convenience so we don't carry misleading examples. Feel free to resolve after reading ;)

ollips (approved):

site/src/components/hero.md
homing1: Should we include the required dimensions on this page then?

AaronAstle: For our development and design team, I think we should explicitly add some information about background image dimensions and composition requirements.

BenDMyers (approved):

PR #640, opened 2024-07-12
Form Behavior Labels - Radios

AaronAstle:
Task: task-[work-item-number]

## Testing

1. Pull and Run branch 
2. Open the Form Page
3. Without selecting a radio in the form behavior section, submit the form.
4. Expect the form errors to now represent aRadio input error label.  Expect that the error is representing the parent `.field .field-label` text OR to the first radios `aria-label` text.

## Additional information

[Optional]

## Contributor checklist

- [x] Does your pull request implement complex UI logic (js/ts)? Consider adding an integration test to test your user flow.

### DIFF
--- a/.changeset/eight-turkeys-invent.md
+++ b/.changeset/eight-turkeys-invent.md
@@ -1,0 +1,6 @@
+---
+'@microsoft/atlas-site': patch
+'@microsoft/atlas-js': patch
+---
+
+Minor update to the form behavior scripts that improves error messaging on radio input
--- a/.changeset/eight-turkeys-invent.md
+++ b/.changeset/eight-turkeys-invent.md
@@ -1,6 +1,6 @@
 ---
-'@microsoft/atlas-site': patch
-'@microsoft/atlas-js': patch
+'@microsoft/atlas-site': minor
+'@microsoft/atlas-js': minor
 ---
 
-Minor update to the form behavior scripts that improves error messaging on radio input
+Added form behavior radio group labeling and error handling.
--- a/integration/tests/form-behavior.spec.ts
+++ b/integration/tests/form-behavior.spec.ts
@@ -81,9 +81,21 @@ test.describe('form behavior validation', () => {
 		await submitBtn.click();
 
 		expect(errorContainer).toContainText('Please fix the following issues to continue:');
-		expect(errorContainer).toContainText([
-			`At least one selection from ${groupLabel} is required.`
-		]);
+		expect(errorContainer).toContainText([`A selection for "${groupLabel}" is required.`]);
+	});
+
+	test('show inputGroupRequired message when checkbox group is missing a selection', async ({
+		page,
+		errorContainer,
+		submitBtn
+	}) => {
+		const input = page.locator('#sample-multi-checkbox-2');
+		const groupLabel = await input.getAttribute('aria-label');
+
+		await submitBtn.click();
+
+		expect(errorContainer).toContainText('Please fix the following issues to continue:');
+		expect(errorContainer).toContainText([`A selection for "${groupLabel}" is required.`]);
 	});
 
 	test('show thereAreNoEditsToSubmit message when no edits are made on the form', async ({
--- a/integration/tests/form-behavior.spec.ts
+++ b/integration/tests/form-behavior.spec.ts
@@ -70,18 +70,54 @@ test.describe('form behavior validation', () => {
 		expect(errorContainer).toContainText([`${label} is required.`]);
 	});
 
-	test('show inputGroupRequired message when radio group is missing a selection', async ({
+	test('show required message with label from legend when radio group is missing a selection', async ({
 		page,
 		errorContainer,
 		submitBtn
 	}) => {
-		const input = page.locator('#question-id-1');
-		const groupLabel = await input.getAttribute('aria-label');
+		const input = await page.locator('#question-id-1');
+		const inputField = page.locator('fieldset', { has: input });
+		const groupLabel = inputField.locator('legend');
+		const label = (await groupLabel.textContent())?.trim();
 
 		await submitBtn.click();
 
 		expect(errorContainer).toContainText('Please fix the following issues to continue:');
-		expect(errorContainer).toContainText([`A selection for "${groupLabel}" is required.`]);
+		expect(errorContainer).toContainText([`A selection for "${label}" is required.`]);
+	});
+
+	test('show required message with label from aria-label when radio group is missing a selection', async ({
+		page,
+		errorContainer,
+		submitBtn
+	}) => {
+		const input = await page.locator('#question-id-1');
+		const inputField = page.locator('fieldset', { has: input });
+
+		// remove legend from inputField
+		const inputLabel = inputField.locator('legend');
+		inputLabel.evaluate(el => el.remove());
+		const label = await input.getAttribute('aria-label');
+
+		await submitBtn.click();
+
+		expect(errorContainer).toContainText('Please fix the following issues to continue:');
+		expect(errorContainer).toContainText([`A selection for "${label}" is required.`]);
+	});
+
+	// Checkbox
+	test('show inputGroupRequired field label message when checkbox group is missing a aria-label', async ({
+		page,
+		errorContainer,
+		submitBtn
+	}) => {
+		const inputLabel = page.locator('label', { has: page.locator('#sample-checkbox') });
+		const label = (await inputLabel.textContent())?.trim();
+
+		await submitBtn.click();
+
+		expect(errorContainer).toContainText('Please fix the following issues to continue:');
+		expect(errorContainer).toContainText([`A selection for "${label}" is required.`]);
 	});
 
 	test('show inputGroupRequired message when checkbox group is missing a selection', async ({
--- a/integration/tests/form-behavior.spec.ts
+++ b/integration/tests/form-behavior.spec.ts
@@ -70,6 +70,22 @@ test.describe('form behavior validation', () => {
 		expect(errorContainer).toContainText([`${label} is required.`]);
 	});
 
+	test('show inputGroupRequired message when radio group is missing a selection', async ({
+		page,
+		errorContainer,
+		submitBtn
+	}) => {
+		const input = page.locator('#question-id-1');
+		const groupLabel = await input.getAttribute('aria-label');
+
+		await submitBtn.click();
+
+		expect(errorContainer).toContainText('Please fix the following issues to continue:');
+		expect(errorContainer).toContainText([
+			`At least one selection from ${groupLabel} is required.`
+		]);
+	});
+
 	test('show thereAreNoEditsToSubmit message when no edits are made on the form', async ({
 		page,
 		errorContainer,
--- a/integration/tests/form-behavior.spec.ts
+++ b/integration/tests/form-behavior.spec.ts
@@ -117,7 +117,7 @@ test.describe('form behavior validation', () => {
 		await submitBtn.click();
 
 		expect(errorContainer).toContainText('Please fix the following issues to continue:');
-		expect(errorContainer).toContainText([`A selection for "${label}" is required.`]);
+		expect(errorContainer).toContainText([`${label} is required.`]);
 	});
 
 	test('show inputGroupRequired message when checkbox group is missing a selection', async ({
@@ -131,7 +131,7 @@ test.describe('form behavior validation', () => {
 		await submitBtn.click();
 
 		expect(errorContainer).toContainText('Please fix the following issues to continue:');
-		expect(errorContainer).toContainText([`A selection for "${groupLabel}" is required.`]);
+		expect(errorContainer).toContainText([`${groupLabel} is required.`]);
 	});
 
 	test('show thereAreNoEditsToSubmit message when no edits are made on the form', async ({
--- a/js/src/elements/form-behavior.ts
+++ b/js/src/elements/form-behavior.ts
@@ -27,7 +27,7 @@ export class FormBehaviorElement extends HTMLElement {
 
 	validators: Validator[] = [
 		this.validateMinLength.bind(this), // min length before required
-		this.validateGroupRequired.bind(this), // Checkbox or Radio groups
+		// this.validateGroupRequired.bind(this), // Checkbox or Radio groups
 		this.validateRequired.bind(this),
 		this.validateMaxLength.bind(this)
 	];
@@ -360,25 +360,10 @@ export class FormBehaviorElement extends HTMLElement {
 		if (input.validity.valueMissing) {
 			return this.locStrings.inputRequired.replace(
 				'{inputLabel}',
-				customElements.get(input.localName) ? `A selection for "${label}"` : label
+				customElements.get(input.localName) || input.type === 'radio' || input.type === 'checkbox'
+					? `A selection for "${label}"`
+					: label
 			);
-		}
-		return null;
-	}
-
-	validateGroupRequired(input: HTMLValueElement, label: string): string | null {
-		if (input.type === 'radio' || input.type === 'checkbox') {
-			const group = getField(input);
-			const inputs = Array.from(group.querySelectorAll<HTMLInputElement>('input'));
-			// Return null if a single checkbox, so that validateRequired can handle it
-			if (inputs.length === 1 && inputs[0].type === 'checkbox') {
-				return null;
-			}
-
-			const hasCheckedInput = inputs.some(input => input.checked);
-			if (!hasCheckedInput) {
-				return this.locStrings.inputGroupRequired.replace('{inputGroup}', label);
-			}
 		}
 		return null;
 	}
@@ -544,7 +529,6 @@ export class FormBehaviorElement extends HTMLElement {
 			return;
 		}
 
-		// TODO make label aware of radio and where label can live.
 		const label = getLabel(input);
 		const group = getField(input);
 
--- a/js/src/elements/form-behavior.ts
+++ b/js/src/elements/form-behavior.ts
@@ -2,6 +2,7 @@ import { generateElementId, kebabToCamelCase } from '../utilities/util';
 
 export const defaultMessageStrings = {
 	contentHasChanged: 'Content has changed, please reload the page to get the latest changes.',
+	inputGroupRequired: 'At least one selection for {inputLabel} is required.',
 	inputMaxLength: '{inputLabel} cannot be longer than {maxLength} characters.',
 	inputMinLength: '{inputLabel} must be at least {minLength} characters.',
 	inputRequired: '{inputLabel} is required.',
@@ -27,6 +28,7 @@ export class FormBehaviorElement extends HTMLElement {
 	validators: Validator[] = [
 		this.validateMinLength.bind(this), // min length before required
 		this.validateRequired.bind(this),
+		this.validateGroupRequired.bind(this), // Checkbox or Radio groups
 		this.validateMaxLength.bind(this)
 	];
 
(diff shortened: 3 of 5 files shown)


js/src/elements/form-behavior.ts
```diff
@@ -676,7 +684,9 @@ function setBusySubmitButton(event: Event, form: HTMLFormElement, isLoading: boo
 
 export function getLabel(input: HTMLValueElement): string {
 	const label =
-		input.labels && input.labels.length
+		input.type === 'radio'
+			? input.getAttribute('data-error-label')
+			: input.labels && input.labels.length
 			? input.labels[0].textContent
 			: input.getAttribute('aria-label');
```
BenDMyers, line 691: Personal preference here, I think it'd be good to try to avoid nesting ternaries for readability's sake. I'll open this up to other Atlas contributors to weigh in on how they feel, but I might recommend restructuring this to something like:

```ts
let label = '';

if (input.type === 'radio') {
    label = input.getAttribute('data-error-label');
} else if (input.labels?.length) {
    label = input.labels[0].textContent;
} else {
    label = input.getAttribute('aria-label');
}
```

Thoughts?

AaronAstle: Yeah.  I agree.

I had it split out to be more readable but the linting was hating me so I went path of least resistance to start.  I will bake this in and let you know when its ready for review again.

AaronAstle: @BenDMyers  updated

js/src/elements/form-behavior.ts
```diff
@@ -675,10 +683,16 @@ function setBusySubmitButton(event: Event, form: HTMLFormElement, isLoading: boo
...
 
 export function getLabel(input: HTMLValueElement): string {
-	const label =
-		input.labels && input.labels.length
-			? input.labels[0].textContent
-			: input.getAttribute('aria-label');
+	let label: string | null = null;
+
+	if (input.type === 'radio') {
+		label = input.getAttribute('data-error-label');
```
sahiljohari, line 689: just curious - are we not supporting the retrieval of `aria-label` for radio inputs? Also, is the name `data-error-label` accurately representing the values that can be assigned to it? In your example below, it seems like you are wanting to assign a descriptive name to the input. If that's the case, shouldn't we just call it `data-validation-label` or something?

AaronAstle: Yeah validation-label sounds like a better name.

As for the aria label, @BenDMyers correct me if Im wrong here, I believe the aria-label is specific to the inputs themselves.  Not the Group.  So with the separate data-attr, we are separating the label to the group rather than the individual input the aria-label would be associated to.

AaronAstle: Updated label name

js/src/elements/form-behavior.ts
```diff
@@ -675,10 +683,16 @@ function setBusySubmitButton(event: Event, form: HTMLFormElement, isLoading: boo
 }
 
 export function getLabel(input: HTMLValueElement): string {
-	const label =
-		input.labels && input.labels.length
-			? input.labels[0].textContent
-			: input.getAttribute('aria-label');
+	let label: string | null = null;
+
+	if (input.type === 'radio') {
```
wibjorn, line 688: Is there a way we can get this to work without the introduction of a new `data-` specifically for `radio`?

js/src/elements/form-behavior.ts
```diff
@@ -421,6 +422,11 @@ export class FormBehaviorElement extends HTMLElement {
 				continue;
 			}
 
+			// For Radios where name is the same across options, only validate the first one
+			if (prevInputName !== undefined && input.name === prevInputName) {
```
wibjorn, line 426: Can we scope this check to radios?

js/src/elements/form-behavior.ts
```diff
@@ -395,6 +395,7 @@ export class FormBehaviorElement extends HTMLElement {
 	): Promise<FormValidationResult> {
 		const errors: FormValidationError[] = [];
 		const { errorAlert, errorList } = this.getErrorAlert(form);
+		let prevInputName: string = ''; // Used for Radios that share the same name
```
wibjorn, line 398: If this is only used for radios, it should be named something that indicates that.

js/src/elements/form-behavior.ts
```diff
@@ -434,7 +440,9 @@ export class FormBehaviorElement extends HTMLElement {
 			}
 
 			const isCustomElement = !!customElements.find(el => el === input);
+
 			this.runBasicValidation(input, displayValidity, errors, errorList, isCustomElement);
+			prevInputName = input.name;
```
wibjorn, line 445: ditto to other comments

site/src/components/form.md
```diff
@@ -216,6 +223,47 @@ If you want to skip the basic validation on an input, apply a `data-skip-validat
...
+					type="radio"
+					class="radio-dot"
+					value="Yes"
+					data-validation-label="Radio Input selection"
+					required
+				/>
+				<span class="radio-label-text">Yes</span>
+			</label>
+			<label class="radio" for="question-id-2">
+				<input name="question-1" id="question-id-2" type="radio" class="radio-dot" value="No" />
```
wibjorn, line 242: super nit: indentation to match other examples?

.changeset/eight-turkeys-invent.md
```diff
@@ -0,0 +1,6 @@
+---
+'@microsoft/atlas-site': patch
+'@microsoft/atlas-js': patch
```
wibjorn, line 3: Seems like new functionality `minor`?

js/src/elements/form-behavior.ts
wibjorn: After a little more thinking, it'd be idea if we could consider introducing the concept of radio groups or even checkbox groups. Something that says conceptual one input in this group is required. `loc-input-group-required="{inputGroupId} is required."`

.changeset/eight-turkeys-invent.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+Minor update to the form behavior scripts that improves error messaging on radio input
```
homing1, line 6: What do you think about something like "Added form behavior radio group labeling and error handling."?

package-lock.json
```diff
@@ -10,11 +10,8 @@
 			"license": "MIT",
```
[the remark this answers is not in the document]
AaronAstle: I think i pulled latest and ran install.  

Not intentionally checked in/added.  I can remove.

site/src/components/form.md
AaronAstle: Yep.  On save auto formatting caused the shift here.

js/src/elements/form-behavior.ts
```diff
@@ -364,6 +366,16 @@ export class FormBehaviorElement extends HTMLElement {
 		return null;
 	}
 
+	validateGroupRequired(input: HTMLValueElement, label: string): string | null {
+		const group = getField(input);
+		const inputs = Array.from(group.querySelectorAll<HTMLInputElement>('input'));
+		const selected = inputs.some(input => input.checked);
```
BenDMyers, line 372: Super nit, feel free to ignore:

The variable name here makes me think I'd get a reference to a checked/selected input. Instead, we're going for more of a boolean as to whether there is an active selection.

What about `hasCheckedInput`?

js/src/elements/form-behavior.ts
```diff
@@ -364,6 +366,16 @@ export class FormBehaviorElement extends HTMLElement {
...
+		if (!selected) {
+			return `${this.locStrings.inputGroupRequired.replace('{inputLabel}', label)}`;
```
BenDMyers, line 374: Super nit, feel free to ignore.

Template string shouldn't be required here.
```suggestion
			return this.locStrings.inputGroupRequired.replace('{inputLabel}', label);
```

js/src/elements/form-behavior.ts
```diff
@@ -364,6 +366,16 @@ export class FormBehaviorElement extends HTMLElement {
...
+		}
+		return null;
```
BenDMyers, line 376: Do we need this `return null` if we're already at the end of the function?

js/src/elements/form-behavior.ts
```diff
@@ -407,6 +419,13 @@ export class FormBehaviorElement extends HTMLElement {
 				continue;
 			}
 
+			if (input.type === 'radio') {
+				const radioGroup = form.querySelectorAll<HTMLInputElement>(`input[name="${input.name}"]`);
+				if (radioGroup.length > 1 && radioGroup[0] !== input) {
```
BenDMyers, line 424: Can you help me understand why we need to check that the given input isn't the first in its radiogroup?

AaronAstle: If I do not continue on the first of the group, the report list causes a line items for each radio input.  So instead of just one "radio group" error notification, it would be 1:1 on labels. Not ideal.

This way, when a radio is required, you mark the first with required and all of the same `name` get the applied required logic.

site/src/components/form.md
```diff
@@ -205,9 +205,16 @@ If you want to skip the basic validation on an input, apply a `data-skip-validat
...
+		<legend class="field-label">Checkbox Field label</legend>
 		<div class="field-body">
 			<label class="checkbox">
-				<input type="checkbox" name="checkbox-example" required id="sample-checkbox" />
+				<input
+					type="checkbox"
+					name="checkbox-example"
+					required
+					id="sample-checkbox"
+					data-error-label="Checkbox Example"
```
sahiljohari: rename this as well?

site/src/components/form.md
```diff
@@ -216,6 +223,47 @@ If you want to skip the basic validation on an input, apply a `data-skip-validat
...
+		<legend class="field-label">Radio Field label</legend>
+		<div class="field-body">
+			<label class="radio" for="question-id-1">
+				<input
+					name="question-1"
+					id="question-id-1"
+					type="radio"
+					class="radio-dot"
+					value="Yes"
+					data-error-label="Radio Input selection"
```
sahiljohari: and this?

js/src/elements/form-behavior.ts
AaronAstle: I believe the null allows for fall-through checks on the validators iteration.  Previous implementation pattern present with other validators too.

js/src/elements/form-behavior.ts
AaronAstle: Adjusted the other validation methods to remove the template string syntax

site/src/components/form.md
AaronAstle: removed the label from the checkbox.

site/src/components/form.md
AaronAstle: Removed.

js/src/elements/form-behavior.ts
BenDMyers: This makes sense to me - this might be a useful time for a code comment, something along the lines of:

```ts
// Apply validation to first radio button in group
```

js/src/elements/form-behavior.ts
AaronAstle: ✅

js/src/elements/form-behavior.ts
```diff
@@ -27,6 +28,7 @@ export class FormBehaviorElement extends HTMLElement {
 	validators: Validator[] = [
 		this.validateMinLength.bind(this), // min length before required
 		this.validateRequired.bind(this),
+		this.validateGroupRequired.bind(this), // Checkbox or Radio groups
```
wibjorn, line 31: Is this in the correct order? For radios, I do not see this running. It seems that for required radios, the validator on line 30 fails before we reach this.

wibjorn: As you can see this results in different labels between select and radio.

image

js/src/elements/form-behavior.ts
```diff
@@ -356,10 +358,20 @@ export class FormBehaviorElement extends HTMLElement {
...
 				customElements.get(input.localName) ? `A selection for "${label}"` : label
-			)}`;
+			);
+		}
+		return null;
+	}
+
+	validateGroupRequired(input: HTMLValueElement, label: string): string | null {
+		const group = getField(input);
+		const inputs = Array.from(group.querySelectorAll<HTMLInputElement>('input'));
```
wibjorn, line 371: Do we want to check _all_ types inputs are checked? or only certain types? Should we query `input[type="checkbox"],input[type="radio"]`?

js/src/elements/form-behavior.ts
```diff
@@ -2,6 +2,7 @@ import { generateElementId, kebabToCamelCase } from '../utilities/util';
 
 export const defaultMessageStrings = {
 	contentHasChanged: 'Content has changed, please reload the page to get the latest changes.',
+	inputGroupRequired: 'At least one selection for {inputGroup} is required.',
```
wibjorn, line 5: Is this the correct validation string for radio?

image

AaronAstle: That is the text of the `.field-label` for the radio.  So yes.

site/src/components/form.md
```diff
@@ -205,9 +205,16 @@ If you want to skip the basic validation on an input, apply a `data-skip-validat
 		</div>
 	</div>
 	<div class="field">
+		<legend class="field-label">Checkbox Field label</legend>
```
wibjorn: per mdn, legends exist in the context of form fieldset, do we need the markup to reflect that? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend
```
<fieldset>
  <legend>Choose your favorite monster</legend>
```

site/src/components/form.md
```diff
@@ -216,6 +223,47 @@ If you want to skip the basic validation on an input, apply a `data-skip-validat
 			</label>
 		</div>
 	</div>
+	<div class="field">
```
wibjorn: fieldset too?

js/src/elements/form-behavior.ts
wibjorn: bump

js/src/elements/form-behavior.ts
wibjorn: Looks better now

js/src/elements/form-behavior.ts
```diff
@@ -356,10 +358,27 @@ export class FormBehaviorElement extends HTMLElement {
...
+		}
+		return null;
+	}
+
+	validateGroupRequired(input: HTMLValueElement, label: string): string | null {
+		if (input.type === 'radio' || input.type === 'checkbox') {
+			const group = getField(input);
+			const inputs = Array.from(group.querySelectorAll<HTMLInputElement>('input'));
+			// Return null if a single checkbox, so that validateRequired can handle it
+			if (inputs.length === 1 && inputs[0].type === 'checkbox') {
```
wibjorn, line 374: I'm curious about the case where there is more than 1 checkbox, but neither are required? Can we still submit?

js/src/elements/form-behavior.ts
AaronAstle: This is outdated.  Is there more you want to change here?

integration/tests/form-behavior.spec.ts
```diff
@@ -70,6 +70,34 @@ test.describe('form behavior validation', () => {
 		expect(errorContainer).toContainText([`${label} is required.`]);
 	});
 
+	test('show inputGroupRequired message when radio group is missing a selection', async ({
```
BenDMyers, line 73: Can we add tests specifically around the fallback chain - i.e., that it uses the group label first?

js/src/elements/form-behavior.ts
```diff
@@ -26,6 +27,7 @@ export class FormBehaviorElement extends HTMLElement {
 
 	validators: Validator[] = [
 		this.validateMinLength.bind(this), // min length before required
+		// this.validateGroupRequired.bind(this), // Checkbox or Radio groups
```
BenDMyers, line 30: Should this be commented out? If we need to get rid of it, let's delete this line of code altogether.

js/src/elements/form-behavior.ts
AaronAstle: Whole method ended up being superfluous.  Removed.

integration/tests/form-behavior.spec.ts
AaronAstle: Adjusted.

js/src/elements/form-behavior.ts
AaronAstle: Removed

js/src/elements/form-behavior.ts
```diff
@@ -675,10 +706,22 @@ function setBusySubmitButton(event: Event, form: HTMLFormElement, isLoading: boo
...
 export function getLabel(input: HTMLValueElement): string {
-	const label =
-		input.labels && input.labels.length
-			? input.labels[0].textContent
-			: input.getAttribute('aria-label');
+	let label: string | null = null;
+
+	if (input.type === 'radio') {
+		const group = getField(input);
+		const groupLabel = group.querySelector('.field-label');
```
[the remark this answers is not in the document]
AaronAstle: I'd love the opportunity to be part of that conversation.

site/src/components/form.md
AaronAstle: adjusted.

site/src/components/form.md
AaronAstle: adjusted but kept class names

js/src/elements/form-behavior.ts
AaronAstle: removed method.  No longer needed.

js/src/elements/form-behavior.ts
```diff
@@ -519,10 +531,10 @@ export class FormBehaviorElement extends HTMLElement {
...
 		const group = getField(input);
 
-		if (displayValidity) {
-			setValidationMessage(input, '');
-			group.classList.remove('errored');
-		}
+		// if (displayValidity) {
+		// 	setValidationMessage(input, '');
+		// 	group.classList.remove('errored');
+		// }
```
AaronAstle, line 537: Commented out this block as it was resetting on the checkbox group error notes.  Lets test that is not actually needed.

wibjorn: I think we should try to get this one right (ensure it isn't needed and remove the comment, or determine it is needed and uncomment) before we commit this change

AaronAstle: I tested it several times and all seems to operate.  What I am asking is for each of you to do the same.

I do not intend to close this PR with it remaining.

AaronAstle: Removed as my testing was not affected by removing it.  (or I am missing something completely)

wibjorn (approved):

site/src/components/form.md
```diff
@@ -215,7 +270,48 @@ If you want to skip the basic validation on an input, apply a `data-skip-validat
 				</span>
 			</label>
 		</div>
-	</div>
+	</fieldset>
+
+	<fieldset class="field">
+		<legend class="field-label">Multiple Checkbox Field label</legend>
+		<div class="field-body">
+			<label class="checkbox">
```
wibjorn, line 278: An obligatory comment here, since this _seems to me at least_ to be validating that "a required checkbox is required" (always true anyway), not "one of these checkboxes in this group is required". As long as that is understood, and we know it won't break stuff, I think we're alright.

AaronAstle: Let me address this.  I think it is blanketed by the group label.  Ping you once that change is in.

wibjorn (approved):

sahiljohari (approved):

BenDMyers (approved):

PR #662, opened 2024-08-22
Adds Accordion component and patterns

jwmunn:
Task: task-[988155](https://dev.azure.com/ceapex/Engineering/_workitems/edit/988155)

Links: [Accordion component](http://localhost:1111/components/accordion.html) |  [Accordion patterns](http://localhost:1111/patterns/accordion.html)

**Component**
http://localhost:1111/components/accordion.html

Fluent 2 - [React Accordion - Fluent 2 Design System (microsoft.design)](https://fluent2.microsoft.design/components/web/react/accordion/usage)
Figma - [Fluent 2 web – Figma](https://www.figma.com/design/GvIcCw0tWaJVDSWD4f1OIW/Fluent-2-web?node-id=292-48147&t=oWRDkUAktt9miDFa-0)
Accessibility - [Accordion Pattern (Sections With Show/Hide Functionality) | APG | WAI | W3C](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/)

Handles:
* 3 sizes (matches fluent)
* Multiple open panels or only 1 open panel, designated open panel on page load
* Left and right aligned icon (matches fluent arrow direction)
~~* Automatically adds left/right padding if background color atomic classes are used~~
* Uses same animation for chevron direction change + timing as current _details.scss icon (transition: transform 0.15s)

**Pattern**
http://localhost:1111/patterns/accordion.html
Serves 2 functions
1. Use atomics to show pattern that will be used for upcoming Ignite designs.
2. This component will deprecate the need for _details.scss currently used in docs-ui, 2 patterns used to show the 2 scenarios currently being used on docs-ui.

Form-based advisor/Plan adjuster pattern for Ignite
Figmas (press Shift + 2 after page load to zoom in on component selection):
* [Form-based advisor - Plan generation view](https://www.figma.com/design/a5gqz0qtUMAKF36lNS6b61/Adaptive-Learning---Phase-I-(2024)?node-id=3090-25025&m=dev&focus-id=2826-46180)
* [Plan Adjuster UI](https://www.figma.com/design/a5gqz0qtUMAKF36lNS6b61/Adaptive-Learning---Phase-I-(2024)?node-id=2826-57549&m=dev&focus-id=2826-46180)

`_details.scss` usage in docs-ui
* 2nd example would replace Training journey section below first bifold
   *  https://learn.microsoft.com/en-us/training/career-paths/ai-engineer

3rd pattern example would replace:
* Marketing page FAQ pattern (right aligned chevron, body gray content container.)
  * https://review.learn.microsoft.com/en-us/test/marketing/?branch=main#frequently-asked-questions-title

## Testing

1. Visit preview pages to view new pages
    Expected result: Component page matches Figma. Patterns match docs-ui .details usage (FAQ, Career Path Journey section), and Form-based advisor/Plan adjuster designs proposed for Ignite
3. Visit [Select component page](http://localhost:1111/components/select.html) 
   Expected result: Verify refactor to use shared chevron mixin results in no visual changes from existing page

## Additional information

[Optional]

## Contributor checklist


css/src/components/accordion.scss
```diff
@@ -0,0 +1,157 @@
...
+}
+
+.accordion {
+	@include accordion-summary-sizes(
+		$accordion-font-size-md,
+		$accordion-spacing-md,
+		$accordion-spacing-offset-md
+	);
+
+	line-height: 1.5;
```
BenDMyers, line 87: This will override the line-height inside of the `.accordion-content`, instead of using our typography defaults. If this is meant for the summary, could we scope it to that directly?

site/src/components/accordion.md
```diff
@@ -0,0 +1,137 @@
...
+# need design to create
+# figmaEmbed: https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FuVA2amRR71yJZ0GS6RI6zL%2F%25F0%259F%258C%259E-Atlas-Design-Library%3Fnode-id%3D1488%253A35182
+classType: Component
+classPrefixes:
+  - accordion
+---
+
+# Accordion
+
+The Accordion component contains a disclosure widget, where information is only visible when the element is toggled to an "open" state. A Summary or label must be provided using the `<summary>` element.
```
ollips, line 14: Accordion is a component groups sections of related content that can be opened or closed. Accordion is built on top of the [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#customizing_the_disclosure_widget) HTML component.

site/src/components/accordion.md
```diff
@@ -0,0 +1,137 @@
+---
+title: Accordion
+description: The accordion component in the Atlas Design System
+template: standard
+# need design to create
```
ollips, line 5: Remove entirely figma section, I doubt atlas's figma will be updated any time soon

site/src/components/accordion.md
```diff
@@ -0,0 +1,137 @@
...
+  - accordion
+---
+
+# Accordion
+
+The Accordion component contains a disclosure widget, where information is only visible when the element is toggled to an "open" state. A Summary or label must be provided using the `<summary>` element.
+
+## Usage
+
+Here is an example of a standard `.accordion` usage. By default `<details>` elements are closed on page load and toggle open or closed when the user clicks the `<summary>`.
```
ollips, line 18: Here is an example of a standard `.accordion` usage. By default, accordion items are closed and allow multiple items to be open at once.

site/src/components/accordion.md
```diff
@@ -0,0 +1,137 @@
...
+		<p>Accordion panel 3</p>
+	</div>
+</details>
+```
+
+## Attributes
+
+### Open
+
+There is an option to have the accordion component load in the expanded state by adding the `open` attribute.
```
ollips, line 45: .. adding the `open` attribute to one of the items.

site/src/components/accordion.md
```diff
@@ -0,0 +1,137 @@
...
+	<summary>Accordion header 3</summary>
+	<div class="accordion-content">
+		<p>Accordion panel 3</p>
+	</div>
+</details>
+```
+
+### Name
+
+Using the `name` attribute, there is an option to group multiple accordion components, which allows only one open panel at a time.
```
ollips, line 70: Applying the `name` attribute to all accordion items with the same value will allow opening only one of the items at a time.

site/src/components/accordion.md
```diff
@@ -0,0 +1,137 @@
...
+	<summary>Accordion header 3</summary>
+	<div class="accordion-content">
+		<p>Accordion panel 3</p>
+	</div>
+</details>
+```
+
+## Icon location
+
+By default the expansion indicator icon is before the summary content. Using the `.accordion-icon-end` class, the icon can be positioned at the end of the summary. The icon is absolutely positioned to remain at the top if Summary content wraps or contains multiple lines.
```
ollips, line 95: By default, the expansion indicator icon located before the accordion item's header.  Applying the `.accordion-icon-end` class with change the icon's position to the end of the header.`

site/src/components/accordion.md
```diff
@@ -0,0 +1,137 @@
...
+</details>
+<details class="accordion">
+	<summary>Accordion header 3</summary>
+	<div class="accordion-content">
+		<p>Accordion panel 3</p>
+	</div>
+</details>
+```
+
+### Name
```
ollips, line 68: In Firefox `name` doesn't work =\ need to figure out a work around this

site/src/components/accordion.md
```diff
@@ -0,0 +1,137 @@
...
+  	<p>Accordion panel 2</p>
+  </div>
+</details>
+
+## Sizes
+
+`accordion-sm`/`accordion-lg` modifier classes make the text size and padding a little bigger or smaller than the default.
+
+```html
+<details class="accordion accordion-sm" name="accordion-group-1">
```
ollips, line 115: please remove `name` from all examples but where it's explained OR assign different name. I'd prefer removing but up to you

site/src/patterns/accordion.md
```diff
@@ -0,0 +1,163 @@
+---
+title: Accordion
+description: Accordion patterns in the Atlas Design System
+template: standard
+# figmaEmbed: https://www.figma.com/design/a5gqz0qtUMAKF36lNS6b61/Adaptive-Learning---Phase-I-(2024)?node-id=2581-106784&t=lkaa1re80XFTF9O3-0
+---
+
+# Accordion patterns
+
+Because the `details` component has so many possible permutations, it can be helpful to provide examples of certain styles.
```
ollips, line 10: I wouldn't refer to details component but accordion instead. 

Because of the accordion component ....

css/src/components/accordion.scss
jwmunn: Good call, updated ✅

site/src/components/accordion.md
jwmunn: Looks like they released a fix for this, it isn't in the latest 129.0.2 but will be included in v130. Not sure when that is scheduled to be released.
https://bugzilla.mozilla.org/show_bug.cgi?id=1856460
https://github.com/mdn/content/pull/35447

css/src/components/accordion.scss
```diff
@@ -0,0 +1,156 @@
+/* stylelint-disable selector-no-qualifying-type */
```
ollips, line 1: Regarding the sizing: after reviewing fluent's figma and sites, it turned out the only thing changes is the font size of the summary, and nothing else. And the content's size stays the same for all variations and is 14px (need to be updated here). So I'd say keep it simple and follow the same pattern

site/src/components/accordion.md
jwmunn: Found it, being release Sept 3 🎉
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/130

css/src/components/accordion.scss
```diff
@@ -0,0 +1,156 @@
...
+
+	.accordion-content {
+		padding-block: $padding-size;
+
+		&[class*='background-color-'] {
+			padding-inline: $padding-size;
+		}
+	}
+
+	&[class*='background-color-'] {
```
ollips, line 54: I feel like if there are scenarios where background color modified with atomics - the rest of the stuff should be modified with atomics as well. Remove?

css/src/components/accordion.scss
```diff
@@ -0,0 +1,156 @@
...
+			transform: $chevron-down-rotate translate(calc($chevron-arrow-size / 2), 0);
+		}
+
+		summary.accordion-icon-end::after {
+			transform: $chevron-up-rotate;
+		}
+	}
+}
+
+// Sizes
```
ollips, line 140: this should be nested within `.accordion`

css/src/components/accordion.scss
```diff
@@ -0,0 +1,156 @@
...
+		&:hover {
+			color: $primary-hover;
+
+			&::before,
+			&::after {
+				border-color: $primary-hover;
+			}
+		}
+
+		&:not(.accordion-icon-end) {
```
ollips, line 102: Padding on the right side is missed

image

site/src/components/accordion.md
```diff
@@ -0,0 +1,137 @@
...
+By default the expansion indicator icon is before the summary content. Using the `.accordion-icon-end` class, the icon can be positioned at the end of the summary. The icon is absolutely positioned to remain at the top if Summary content wraps or contains multiple lines.
+
+<details class="accordion">
+  <summary>Accordion header 1</summary>
+  <div class="accordion-content">
+  	<p>Accordion panel 1</p>
+  </div>
+</details>
+<details class="accordion border-top">
+  <summary class="accordion-icon-end">Accordion header 2</summary>
```
ollips, line 104: two things:
- can you wrap the whole thing into 'html``' so the markup example is visible. also, can remove the first (default) scenario and the `border-top` atomic
- can that class be added to the parent element so it's a little easier to read it and less things to remember where which thing should be assigned

site/src/components/accordion.md
ollips: Haha lucky you! Still need to keep an eye on it and verify it works soon. Otherwise, we have to figure out another way to make it work consistently throughout all the browsers

site/src/components/accordion.md
jwmunn: ✅Updated

site/src/components/accordion.md
jwmunn: ✅Updated in both md files

site/src/components/accordion.md
jwmunn: ✅Updated

site/src/components/accordion.md
jwmunn: ✅Updated

site/src/components/accordion.md
jwmunn: ✅Updated

site/src/components/accordion.md
jwmunn: ✅I agree, updated

site/src/patterns/accordion.md
jwmunn: ✅Changed it to be more in line with the card component's reference to the card patterns page.

css/src/components/accordion.scss
```diff
@@ -0,0 +1,156 @@
...
+			padding-inline-end: $icon-offset-size;
+
+			&::after {
+				inset-block-start: calc($padding-size * 1.5);
+			}
+		}
+	}
+
+	.accordion-content {
+		padding-block: $padding-size;
```
homing1, line 47: Seems like line 23 already sets the same padding for `.accordion-content`?

site/src/components/accordion.md
BenDMyers: @jwmunn @ollips It's also worth keeping in mind that it'll take some time for enough Firefox users to update to support `name` - see ["Evergreen" Does Not Mean Immediately Available](https://css-tricks.com/evergreen-does-not-mean-immediately-available/) - and so even when Firefox releases, we shouldn't necessarily *rely* on `name` to guarantee exclusivity.

That said, I think it's worth calling out that we can use `name` as a progressive enhancement -- if a browser version supports `name`, great, but if it doesn't, we wouldn't consider that "broken."

Which brings me to another point, which is should we put in guidance about when exclusivity is appropriate? I think Atlas should *discourage* exclusivity unless there's a really compelling reason why specifically only one `<details>` can be open at a time (where tidiness/compactness would not be a satisfactory reason).

If we had such guidance, it would also be appropriate to add a note about progressive enhancement in there.

site/src/components/accordion.md
```diff
@@ -0,0 +1,165 @@
...
+		<p>Accordion panel 3</p>
+	</div>
+</details>
+```
+
+## Attributes
+
+### Open
+
+There is an option to have the accordion component load in the expanded state by adding the `open` attribute to one of the items.
```
BenDMyers, line 49: Nit, slight rewording, please feel free to disregard:

```suggestion
To load the accordion component in the expanded state by default, add the `open` attribute to the appropriate `<details>` element.
```

site/src/components/accordion.md
```diff
@@ -0,0 +1,165 @@
...
+	</summary>
+	<div class="accordion-content">
+		<p>Accordion panel 2</p>
+	</div>
+</details>
+```
+
+## Sizes
+
+`accordion-sm`/`accordion-lg` modifier classes make the text size and padding a little bigger or smaller than the default.
```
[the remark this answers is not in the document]
BenDMyers: The reordering of "smaller" or "bigger" is to match the order that the modifier classes themselves are listed in.

site/src/patterns/accordion.md
```diff
@@ -0,0 +1,164 @@
...
+	</div>
+</details>
+```
+
+```html
+<details class="accordion accordion-lg border-top">
+	<summary>
+		<div class="accordion-header margin-left-xxs">
+			<h3 class="font-weight-semibold font-size-h5 line-height-normal">Accordion header 1 text</h3>
+			<p class="font-size-md color-text-subtle">Accordion subheader text</p>
```
BenDMyers, line 66: Two things about this nesting:

`<summary>`, having an implicit `button` role, will likely obliterate any semantics inside of it, meaning this `h3` will semantically not be treated as an `h3`. It might be worth just switching this to something more generic like a `<p>` and still applying the larger font size and whatnot - that way, we keep the presentation, but don't imply things like impacting heading-based navigation.

Secondly, every element in here gets collapsed into forming the `summary` button's accessible name - so in this case, the `summary` is named "Accordion header 1 text Accordion subheader text." Screenreaders will announce that in full, no differentiation between children, and in real world use cases, this could get quite verbose (especially when you throw in translation). Even worse, _some_ speech-dictation users would need to say this whole name in full to trigger the accordion: "Click Accordion header 1 text Accordion subheader text." This can also get really tedious in real world use cases.

I say this not to say that we _can't_ do this kind of subheading design, but rather, to suggest we minimize how often we do it, since it can introduce some heavy tediousness for some disabled users.

site/src/components/accordion.md
```diff
@@ -0,0 +1,165 @@
...
+	</summary>
+	<div class="accordion-content">
+		<p>Accordion panel content</p>
+	</div>
+</details>
+```
+
+## Patterns
+
+This page discusses the component and subcomponents that make up `.accordion`, but it does not prescribe particular patterns. For examples of specific accordion builds, look to the [accordion patterns page](../patterns/accordion.md) for guidance.
```
BenDMyers, line 165: This is a great disclaimer that I nearly missed because it's at the bottom. Could we find a way to incorporate this in the introductory section?

css/src/components/accordion.scss
jwmunn: I refactored it to no longer use absolute positioning and removed the background color styles. Updated the patterns to use atomics for padding.

site/src/components/accordion.md
jwmunn: ✅Updated

css/src/components/accordion.scss
jwmunn: ✅ Updated.

css/src/components/accordion.scss
jwmunn: ✅Updated

site/src/components/accordion.md
jwmunn: added the missing markup wrapper, removed the first example. The refactor of the arrow position keeps it no outer padding in both start and end positions so it can be added with atomics if desired.

image

css/src/components/accordion.scss
jwmunn: This was removed.

site/src/patterns/accordion.md
jwmunn: Understood, this pattern was included because it's being used on the [Career Path](https://learn.microsoft.com/en-us/training/career-paths/ai-engineer), if we want to discourage it we don't need to document it in atlas, was just documenting the existing patterns we'll be replacing with this component to deprecate the _details.scss in docs-ui.

Want me to remove this pattern from atlas? Updated it to use `<p>` tags

site/src/components/accordion.md
jwmunn: Moved it up to the intro section.

site/src/components/accordion.md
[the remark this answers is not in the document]
jwmunn: ✅ Updated copy

site/src/components/accordion.md
jwmunn: ✅Updated since I was already updating other copy.

css/src/components/accordion.scss
jwmunn: With the refactor of no longer using absolute positioning the padding is no longer needed as the ::before and ::after elements take up space in the flex layout

site/src/components/accordion.md
jwmunn: Both of the existing implementations currently in docs-ui allow multiple panels to be open at once, I agree that "name" can be a progressive enhancement.

jwmunn:
@ollips @BenDMyers @homing1 all feedback has been addressed. Ready for review.

css/src/mixins/chevron.scss
```diff
@@ -0,0 +1,33 @@
...
+	border: $chevron-arrow-border-width solid $chevron-color;
+	border-block-start: 0;
+	border-inline-end: 0;
+	pointer-events: none;
+	content: ' ';
+	z-index: $zindex-active;
+	inset-inline-end: $control-padding-horizontal;
+}
+
+@mixin chevron-right {
```
homing1, line 23: Do we need to account for rtl?

jwmunn: Yea we should, I've added rtl styles. To test add `dir="rtl"` to any parent element or details element

jwmunn: We don't have to change the direction in doc's UI because it's an up and down docon that aligns right or left. Since we've removed the docon dependency and are using rotations this needs to be accounted for in rtl for this component.

css/src/components/accordion.scss
```diff
@@ -0,0 +1,124 @@
+/* stylelint-disable selector-max-specificity, selector-no-qualifying-type */
```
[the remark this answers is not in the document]
ollips: I figured it out, apply `flex-shrink: 0;` to the icon and it will keep it in place

site/src/patterns/accordion.md
ollips: still see the `details` uses, please change to `accordion`

site/src/patterns/accordion.md
```diff
@@ -0,0 +1,164 @@
...
+	<div class="accordion-content padding-xs">
+		<p class="margin-bottom-xxs">
+			Viverra aliquet eget sit amet tellus cras. Nisl nisi scelerisque eu ultrices vitae.
+			Condimentum id venenatis a condimentum. Ut eu sem integer vitae.
+		</p>
+		<h4 class="font-weight-semibold">Accordion content subheading</h4>
+		<article class="card margin-top-xs">
+			<div class="card-template">
+				<p class="card-supertitle">Content Type</p>
+				<a href="#" class="card-title margin-bottom-none">The headline of the content item</a>
```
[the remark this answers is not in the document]
ollips: same thing on all demo examples

site/src/patterns/accordion.md
```diff
@@ -0,0 +1,164 @@
...
+		<div class="accordion-header font-weight-semibold">Accordion headline 1 text</div>
+	</summary>
+	<div class="accordion-content padding-xs">
+		<p class="margin-bottom-xxs">
+			Viverra aliquet eget sit amet tellus cras. Nisl nisi scelerisque eu ultrices vitae.
+			Condimentum id venenatis a condimentum. Ut eu sem integer vitae.
+		</p>
+		<h4 class="font-weight-semibold">Accordion content subheading</h4>
+		<article class="card margin-top-xs">
+			<div class="card-template">
```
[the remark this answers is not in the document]
ollips: same thing on all demo examples

site/src/patterns/accordion.md
```diff
@@ -0,0 +1,164 @@
...
+		</article>
+	</div>
+</details>
+```
+
+```html
+<details class="accordion accordion-lg border-top">
+	<summary>
+		<div class="accordion-header margin-left-xxs">
+			<p class="font-weight-semibold font-size-h5 line-height-normal">Accordion header 1 text</p>
```
ollips, line 65: the sizing should be controlled by the accordion's sizing atomics. If there is a need of even a bigger size variation - it should be added to the component. Remove `font-size-h5 line-height-normal`

site/src/patterns/accordion.md
```diff
@@ -0,0 +1,164 @@
...
+			<p class="font-size-md color-text-subtle">Accordion subheader text</p>
```
ollips, line 66: Adjust the font size atomic to match with `accordion-lg` from my previous comment. `font-size-sm`?

site/src/patterns/accordion.md
ollips: Revisit all examples

site/src/patterns/accordion.md
```diff
@@ -0,0 +1,164 @@
...
+title: Accordion
+description: Accordion patterns in the Atlas Design System
+template: standard
+---
+
+# Accordion patterns
+
+Because the `details` component has so many possible permutations, it can be helpful to provide examples of certain styles.
+
+## Accordion
```
ollips, line 11: Can you provide a brief title and explanation for every pattern? Check the cards pattern page for the reference. Right now it's just a pile of code which kinda hard to go through and understand what are the scenarios

site/src/components/accordion.md
ollips: Can you add a link to `details` please? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#customizing_the_disclosure_widget

css/src/components/accordion.scss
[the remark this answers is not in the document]
BenDMyers: That's a great catch - can we make sure to add a comment alongside the `flex-shrink`? Something like:

`# Prevent caret from being clipped when accordion title is long`

css/src/components/accordion.scss
```diff
@@ -0,0 +1,124 @@
...
+$accordion-font-size-sm: $font-size-9 !default;
+$accordion-font-size-md: $font-size-8 !default;
+$accordion-font-size-lg: $font-size-7 !default;
+
+$accordion-spacing-xs: $spacer-2 !default;
```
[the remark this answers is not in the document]
jwmunn: Sounds good, updated 👍

css/src/components/accordion.scss
```diff
@@ -0,0 +1,124 @@
...
+		.accordion-header {
+			flex-grow: 1;
+		}
+	}
+
+	&:not(.accordion-icon-end) {
+		summary::before {
+			@include chevron-right;
+
+			margin-inline-end: $accordion-spacing-sm;
```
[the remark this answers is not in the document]
jwmunn: Works great and now spacing can be controlled with the gap atomic on the `summary` element. Good call 👍

site/src/patterns/accordion.md
jwmunn: This has been updated.

site/src/components/accordion.md
jwmunn: Confirmed it now works in the production latest firefox release 🎉

site/src/patterns/accordion.md
jwmunn: Added xl and xxl large sizes and removed the font size atomics from the headers for this pattern

css/src/components/accordion.scss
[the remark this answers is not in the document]
jwmunn: Great catch, `flex-shrink` works great, also added a comment

jwmunn (commented):
Still need to come up with the pattern descriptions

jwmunn (commented):
@homing1 @BenDMyers @ollips feedback has been addressed, ready for review.

site/src/patterns/accordion.md
jwmunn: Added MDN link to this pattern about the button role's effects on semantics on L63

site/src/patterns/accordion.md
```diff
@@ -0,0 +1,153 @@
...
+				<a href="#" class="button button-sm button-primary">Button text</a>
+			</div>
+		</div>
+	</details>
+</div>
+```
+
+## Coda accordion
+
+These accordions us standard `.accordion` subcomponents, border-atomics to create a . They use `.accordion-icon-end` to right align the icon. Background color atomics are used for the `.accordion-content` subcomponent. Due to the border and backgound color atomics used, a consistent spacing padding atomic is used to align the header and panel content. Note that is recommended right aligned icon accordions not be full-width elements at desktop sizes.
```
homing1, line 126:
```suggestion
These accordions use standard `.accordion` subcomponents and border-atomics to create an accordion . They use `.accordion-icon-end` to right align the icon. Background color atomics are used for the `.accordion-content` subcomponent. Due to the border and background color atomics used, a consistent spacing padding atomic is used to align the header and panel content. It is recommended that accordions with right aligned icons not be full-width elements at desktop sizes.
```

site/src/patterns/accordion.md
```diff
@@ -0,0 +1,153 @@
...
+			</article>
+			<div class="text-align-center margin-top-xs">
+				<a href="#" class="button button-sm button-primary">Button text</a>
+			</div>
+		</div>
+	</details>
+</div>
+```
+
+## Coda accordion
```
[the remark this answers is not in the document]
jwmunn: image

Was trying to think of a word signifying end, it's the pattern used for our FAQ page which is generally at the end of a page, the icon is aligned to the end of the component. Open to other names

site/src/patterns/accordion.md
jwmunn: Added titles for the accordion patterns

site/src/patterns/accordion.md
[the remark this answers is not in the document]
homing1: I can't think of a word either. What do you think about calling it as it is? Accordion with right-aligned icon?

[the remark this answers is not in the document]
ollips: Gonna remove the patterns page for now. Designs have changed again so not a good time to settle on patterns.

ollips (approved):

homing1 (approved):

PR #672, opened 2024-09-06
Add scrollbar-gutter atomics

BenDMyers:
Task: task-[work-item-number]

Link: [preview](http://localhost:1111/atomics/overflow.html)

Introduces the `.scrollbar-gutter-stable` and `.scrollbar-gutter-stable-both-edges` atomics, to be used with overflowing/scrolling containers. These atomics can be used to prevent text from jumping around when a scrollbar appears or disappears.

## Testing

1. Navigate to http://localhost:1111/atomics/overflow.html, and scroll down to the two bottommost examples.
2. Use the examples' resize handles to resize the containers until scrollbars appear and disappear.
   Expected result: Text inside the containers should not jump around as scrollbars (dis)appear.

## Additional information

https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter

## Contributor checklist

- [X] Did you update the description of this pull request with a review link and test steps?
- [X] Did you submit a changeset? Changesets are required for all code changes.

### DIFF
--- a/.changeset/great-eyes-stare.md
+++ b/.changeset/great-eyes-stare.md
@@ -1,0 +1,5 @@
+---
+'@microsoft/atlas-site': patch
+---
+
+Document scrollbar-gutter atomics
--- a/.changeset/great-eyes-stare.md
+++ b/.changeset/great-eyes-stare.md
@@ -1,5 +1,5 @@
 ---
-'@microsoft/atlas-site': patch
+'@microsoft/atlas-site': minor
 ---
 
 Document scrollbar-gutter atomics
--- a/site/src/atomics/overflow.md
+++ b/site/src/atomics/overflow.md
@@ -5,16 +5,18 @@ template: standard
 classType: Atomics
 classPrefixes:
   - overflow
+  - scrollbar-gutter
 ---
 
 # Overflow
 
 At times, you'll need to determine the overflow behavior of an element. Atlas provides several classes to do this.
 
-| cssproperty | value    | screensize |
-| ----------- | -------- | ---------- |
-| `overflow`  | `hidden` | `tablet`   |
-| `overflow-x`  | `hidden` | `tablet`   |
+| cssproperty        | value                         | screensize |
+| ------------------ | ----------------------------- | ---------- |
+| `overflow`         | `hidden`                      | `tablet`   |
+| `overflow-x`       | `hidden`                      | `tablet`   |
+| `scrollbar-gutter` | `stable`, `stable-both-edges` |            |
 
 ## Usage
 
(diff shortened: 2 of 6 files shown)


.changeset/great-eyes-stare.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-site': patch
```
homing1, line 2: Should this be minor since we're not fixing existing atomics?

BenDMyers: Updated!

site/src/atomics/overflow.md
```diff
@@ -36,12 +38,63 @@ You can use `.overflow-x-hidden` to set the clipping behavior for the left/right
...
+				This text is clipped on its right edge because we gave the parent container overflow-hidden.
+			</p>
+		</div>
+	</div>
 </div>
 ```
+
+## Scrollbar gutter
+
+When using automatic overflow, as with the [scrolling components](../components/scroll.md), the sudden appearance or disappearance of a container's scrollbar can cause the container's content to jump around. While this might be fine if the scrollbar appears or disappears as a result of the user zooming or resizing their browser, it can look clunky otherwise — such as when messages are appended to a constrained-width or constrained-height chat container, resulting in a scrollbar where there wasn't one previously.
```
wibjorn, line 53: Not wholesale against this, but curious if this makes the scrollbars appear on platforms like mac/safari that hide them when they're inactive.

site/src/atomics/overflow.md
```diff
@@ -5,16 +5,18 @@ template: standard
...
 
-| cssproperty | value    | screensize |
-| ----------- | -------- | ---------- |
-| `overflow`  | `hidden` | `tablet`   |
-| `overflow-x`  | `hidden` | `tablet`   |
+| cssproperty        | value                         | screensize |
+| ------------------ | ----------------------------- | ---------- |
+| `overflow`         | `hidden`                      | `tablet`   |
+| `overflow-x`       | `hidden`                      | `tablet`   |
+| `scrollbar-gutter` | `stable`, `stable-both-edges` |            |
```
wibjorn, line 19: wow that is some verbose css :D

site/src/atomics/overflow.md
BenDMyers: [Safari doesn't support this](https://caniuse.com/mdn-css_properties_scrollbar-gutter) I suspect precisely because it uses overlay scrollbars.

But this would absolutely be worth further testing to be sure!

wibjorn (approved):

site/src/atomics/overflow.md
BenDMyers: Per [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter), when overlay scrollbars are used, a la macOS, there is no gutter, so this wouldn't have any impact.

homing1 (approved):

PR #694, opened 2024-10-08
move flyout to be under the header

wibjorn:
Flyout should render under the header and above the footer, not adjacent to them.

### DIFF
--- a/css/src/components/layout.scss
+++ b/css/src/components/layout.scss
@@ -217,7 +217,7 @@ $default-flyout-width-widescreen: 480px;
 				'header header header'
 				'hero hero flyout'
 				'menu main flyout'
-				'footer footer flyout';
+				'footer footer footer';
 		}
 		@include widescreen {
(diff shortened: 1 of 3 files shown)


css/src/components/layout.scss
```diff
@@ -214,7 +214,7 @@ $default-flyout-width-widescreen: 480px;
 					#{$layout-flyout-width-name}
 				);
 			grid-template-areas:
-				'header header flyout'
+				'header header header'
 				'hero hero flyout'
 				'menu main flyout'
 				'footer footer flyout';
```
ellynliu, line 220: why do you have 'footer footer flyout' here in line 220 but 'footer footer footer footer' in line 169? like, is line 220 supposed to say 'footer footer footer'?

wibjorn: Great catch! Updated

ellynliu (approved):

homing1 (approved):

BenDMyers (approved):

PR #696, opened 2024-10-09
Add steps component 

ellynliu:
Task: task-[work-item-number]

Link: [preview](http://localhost:1111/components/steps.html)

Add the steps component

## Testing

1. http://localhost:1111/components/steps.html
2. Review changes

## Additional information

[Optional]

## Contributor checklist

### DIFF
--- a/css/src/components/steps.scss
+++ b/css/src/components/steps.scss
@@ -9,13 +9,11 @@ $steps-size-double: 2 * $steps-size;
 	.step {
 		width: $steps-size;
 		height: $steps-size;
-		border-radius: 50%;
+		border-radius: $border-radius-rounded;
 		background-color: $secondary;
 
 		&.step-selected {
 			width: $steps-size-double;
-			height: $steps-size;
-			border-radius: $steps-size;
 			background-color: $primary;
 		}
 	}
--- a/css/src/components/steps.scss
+++ b/css/src/components/steps.scss
@@ -1,0 +1,22 @@
+$steps-gap: 7px;
+$steps-size: 8px;
+$steps-size-double: 16px;
+
+.steps {
+	display: flex;
+	gap: $steps-gap;
+
+	.step {
+		width: $steps-size;
+		height: $steps-size;
+		border-radius: 50%;
+		background-color: $secondary;
+
+		&.step-selected {
+			width: $steps-size-double;
+			height: $steps-size;
+			border-radius: $steps-size;
+			background-color: $primary;
+		}
+	}
+}
--- a/css/src/components/steps.scss
+++ b/css/src/components/steps.scss
@@ -1,7 +1,6 @@
 $step-font-size: $font-size-7 !default;
 $steps-gap: 0.4375em !default;
 $steps-size: 0.5em !default;
-$steps-size-double: 2 * $steps-size;
 
 .steps {
 	display: flex;
--- a/css/src/components/steps.scss
+++ b/css/src/components/steps.scss
@@ -1,6 +1,6 @@
 $steps-gap: 7px;
 $steps-size: 8px;
-$steps-size-double: 16px;
+$steps-size-double: 2 * $steps-size;
 
 .steps {
 	display: flex;
--- a/css/src/components/steps.scss
+++ b/css/src/components/steps.scss
@@ -1,19 +1,22 @@
-$steps-gap: 7px;
-$steps-size: 8px;
+$step-font-size: $font-size-7 !default;
+$steps-gap: 0.4375em !default;
+$steps-size: 0.5em !default;
 $steps-size-double: 2 * $steps-size;
 
 .steps {
 	display: flex;
 	gap: $steps-gap;
+	flex-wrap: nowrap;
+	font-size: $step-font-size;
 
 	.step {
-		width: $steps-size;
 		height: $steps-size;
+		aspect-ratio: 1 / 1;
 		border-radius: $border-radius-rounded;
 		background-color: $secondary;
 
 		&.step-selected {
-			width: $steps-size-double;
+			aspect-ratio: 2 / 1;
 			background-color: $primary;
 		}
 	}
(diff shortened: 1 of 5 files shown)


site/src/components/steps.md
```diff
@@ -0,0 +1,53 @@
+---
+title: Steps
+description: Steps the Atlas Design System
+template: standard
+figmaEmbed: https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com/design/7raji0IWkuBRxs5kLiADgI/Copilot-Web-UI-kit?node-id=70108-9255&node-type=frame&t=G0FoPLEXPee1yysa-0
+classType: Component
+classPrefixes:
+  - steps
```
BenDMyers, line 8: This also needs the `step` prefix, to catch the `.step` and `.step-selected` classes.

```suggestion
classPrefixes:
  - step
  - steps
```

css/src/components/steps.scss
```diff
@@ -0,0 +1,22 @@
...
+	gap: $steps-gap;
+
+	.step {
+		width: $steps-size;
+		height: $steps-size;
+		border-radius: 50%;
+		background-color: $secondary;
+
+		&.step-selected {
+			width: $steps-size-double;
```
BenDMyers, line 16: As a possibility:

```suggestion
			width: 2 * $steps-size;
```

This way we can use fewer variables, and if the step-size changes, the wider version automatically updates as well?

site/src/components/steps.md
```diff
@@ -0,0 +1,53 @@
+---
+title: Steps
+description: Steps the Atlas Design System
```
homing1, line 3:
```suggestion
description: The steps component in the Atlas Design System
```

css/src/components/steps.scss
```diff
@@ -0,0 +1,22 @@
...
+$steps-size-double: 2 * $steps-size;
+
+.steps {
+	display: flex;
+	gap: $steps-gap;
+
+	.step {
+		width: $steps-size;
+		height: $steps-size;
+		border-radius: 50%;
```
homing1, line 12: Would it be the same if we use `$border-radius-rounded` for the border radius?

css/src/components/steps.scss
```diff
@@ -0,0 +1,22 @@
...
+		background-color: $secondary;
+
+		&.step-selected {
+			width: $steps-size-double;
+			height: $steps-size;
```
BenDMyers, line 17: Do we need to repeat this `height`?

css/src/components/steps.scss
```diff
@@ -0,0 +1,20 @@
+$steps-gap: 7px;
+$steps-size: 8px;
```
homing1, line 2: Any chance that the size will need to scale with font size?

ellynliu: I don't think so, I got these measurements from the designer and the fluent figma has a fixed size

css/src/components/steps.scss
```diff
@@ -0,0 +1,20 @@
...
+	display: flex;
+	gap: $steps-gap;
+
+	.step {
+		width: $steps-size;
+		height: $steps-size;
+		border-radius: $border-radius-rounded;
+		background-color: $secondary;
+
+		&.step-selected {
```
homing1, line 15: What do you think about something like `step-current` instead to indicate current progress? Feel free to resolve if you don't see a difference.

site/src/components/steps.md
```diff
@@ -0,0 +1,54 @@
...
+	<div class="step"></div>
+</div>
+<div class="steps padding-xxs">
+	<div class="step"></div>
+	<div class="step"></div>
+	<div class="step"></div>
+	<div class="step"></div>
+	<div class="step step-selected"></div>
+</div>
+```
```
wibjorn, line 54: I'm thinking that that our examples need to be paired with some element containing the description of the current state of the steps: such as `<p class="visually-hidden">Step 1 of 5</p>`.

```suggestion
```

## Accessibility concerns

Although the step component is not interactive, it is also not entirely presentational. Thus, it's recommended that you put `aria-hidden` on the component, while including a `aria-hidden` element that communicates the current state of the user's progress through the steps.
```

Comment (BenDMyers) [reply site/src/components/steps.md:54]: Nit to @wibjorn's suggestion, this should be

```diff
-while including a `aria-hidden` element that communicates the current state of the user's progress through the steps.
+while including a `visually-hidden` element that communicates the current state of the user's progress through the steps.
```

css/src/components/steps.scss
wibjorn: Was going to suggest that we do font-size scaling anyway; it's something nearly all components do and will protect us from design churn ;).

I think this will work (I've also used aspect-ratio too instead of definite heights):

```scss
$step-font-size: $font-size-7 !default;
$steps-gap: 0.4375em !default;
$steps-size: 0.5em !default;
$steps-size-double: 2 * $steps-size;

.steps {
	display: flex;
	gap: $steps-gap;
	font-size: $step-font-size;
	flex-wrap: nowrap;

	.step {
		height: $steps-size;
		aspect-ratio: 1 / 1;
		background-color: $secondary;

		&.step-selected {
			aspect-ratio: 2 / 1;
			background-color: $primary;
		}
	}
}

File: site/src/components/steps.md (hunk 8 above)
Comment (wibjorn) [reply site/src/components/steps.md:59]: like the nit

Comment (wibjorn) [reply site/src/components/steps.md:59]: since I was exactly wrong :D

**Review by wibjorn — APPROVED** [review]

File: css/src/components/steps.scss [hunk 9]
```diff
@@ -0,0 +1,23 @@
+$step-font-size: $font-size-7 !default;
+$steps-gap: 0.4375em !default;
+$steps-size: 0.5em !default;
+$steps-size-double: 2 * $steps-size;
```
BenDMyers, line 4: With the `aspect-ratio` addition below, this variable isn't used anymore.

ellynliu: thanks, removed!

BenDMyers (approved):

wibjorn (approved):

PR #723, opened 2024-11-14
Update primary color

homing1:
Task: task-1012252

- Hyperlink color contrast fails when I update it to Fluent color 115ea3, so not updating for now.
  - https://github.com/microsoft/atlas-design/actions/runs/11845506600
- Using Fluent's primary button as basis for `primary-base`, `primary-dark`, `primary-hover`
  - http://localhost:1111/components/button.html
  - https://react.fluentui.dev/?path=/docs/components-button-button--docs
    - Note that Fluent does not have primary + outline, primary + subtle/clear buttons, just one or the other.
- Using Fluent's tag background as basis for `primary-background`, 
  - http://localhost:1111/components/tag.html
  - https://react.fluentui.dev/?path=/docs/components-tag-interactiontag--docs
    - Note 1: For all Fluent interactive tags, hover color is blue 115ea3 on the dismiss button. 
- Introducing two new items to the theme map for semantic colors
    -  `-background-hover` and `-dark-hover`. 
      -  The values for non-primary colors are not finalized. They will be re-evaluated as their colors in the existing palette are converted to Fluent colors in future PRs.
- Took colors from Fluent's color token page to create the palette https://react.fluentui.dev/?path=/docs/theme-colors--docs
  - Fluent's true color palette https://github.com/microsoft/fluentui/blob/master/packages/tokens/src/global/colors.ts and https://github.com/microsoft/fluentui/blob/master/packages/tokens/src/global/brandColors.ts
    - Comparing with the colors on the color token page, it seems like many blue colors in the palette file are not being used for a color token.
- Fluent Figma https://www.figma.com/design/FO4ImdAytGCc5DQXpNFObI/Microsoft-Fluent-2-Web-(Community)?node-id=8911-3188&node-type=canvas&t=t2BY46Qo57xosf94-0

### Affected components
$primary
- card title color
- pagination background
- progress bar background
- segmented controls font
- steps background
- toggle background
- checkbox color
- input, select, textarea focus border
- button

$primary-background
- message background
- segmented controls hover background
- primary tag background, border

$primary-dark
- primary interactive tag color, hover

$primary-hover
- accordion hover
- pagination hover
- segmented controls hover
- primary tag hover

NEW
**$primary-background-hover**
- primary interactive tag background hover

**$primary-dark-hover**
-  primary interactive tag hover

## Testing

1. Visit http://localhost:1111/components/button.html and http://localhost:1111/components/tag.html to review the main changes. Compare with Fluent's page https://react.fluentui.dev/?path=/docs/components-button-button--docs and existing Atlas design.atlas.microsoft.com
2. Check all 3 themes.
3. Review other affected components
- http://localhost:1111/components/accordion.html
- http://localhost:1111/components/badge.html
- http://localhost:1111/components/card.html
- http://localhost:1111/components/checkbox.html
- http://localhost:1111/components/input.html
- http://localhost:1111/components/message.html
- http://localhost:1111/components/notification.html
- http://localhost:1111/components/pagination.html
- http://localhost:1111/components/progress-bar.html
- http://localhost:1111/components/radio.html
- http://localhost:1111/components/segmented-controls.html
- http://localhost:1111/components/steps.html
- http://localhost:1111/components/toggle.html
4. Accessibility scan passes, meaning that components meet contrast ratio on light and dark theme. Button and tag documentation page includes base and hover states. For other components, check hover states.

## Additional information

[Optional]

## Contributor checklist


css/src/tokens/themes.scss
```diff
@@ -42,15 +42,17 @@ $themes: (
...
-		'primary-dark': $palette-blue-80,
-		'primary-hover': $palette-blue-60,
-		'primary-active': $palette-blue-70,
+		'primary-background-glow-high-contrast': $palette-blue-10-deprecated,
+		'primary-dark': $palette-blue-120,
+		'primary-hover': $palette-blue-120,
+		'primary-active': $palette-blue-170,
 		'primary-box-shadow': $palette-blue-opacity-30,
 		'primary-invert': $palette-white,
+		'primary-hover-light-background': $palette-blue-20,
```
homing1, line 54: Added this to introduce a new color for tags. I'm not confident that the other semantic colors will also need such item, but we will need to add it to other colors make the color index in colors.scss consistent.

css/src/tokens/themes.scss
```diff
@@ -261,6 +265,8 @@ $themes: (
 		'primary-active': $palette-yellow-high-contrast-hover,
 		'primary-box-shadow': $palette-white,
 		'primary-invert': $palette-black,
+		'primary-hover-light-background': $palette-yellow-high-contrast-hover,
+		'primary-hover-light-text': $palette-black,
```
homing1, line 269: `primary-hover-light-text` is the same as primary-dark except for high contrast. This was added in order to keep the tag hover text color the same as Atlas's current colors in high contrast. Alternatively, we update so that dark and high contrast also matches Fluent.

css/src/components/tag.scss
```diff
@@ -200,6 +200,13 @@ $tag-interactive-color-primary-hover: $primary-invert !default;
 
 		.tag-close {
 			border-inline-start-color: $tag-interactive-divider-color-filled;
+
+			&:hover,
```
wibjorn, line 204: If this is the source of needing a new item added to the theme map, can we get by without it?

css/src/tokens/themes.scss
wibjorn: Perhaps you could outline why this was added? Looking though both the badge, button, and tag. I don't see why we need this one when we have the current -hover value.

I do see the background color in tag seems to be very light (pallete-10), but I don't see the "X" button having any hover styles at all.

https://react.fluentui.dev/?path=/docs/components-tag-tag--docs

css/src/tokens/themes.scss
```diff
@@ -147,15 +149,17 @@ $themes: (
...
 		'score-high-off': $palette-green-opacity-70,
 		'score-high': $palette-green-40,
-		'hyperlink': $palette-blue-30,
-		'primary-base': $palette-blue-30,
-		'primary-background': $palette-blue-80,
-		'primary-background-glow-high-contrast': $palette-blue-80,
-		'primary-dark': $palette-blue-20,
-		'primary-hover': $palette-blue-40,
-		'primary-active': $palette-blue-50,
+		'hyperlink': $palette-blue-90,
```
wibjorn, line 152: Suggestion only: perhaps revert all dark theme changes and swing back to that as a separate PR? There's enough to think through going one theme and one semantic color at a time. The mix of old and new palette may make things harder to think about as well.

css/src/tokens/themes.scss
homing1: Each tag dismiss button has 115ea3 on hover (the color that I've set for `primary-dark` and `primary-hover`, it's just really hard to see. It corresponds to the `colorCompoundBrandForeground1Hover` Fluent color token.

image

image

`primary-base/hover` values were updated to match primary buttons, hence no adjustments are needed for buttons. It still only uses two blue colors. 

For Atlas tag colors, we've kept it simple by using three blue colors (primary-dark on hover). 

`primary-hover-light-background` was added for interactive tags. On hover, Fluent uses a different blue that's not already included in primary-base, dark, nor hover values. There's a total of four blue colors.

image

image

A pattern that I've noticed is that Fluent can have multiple combos for the same light theme color, but different color for dark and/or high contrast.

Another pattern is using more than one set of background/foreground/whatever colors. Having just one set of `primary-background` colors may not fit all of the use cases.

css/src/tokens/themes.scss
homing1: Separate as in a PR with light theme changes, and a second PR with only dark theme changes? or a second PR with both light and dark theme changes? In parallel or in succession? I'm open to having separate PRs if that makes it easier to review.

The situation I'd like to avoid is being caught in a state where we merge the light theme, knowing that some colors don't work temporarily in dark/high-contrast theme. We address those themes in a separate PR, but the time gap between the two PRs may be long and we end up needing to ship something else (timeline component for example) during that in-between state. This situation is avoidable if we make sure to merge all 2 or 3 PRs at the same time and deal with the potential merge conflicts.

css/src/tokens/themes.scss
homing1: A pattern that I've noticed is that Fluent can have multiple combos for the same light theme color, but different color for dark and/or high contrast.

Another pattern is using more than one set of background/foreground/whatever colors. Having just one set of `primary-background` colors may not fit all of the use cases. 

image

css/src/tokens/themes.scss
homing1: The change to dark theme is mainly to point to the new names from the old palette. There were two actual value changes that I've reverted. The remaining changes are the two new items, which need to have a value for every theme.

css/src/components/tag.scss
homing1: This is added to ensure that every tag dismiss button on hover has the 115ea3 dark blue text color. When I remove this, the hover color still works except for the filled dismiss button on hover. The text color remains a black as it currently does now in Atlas.

image

css/src/tokens/themes.scss
homing1: Renamed to `primary-dark-hover`

css/src/tokens/themes.scss
homing1: Renamed to `primary-background-hover`

css/src/components/tag.scss
homing1: Keeping the rule since we've decided to add new items to the theme map.

wibjorn (dismissed):

css/src/tokens/colors.scss
```diff
@@ -136,7 +139,9 @@ $colors: (
 		$primary-active,
 		$primary-invert,
 		$primary-box-shadow,
-		$primary-background-glow-high-contrast
+		$primary-background-glow-high-contrast,
```
wibjorn, line 142: Just confirming, do we need the other -background-hover and dark-hover items in this map?

css/src/tokens/colors.scss
```diff
@@ -66,8 +67,10 @@ $score-high: var(--theme-score-high);
 
 $primary: var(--theme-primary-base);
 $primary-background: var(--theme-primary-background);
+$primary-background-hover: var(--theme-primary-background-hover);
 $primary-background-glow-high-contrast: var(--theme-primary-background-glow-high-contrast);
 $primary-dark: var(--theme-primary-dark);
+$primary-dark-hover: var(--theme-primary-dark-hover);
```
wibjorn, line 73: Similar question as below, do we need the -dark-hover and background-hover css variables created in this file?

css/src/tokens/colors.scss
homing1: They do need to be added at some point. I'm thinking that we can add them to each map as its colors are updated and finalized. I'm also not opposed to adding them now.

css/src/tokens/colors.scss
homing1: Same as above

css/src/tokens/colors.scss
wibjorn: Probably best to add them now, in my opinion. That work?

homing1: Added

wibjorn (approved):

wibjorn (approved):

BenDMyers (approved):

PR #730, opened 2024-11-25
Fixing default chevron icon appearance in accordion component

ollips:
Task: task-1017791
Link: [preview](http://localhost:1111/components/accordion.html)

There is an issue with the default chevron icon appearance in the accordion component (check the screenshot below). This PR fixes it

## Testing

1. `git checkout olga/accordion-fix-icon; npm run start`.
2. Visit accordion page
3. Inspect component, in dev tool disable "display: flex" rule for `.accordion summary`. You should not see the default chevron showing up

image

Here is the same but in `main` branch:

image

## Additional information

image

## Contributor checklist

- [ ] Does your pull request change any transforms? Did you test they work on right to left?

### DIFF
--- a/css/src/components/accordion.scss
+++ b/css/src/components/accordion.scss
@@ -37,9 +37,14 @@ $accordion-transition: transform 0.15s !default;
 	summary {
 		display: flex;
 		line-height: 1.5;
+		list-style: none;
 		cursor: pointer;
 		padding-block: $accordion-spacing;
 		gap: $accordion-gap;
+
+		&::-webkit-details-marker {
+			display: none;
+		}
 
 		&::before,
 		&::after {
--- a/.changeset/late-otters-retire.md
+++ b/.changeset/late-otters-retire.md
@@ -1,0 +1,5 @@
+---
+'@microsoft/atlas-css': minor
+---
+
+Fixing default chevron icon appearance in accordion component.


bbalentine87:
@ollips  The preview link in the description isn't working for me.

ollips:
> @ollips The preview link in the description isn't working for me.

Oh, that's right, you need to run it locally. I'll update test steps in a sec to reflect that

css/src/components/accordion.scss
```diff
@@ -37,10 +37,15 @@ $accordion-transition: transform 0.15s !default;
 	summary {
 		display: flex;
 		line-height: 1.5;
+		list-style: none;
 		cursor: pointer;
 		padding-block: $accordion-spacing;
 		gap: $accordion-gap;
 
+		&::-webkit-details-marker {
```
homing1, line 45: Just to clarify, is the desired behavior to not show any chevron/marker at all, if for some reason the display is changed from flex to something else? If so, wouldn't we still need some sort of indicator to show that there's a toggle button for the accordion panels?

ollips: Oh, we have a custom arrow there. Check out the screenshot, there are two icons but should be just one . 
In test steps I mentioned toggling the flex display rule cause it's the only way to see the difference before merging this PR

homing1: I couldn't get both chevrons to show at the same time, but I see the `<marker>` appearing while we only want to show the `:before`.  Resolving.

homing1 (approved):

bbalentine87 (approved):

PR #742, opened 2025-02-20
Update body background, body background medium, inline-code, and hover-invert colors on dark theme

homing1:

This PR updates the body-background color on dark theme. Body-background-medium, inline-code, and hover-invert colors are also updated to maintain sufficient color contrast.

Fluent colors: https://react.fluentui.dev/?path=/docs/theme-colors--docs

## Testing

1. Visit http://localhost:1111/atomics/colors.html, switch to dark theme. Review the body and body medium colors. Note that `body-background-medium` is lighter than `body-background`, same as before.
2. Visit http://localhost:1111/components/button.html . Review the default button hover color. It matches Fluent's `colorNeutralBackground1Hover` color on dark theme, #3d3d3d .
3. Review the inline code in the table's `Class` column. I couldn't find any Fluent guidance on inline code colors, so I picked a bright enough color to make the inline code stand out against the new background. The contrast ratio between the background and inline code block background is 2.013, which is an improvement compared to the previous 1.7~. 
4. Visit http://localhost:1111/components/textarea.html . Review the disabled text area background color for the new body background medium color # 333. The former color #2F2F2F only has 1.086:1 contrast ratio with the new body background color. The previous ratio is 1.339:1. The new body background medium color offer 1.151:1 ratio. It matches Fluent's `colorNeutralBackground6` color, which is already the lightest grey color for dark theme amongst all 6 `colorNeutralBackground` colors. To improve the ratio even more, I see two options:
   - Use # 000 black to make body medium darker than body instead of lighter, which increases the ratio to 1.443:1. This would change the current pattern of moving dark to light. 
   - Use #3d3d3d which is `colorNeutralBackground1Hover`, which increases the ratio to 1.339:1

## Additional information

[Optional]

## Contributor checklist

- [ ] Does your pull request change any transforms? Did you test they work on right to left?


wibjorn (approved):

wibjorn:
Code is good, and this works well. This does seem quite light for a dark mode. Wonder if we should discuss the possibility of a darker color instead?

PR #819, opened 2026-04-06
add a new reading-width component and adjust site layout to make use …

wibjorn:
The first of several PRs allowing for a nicer reading experience for users. This one:
1. Adds a new `.reading-width` class component, which is 688 plus padding on both sides.
1. Adjusts the `--layout-gap` variable to be smaller 16px on mobile and 24px on desktop (desktop is unchanged).
1. Adjusts the site's css a little bit, to add one level of nesting to allow for the application of `.reading-width`.

Link: [preview](http://localhost:1111/components/reading-width.html)

## Testing

1. Run locally.
2. Visit page above.
3. Resize screen, you ought to see a constrained container and it should never overlfow the boundaries of the page.

## Additional information

image

## Contributor checklist

### DIFF
--- a/css/src/tokens/layout.scss
+++ b/css/src/tokens/layout.scss
@@ -8,6 +8,7 @@ $column-gap: 0.75rem !default;
 $large-column-gap: 1.5rem !default;
 $reading-max-width: 50rem !default;
 
+$layout-gap-narrow: 16px !default;
 $layout-gap: 24px !default;
 $layout-widescreen-width: breakpoints.$breakpoint-widescreen - $layout-gap * 2 !default;
 $layout-widescreen-gap: max(
--- a/site/src/components/reading-width.md
+++ b/site/src/components/reading-width.md
@@ -1,0 +1,31 @@
+---
+title: Reading Width
+description: The reading width component in the Atlas Design System
+template: standard
+classType: Component
+classPrefixes:
+  - reading-width
+---
+
+# Reading width
+
+The reading width component constrains an element's maximum width to an optimal reading measure. This improves readability for long-form text content by preventing lines from becoming too wide, which can make it difficult for readers to track from the end of one line to the beginning of the next.
+
+## Usage
+
+Apply the `.reading-width` class to any container whose content should be constrained to a comfortable reading width. This page's text uses this very class!
+
+```html-no-example
+<div class="reading-width">
+	<p>
+		This paragraph is constrained to an optimal reading width. Long lines of text are harder to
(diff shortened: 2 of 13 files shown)


bbalentine87:
LGTM!

bbalentine87 (approved):

site/src/components/reading-width.md
```diff
@@ -0,0 +1,31 @@
...
+		read because the eye has to travel further from the end of one line to the beginning of the
+		next, making it easy to lose your place. By limiting the maximum width, we ensure a
+		comfortable reading experience.
+	</p>
+</div>
+```
+
+## Recommended implementation
+
+It's a good idea to use `.reading-width` in combination with a container that is centered either within the viewport or a parent element. Try using it in combination with `margin-inline-auto` to automatically center the content horizontally. If used in a grid, you may need to ensure the grid-column that container the reading-width container uses the `minmax` function to ensure it `margin-inline-auto` does not break outside the layout.
```
bbalentine87: What do you think about changing this to
"Use `.reading-width` within a container that is centered in the viewport or a parent element. Pair it with `.margin-inline-auto` to automatically center the content horizontally. If used in a grid, ensure the grid column containing the `.reading-width` container uses the `minmax` function so that `margin-inline-auto` does not cause content to break outside the layout."

bbalentine87 (changes requested):

css/src/tokens/layout.scss
```diff
@@ -8,6 +8,7 @@ $column-gap: 0.75rem !default;
 $large-column-gap: 1.5rem !default;
 $reading-max-width: 50rem !default;
 
+$layout-gap-narrow: 16px !default;
```
sahiljohari, line 11: just out of curiosity, why do we use both `rem` and `px` units instead of standardizing on any one of them?

lmackenzie8 (dismissed):
LGTM

css/src/tokens/layout.scss
wibjorn: It has to do with how each unit works. `rem` is based on the default document font size (px does not scale, and stays static). If we/the user changes their document font size, you get a different value. For this particular value, we don't want the gutters of the change, thus the usage of pixels.

site/src/components/reading-width.md
wibjorn: Updated!

bbalentine87 (approved):

sahiljohari (approved):

PR #826, opened 2026-04-14
fix: resolve Dart Sass deprecation warnings

wibjorn:

Replace deprecated global built-in functions with their modern namespaced equivalents to prepare for Dart Sass 3.0.0:

- nth() -> list.nth() with @use 'sass:list'
- map-get() -> map.get() (sass:map already imported)
- transparentize() -> color.adjust() with @use 'sass:color'
- if() function -> @if/@else block conditionals

## Testing

1. Run `npm start` locally. No Scss related errors.

## Contributor checklist

### DIFF
--- a/css/src/mixins/font-size.scss
+++ b/css/src/mixins/font-size.scss
@@ -29,8 +29,10 @@ $font-size-scaling-factor: 0.75 !default;
 	$y: strip-units($start-rem) * strip-units(tokens.$document-font-size);
 	$x: strip-units($start-width);
 	$b: $y - $m * $x;
-	$importantValue: if($important, !important, '');
-
-	font-size: clamp(#{$start-rem}, #{$b}px + #{$m * 100}vw, #{$end-rem}) #{$importantValue};
+	@if $important {
+		font-size: clamp(#{$start-rem}, #{$b}px + #{$m * 100}vw, #{$end-rem}) !important;
+	} @else {
+		font-size: clamp(#{$start-rem}, #{$b}px + #{$m * 100}vw, #{$end-rem});
+	}
 }
 /* stylelint-enable */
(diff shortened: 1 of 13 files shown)


css/src/mixins/font-size.scss
```diff
@@ -29,8 +29,10 @@ $font-size-scaling-factor: 0.75 !default;
 	$y: strip-units($start-rem) * strip-units(tokens.$document-font-size);
 	$x: strip-units($start-width);
 	$b: $y - $m * $x;
-	$importantValue: if($important, !important, '');
-
-	font-size: clamp(#{$start-rem}, #{$b}px + #{$m * 100}vw, #{$end-rem}) #{$importantValue};
+	@if $important {
```
sahiljohari, line 32: curious about this `$important` variable - it is set to false above and doesn't seem like its being set again in here. Is there a way to set it from outside of this code?

css/src/mixins/font-size.scss
```diff
@@ -29,8 +29,10 @@ $font-size-scaling-factor: 0.75 !default;
...
+		font-size: clamp(#{$start-rem}, #{$b}px + #{$m * 100}vw, #{$end-rem}) !important;
+	} @else {
+		font-size: clamp(#{$start-rem}, #{$b}px + #{$m * 100}vw, #{$end-rem});
```
sahiljohari, line 35: minor but might be worth assigning the `clamp` value to a variable?

wibjorn: don't think we can here

css/src/mixins/font-size.scss
wibjorn: not that i know of

sahiljohari (approved):

ellynliu (approved):

PR #856, opened 2026-05-07
feat: add tree component stylings

sahiljohari:
## Overview

Add a reusable tree CSS component for hierarchical navigation (tables of contents, file browsers, nested menus).

### Features
- Expand/collapse with indicator rotation and RTL support
- Selected state with configurable tokens and forced-colors accessibility
- Focus delegation to expander via focus-visible mixin
- Static (non-collapsible) tree variant (\.tree-static\)
- Icon leaf layout (\.has-icon\)
- Border-radius on interactive items
- Logical properties throughout
- All values use Atlas tokens

### Notes
- The expand/collapse indicator uses a DOM element (not CSS pseudo-elements like accordion) to stay compatible with the existing JS markup contract in consuming apps. This could be aligned to the accordion pattern in a future iteration.
- Documentation page with examples, DOM contract, and accessibility section included.
- Accessibility and visual diff tests registered.

Related work item: [#1143601](https://dev.azure.com/ceapex/Engineering/_workitems/edit/1143601)

### DIFF
--- a/css/src/components/tree.scss
+++ b/css/src/components/tree.scss
@@ -123,8 +123,15 @@ $tree-selected-font-weight: tokens.$weight-semibold !default;
 		}
 	}
 
-	&-item.is-expanded > .tree-expander::before {
-		transform: mixins.$chevron-down-rotate translate(calc(mixins.$chevron-arrow-size / 2), 0);
+	&-item.is-expanded > .tree-expander {
+		&::before {
+			transform: mixins.$chevron-down-rotate translate(calc(mixins.$chevron-arrow-size / 2), 0);
+		}
+
+		&:dir(rtl)::before {
+			transform: mixins.$chevron-down-rotate-rtl
+				translate(calc(mixins.$chevron-arrow-size / 2 * -1), 0);
+		}
 	}
 
 	&.tree-static {
--- a/css/src/components/tree.scss
+++ b/css/src/components/tree.scss
@@ -126,7 +126,7 @@ $tree-selected-font-weight: tokens.$weight-semibold !default;
 	}
 
 	&-item.is-expanded > .tree-expander > .tree-expander-indicator {
-		rotate: $tree-expanded-rotation;
+		rotate: $tree-expanded-rotation; // TODO: compare expanded chevron between this and prod - see if docs-ui is influencing the styling
 	}
 
 	&.tree-static {
--- a/css/src/components/tree.scss
+++ b/css/src/components/tree.scss
@@ -18,7 +18,6 @@ $tree-selected-font-weight: tokens.$weight-semibold !default;
 
 	&,
 	&-group {
-		list-style-type: none;
 		margin-block-end: 0;
 	}
 
--- a/css/src/components/tree.scss
+++ b/css/src/components/tree.scss
@@ -1,0 +1,137 @@
+@use '../tokens/index.scss' as tokens;
+@use '../mixins/index.scss' as mixins;
+
+$tree-font-size: tokens.$font-size-8 !default;
+$tree-indent: tokens.$spacer-5 !default;
+$tree-item-vertical-padding: tokens.$spacer-2 !default;
+$tree-item-gap: tokens.$spacer-2 !default;
+$tree-easing: ease-in-out !default;
+$tree-transition-duration: 0.15s !default;
+$tree-expanded-rotation: 90deg !default;
+$tree-selected-color: tokens.$text-glow-high-contrast !default;
+$tree-selected-background: tokens.$selected-background-subtle !default;
+$tree-selected-font-weight: tokens.$weight-semibold !default;
+
+.tree {
+	position: relative;
+	font-size: $tree-font-size;
+	line-height: 1.75;
+
+	&,
+	&-group {
+		list-style-type: none;
+		margin-block-end: 0;
+	}
+
+	&-group {
+		margin-inline-start: $tree-indent;
+	}
+
+	&-item {
+		margin-block-start: $tree-item-gap;
+
+		> .tree-group {
+			display: none;
+		}
+
+		&.is-expanded > .tree-group {
+			display: block;
+		}
+
+		&.tree-leaf {
+			color: tokens.$text !important;
+		}
+
+		&.is-selected {
+			background-color: $tree-selected-background;
+			color: $tree-selected-color !important;
+			font-weight: $tree-selected-font-weight;
+
+			@include mixins.forced-colors {
+				background-color: SelectedItem;
+				color: SelectedItemText !important;
+				forced-color-adjust: none;
+			}
+		}
+	}
+
+	&-item.tree-leaf.has-icon {
+		display: flex;
+
+		&:hover {
+			text-decoration: none;
+		}
+
+		.icon {
+			flex-shrink: 0;
+			align-items: baseline;
+			width: 1rem;
+			font-size: tokens.$font-size-9;
+			line-height: 2;
+		}
+
+		> span:not(.icon) {
+			padding-inline-start: tokens.$spacer-3;
+		}
+	}
+
+	/* stylelint-disable-next-line selector-max-specificity, rule-empty-line-before */
+	&-item.tree-leaf.has-icon > span:not(.icon):hover {
+		text-decoration: underline !important;
+	}
+
+	// Focus: suppress outline on branch nodes, delegate to expander
+
+	&-item:not(.tree-leaf) {
+		outline: none !important;
+
+		@include mixins.focus-visible {
+			> .tree-expander {
+				@include mixins.focus();
+			}
+		}
+	}
+
+	&-item.tree-leaf,
+	&-expander {
+		display: block;
+		padding-inline-start: $tree-indent;
+		padding-block-start: $tree-item-vertical-padding;
+		padding-block-end: $tree-item-vertical-padding;
+		border-radius: tokens.$border-radius;
+		outline-offset: -(tokens.$focus-width) !important;
+	}
+
+	&-expander {
+		position: relative;
+		cursor: pointer;
+		user-select: none;
+		word-wrap: break-word;
+
+		.tree-expander-indicator {
+			display: inline-block;
+			position: absolute;
+			inset-block-start: tokens.$spacer-3;
+			inset-inline-start: 0;
+			transition: rotate $tree-transition-duration $tree-easing;
+			rotate: 0deg;
+			color: tokens.$text-subtle;
+			font-size: tokens.$font-size-9;
+			font-weight: tokens.$weight-semibold;
+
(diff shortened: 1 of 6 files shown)


css/src/components/tree.scss
```diff
@@ -0,0 +1,137 @@
...
+			font-weight: tokens.$weight-semibold;
+
+			&:dir(rtl) {
+				rotate: 180deg;
+			}
+		}
+	}
+
+	&-item.is-expanded > .tree-expander > .tree-expander-indicator {
+		rotate: $tree-expanded-rotation; // TODO: compare expanded chevron between this and prod - see if docs-ui is influencing the styling
```
sahiljohari: the chevron in expanded state seems to be very slightly misaligned on the Atlas docs example but not on the real page when testing with docs-ui locally. Need to look further into this.

sahiljohari: using the existing mixin for chevron fixed it.

bbalentine87:
@sahiljohari Are chevrons a darker color in this PR than the existing design? Should it be subtle text color?

sahiljohari:
> @sahiljohari Are chevrons a darker color in this PR than the existing design?

@bbalentine87 yes, they were using a font-size of 0.55rem earlier which was an odd value. This PR changes that to get it to a standard font-size.

bbalentine87:
The chevron does not seems to be rotating correctly in RTL:
image

bbalentine87:
image
I like that we have reduced the text size on the selected state. I'm still wondering about the left padding on all of the items in the tree. It looks a little cramped on the left when the item is selected. Feel free to resolve this comment though!

sahiljohari:
> image I like that we have reduced the text size on the selected state. I'm still wondering about the left padding on all of the items in the tree. It looks a little cramped on the left when the item is selected. Feel free to resolve this comment though!

@bbalentine87 I think that's how it currently is for leaf items on trees where the collapsible data attribute is set to `false`. Happy to add some left padding but wondering if we should get designer inputs on this?

bbalentine87:
> > image I like that we have reduced the text size on the selected state. I'm still wondering about the left padding on all of the items in the tree. It looks a little cramped on the left when the item is selected. Feel free to resolve this comment though!
> 
> @bbalentine87 I think that's how it currently is for leaf items on trees where the collapsible data attribute is set to `false`. Happy to add some left padding but wondering if we should get designer inputs on this?

Getting design input makes sense to me, but I'm also ok with just resolving this comment if you and/or others feel differently.

css/src/components/tree.scss
```diff
@@ -0,0 +1,135 @@
...
+		&:dir(rtl)::before {
+			@include mixins.chevron-right-rtl;
+		}
+	}
+
+	&-item.is-expanded > .tree-expander::before {
+		transform: mixins.$chevron-down-rotate translate(calc(mixins.$chevron-arrow-size / 2), 0);
+	}
+
+	&.tree-static {
```
wibjorn: seeing some strange current styling on this. I don't think we want the removal of the inline-start padding. I'm also a little curious if we need  this modifier at all. Thoughts on removal?

Image

css/src/components/tree.scss
```diff
@@ -0,0 +1,135 @@
...
+			transition: transform $tree-transition-duration $tree-easing;
+			font-size: tokens.$font-size-9;
+		}
+
+		&:dir(rtl)::before {
+			@include mixins.chevron-right-rtl;
+		}
+	}
+
+	&-item.is-expanded > .tree-expander::before {
```
wibjorn: stylistic nit, but this &- just makes things hard to read. Thoughts on just writing .tree-item?

site/src/components/tree.md
```diff
@@ -0,0 +1,83 @@
...
+| `.tree-static` | modifier | Non-collapsible tree variant (no expanders) |
+| `.tree-group` | `ul` | Nested group of child items |
+| `.tree-item` | `li` or `a` | A tree node (branch or leaf) |
+| `.tree-leaf` | `a` | Leaf node (no children) |
+| `.tree-expander` | `span` | Clickable expander for branch nodes (chevron drawn via CSS `::before`) |
+| `.is-expanded` | modifier | Applied to expanded branch nodes |
+| `.is-selected` | modifier | Applied to the currently active item |
+
+## Accessibility
+
```
wibjorn, line 66: The `.tree` component is intended to style an element that matches the AAI-ARIA recommended tree example: https://www.w3.org/WAI/ARIA/apg/patterns/treeview/

site/src/components/tree.md
```diff
@@ -0,0 +1,83 @@
...
+	<li role="none">
+		<a class="tree-item tree-leaf" role="treeitem" tabindex="-1" aria-level="1" aria-setsize="4" aria-posinset="3" href="#">Steps</a>
+	</li>
+	<li role="none">
+		<a class="tree-item tree-leaf" role="treeitem" tabindex="-1" aria-level="1" aria-setsize="4" aria-posinset="4" href="#">Next steps</a>
+	</li>
+</ul>
+```
+
+## DOM contract
```
wibjorn: I don't know if we need to be writing things out to this level of detail. Perhaps we can restructure the documentation. Examples and code samples ought to be the main focus.

site/src/components/tree.md
```diff
@@ -0,0 +1,83 @@
...
+description: The tree component in the Atlas Design System
+template: standard
+classType: Component
+classPrefixes:
+  - tree
+---
+
+# Tree
+
+A hierarchical list component implementing the [WAI-ARIA TreeView pattern](https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html). Used for navigable structures like tables of contents, file browsers, or nested menus.
```
wibjorn: Technically, it does not implement the pattern listed. It is intended to _style_ the pattern listed there.

css/src/components/tree.scss
sahiljohari: makes sense. All I did here is migrate as is, but I see how it's not really providing much value if all it does is set padding to 0.

css/src/components/tree.scss
sahiljohari: sure! I used it over the entire file but happy to remove this pattern entirely if it improves readability.

sahiljohari:
> @sahiljohari Are chevrons a darker color in this PR than the existing design? Should it be subtle text color?

@bbalentine87 yes, I matched it to use what the Accordion component uses, but happy to revert to using the subtle text color.

bbalentine87 (dismissed):

css/src/components/tree.scss
lmackenzie8: Completely agree that the removal of the padding looks strange

lmackenzie8 (dismissed):

css/src/components/tree.scss
```diff
@@ -0,0 +1,137 @@
...
+$tree-selected-font-weight: tokens.$weight-semibold !default;
+
+.tree {
+	position: relative;
+	font-size: $tree-font-size;
+	line-height: 1.75;
+
+	&,
+	&-group {
+		list-style-type: none;
```
sahiljohari: Can probably remove this

lmackenzie8 (dismissed):

css/src/components/tree.scss
```diff
@@ -0,0 +1,135 @@
...
+	.tree-expander {
+		display: block;
+		padding-inline-start: $tree-indent;
+		padding-block-start: $tree-item-vertical-padding;
+		padding-block-end: $tree-item-vertical-padding;
+		border-radius: tokens.$border-radius;
+		outline-offset: -(tokens.$focus-width) !important;
+	}
+
+	.tree-item.tree-leaf:where(.has-icon) {
```
wibjorn: .has-icon does not match the atlas-naming conventions, also could it be `:has(.icon)`?

css/src/components/tree.scss
```diff
@@ -0,0 +1,135 @@
...
+			border-color: tokens.$text-subtle;
+			font-size: tokens.$font-size-9;
+		}
+
+		&:dir(rtl)::before {
+			@include mixins.chevron-right-rtl;
+		}
+	}
+
+	.tree-item:where(.is-expanded) > .tree-expander {
```
wibjorn: I notice that in documentation it says:

>Branch nodes use `aria-expanded` to communicate state.

which I think is a good idea; but should the styles also use aria-expanded?

wibjorn: (feel free to resolve)

site/src/components/tree.md
```diff
@@ -0,0 +1,67 @@
...
+  - tree
+---
+
+# Tree
+
+A hierarchical list component that styles the [WAI-ARIA TreeView pattern](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/). Used for navigable structures like tables of contents, file browsers, or nested menus.
+
+## Basic collapsible tree
+
+A tree with expandable branch nodes and selectable leaf nodes. Branch nodes use `aria-expanded` to communicate state.
```
wibjorn, line 16: As mentioned in another file, this does not appear to be correct, but it's a good idea

site/src/components/tree.md
```diff
@@ -0,0 +1,67 @@
...
+	</li>
+	<li role="none">
+		<a class="tree-item tree-leaf" role="treeitem" tabindex="-1" aria-level="1" aria-setsize="3" aria-posinset="3" href="#">FAQ</a>
+	</li>
+</ul>
+```
+
+## Static tree
+
+A tree without expanders — all items are leaf nodes. No additional class is needed; simply omit `.tree-expander` elements.
```
wibjorn: Nice!

site/src/components/tree.md
```diff
@@ -0,0 +1,67 @@
...
+	</li>
+	<li role="none">
+		<a class="tree-item tree-leaf" role="treeitem" tabindex="-1" aria-level="1" aria-setsize="4" aria-posinset="4" href="#">Next steps</a>
+	</li>
+</ul>
+```
+
+## Accessibility
+
+Keyboard navigation (arrow keys, Home, End) and focus management should be handled by the consuming application's JavaScript.
```
wibjorn: Depending on which way you go with the aria-attributes, I think may be worth mentioning them here.

css/src/components/tree.scss
sahiljohari: makes sense. so we get rid of `.is-expanded` class entirely?

wibjorn: I think that's not a bad idea, I like baking accessible markup into the CSS if possible.

wibjorn (dismissed):

lmackenzie8 (dismissed):

bbalentine87 (approved):

lmackenzie8 (approved):

PR #863, opened 2026-05-18
Add JavaScript functionality for restoring layouts

wibjorn:
changes.

I've been trying to create some code that helps us restore the layout related html classes on the documentElement, then create an API subscribers can listen for changes and also use the same api just to make changes they want to happen on page load. It's been hard to fully figure out, and I've been bouncing between ideas to get the right balance.
 
I landed on the following:

We put an IIEF directly in the markup of pages on the consumer site. This is a short snippet that essentially checks local storage and uses it to restore the same, if we're on the same page template. This happens quickly, synchronously, before we get into any bundle code. This is a common method for restoring color themes without flashes of different themes too. It just lives in documentation, not source.

Next, (we are in the bundles now), we read the state again (which is keyed on page templates / view names), and create a thing (`layoutState`) that can both watch for changes to the relevant html classes on the documentElement _and_ save the changes to state.

Subscribers write functions that run when a particular state is true - and they also run right away once if the state matches their conditions.

Put in another set of css helper classes, which are design to hide things if the layout is still restoring.

What this'll do ensure that we never see any jank on load, and give us the flexibility we need to easily write functions that target specific states.

Additionally, there is a bug with parcel that breaks the initial uncached build when a plain script tag is added to HTML. There is a workaround in place in a new workspace in /plugins/parcel-optimizer-inline-script-restore which circumvents the logic to fix the bug.

## Testing

1. Visit layout page.
2. Select a relevant expand/collapse toggle, such as toggling the menu, aside, or the flyout.
3. Refresh page. You should see the same state as you left off.
4. Note that the Atlas doc site does not support expanding/collapsing of elements off the layout page.

## Contributor checklist

- [x] Does your pull request implement complex UI logic (js/ts)? Consider adding an integration test to test your user flow.

### DIFF
--- a/site/src/components/layout.md
+++ b/site/src/components/layout.md
@@ -18,7 +18,7 @@ The layout component provides a flexible and efficient way to structure the majo
 This page is utilizing the holy grail layout, but you can use the buttons below to toggle layouts and test them out by resizing the screen. Try the "Toggle container labels" button below to see the css classes on each of the containers inside `.layout`.
 
 <div class="buttons buttons-addons margin-top-sm">
-  <button class="button" data-set-layout="layout-holy-grail" aria-pressed="true">Holy grail</button>
+  <button class="button" data-set-layout="layout-holy-grail">Holy grail</button>
   <button class="button" data-set-layout="layout-twin">Twin</button>
   <button class="button" data-set-layout="layout-single">Single</button>
   <button class="button" data-set-layout="layout-sidecar-left">Sidecar left</button>
--- a/site/src/components/layout.md
+++ b/site/src/components/layout.md
@@ -6,6 +6,7 @@ classType: Component
 classPrefixes:
   - layout
 hero: true
+layoutViewName: atlas-layout-page
 ---
 
 # Layout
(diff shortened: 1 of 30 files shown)


js/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@microsoft/atlas-js",
-	"version": "1.15.1",
+	"version": "1.15.1-test.2",
```
wibjorn: Not to self, fix before continuing with reviews.

site/src/components/layout.md
```diff
@@ -461,3 +482,50 @@ document.documentElement.classList.add('layout-holy-grail');
...
+`@microsoft/atlas-js` exports `createLayoutState`, which persists `layout-*` classes on `<html>` to `localStorage` and restores them the next time the page loads. See the [atlas-js README](https://github.com/microsoft/atlas-design/tree/main/js#readme) for the API. Skip this section if your page layout never changes after load.
+
+Calling `createLayoutState()` from a module bundle restores classes, but **not before first paint** — module scripts run after HTML parse, so the browser briefly renders the original markup layout, causing a visible reflow.
+
+### Fix: inline restore script in `<head>`
+
+Place this synchronous IIFE in `<head>` before the bundle. It reads the same `localStorage` key that `createLayoutState` uses and applies the saved classes to `<html>` during parse:
+
+```html-no-example
+<head>
```
wibjorn, line 525: Here is the relevant snippet.

lmackenzie8 (approved):

bbalentine87 (approved):

PR #864, opened 2026-05-20
Override serialize-javascript to 7.0.5

davidle-ms:
Overriding serialize-javascript to version 7.0.5 to address CVE


davidle-ms:
@microsoft-github-policy-service agree company="Microsoft"

package-lock.json
```diff
@@ -62,8 +62,6 @@
 		},
 		"css/node_modules/fs-extra": {
 			"version": "11.3.5",
```
wibjorn, line 64: @davidle-ms , thanks for contribution. The package.json file is fine, but the package-lock seems to have an unintentionally removed the integrity hashes from the deps. Please try again - you can try `npm cache clear` before you run it again, that may help.

davidle-ms: Hi @wibjorn, this is now rectified. Let me know if there's anything else I need to change.

wibjorn (approved):

bbalentine87 (approved):

PR #868, opened 2026-05-28
fix: update tsconfig.json with required compiler options

Courtny:
## Description

Fixes TypeScript compilation failures in the `@microsoft/atlas-js` package build process.

### Problem
The build was failing during package publication with TypeScript compiler errors. The `tsconfig.json` was missing essential compiler options required for emitting declaration files and outputting to the dist directory.

### Solution
Updated `js/tsconfig.json` with:
- `lib` option for type definitions
- `rootDir` and `outDir` for proper file structure
- `declaration` and `declarationMap` in config (moved from CLI)
- `esModuleInterop`, `skipLibCheck`, and `forceConsistentCasingInFileNames`
- `include`/`exclude` patterns for clarity

### Fixes
Resolves the release workflow failure preventing `@microsoft/atlas-js@1.15.2` from publishing to npm.


Courtny:
@microsoft-github-policy-service agree

wibjorn:
Hey @Courtny! I think this change was likely done very recently, would you mind looking at latest? Also let me know if there's a use case we should know about.

wibjorn (changes requested):
Not for merging now

PR #902, opened 2026-07-24
add table-cell-max-width class

osolomakha:

Some conceptual content(https://learn.microsoft.com/en-us/azure/templates/microsoft.sql/allversions) includes long resource-like strings that can make table columns overly wide.
This change provides a scoped, explicit solution instead of changing all tables globally.

## Testing

1. Npm run start
2. Go [Tables local host](http://localhost:1111/components/table)
3. Check the screenshot, the part "Cell max width"
   Expected result:
Screenshot (640)

### DIFF
--- a/css/src/components/table.scss
+++ b/css/src/components/table.scss
@@ -101,6 +101,12 @@ $table-caption-spacing: tokens.$letter-spacing-medium !default;
 		}
 	}
 
+	&.table-cell-max-width {
+		td {
+			max-width: 40rem;
+		}
+	}
+
 	&.table-sm {
 		th,
 		td {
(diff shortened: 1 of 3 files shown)


css/src/components/table.scss
```diff
@@ -101,6 +101,12 @@ $table-caption-spacing: tokens.$letter-spacing-medium !default;
 		}
 	}
 
+	&.table-cell-max-width {
+		td {
+			max-width: 40rem;
```
ollips: This doesn't work for all possible screen sizes, tables end up going off screen

Image

ollips (changes requested):

css/src/components/table.scss
ollips: Here is a better view, table gets beyond the main content area (highlighted in blue)

image

PR #910, opened 2026-08-14
Improve safety of form-behavior navigation patterns

sahiljohari:
## Summary

- validate form-submit navigation URLs before passing them to browser navigation sinks
- allow relative and absolute HTTP(S) URLs while rejecting executable or unsupported schemes
- preserve follow, replace, hash-reload, and reload behavior for valid inputs
- add regression coverage for obfuscated javascript URLs and other unsafe schemes
- include a patch changeset for @microsoft/atlas-js

## Validation

- npm test
- npm run lint
- npm run build
- Prettier check

### DIFF
--- a/.changeset/stale-baboons-tickle.md
+++ b/.changeset/stale-baboons-tickle.md
@@ -2,4 +2,4 @@
 '@microsoft/atlas-js': patch
 ---
 
-Validate the url scheme before navigating after a form submit. `navigateAfterSubmit` previously passed the `Location` response header or the `navigation-href` attribute straight to `location.href` and `location.replace()`, which allowed a `javascript:` url to execute script in the page's origin. Those hrefs are now resolved with `new URL()` and only `http:` and `https:` are followed; anything else is refused and logged. Relative and absolute `http(s)` urls, including cross-origin ones, navigate exactly as before.
+Validate the url scheme before navigating after a form submit. `navigateAfterSubmit` previously passed the `Location` response header or the `navigation-href` attribute straight to `location.href` and `location.replace()`, which allowed a `javascript:` url to execute script in the page's origin. Those hrefs are now resolved with `new URL()` and only `https:` urls are followed; anything else is refused and logged. Relative urls continue to work when the page is served over HTTPS.
--- a/.changeset/stale-baboons-tickle.md
+++ b/.changeset/stale-baboons-tickle.md
@@ -1,5 +1,7 @@
 ---
-'@microsoft/atlas-js': patch
+'@microsoft/atlas-js': major
 ---
 
-Validate the url scheme and origin before navigating after a form submit. `navigateAfterSubmit` previously passed the `Location` response header or the `navigation-href` attribute straight to `location.href` and `location.replace()`, which allowed a `javascript:` url to execute script in the page's origin. Those hrefs are now resolved with `new URL()` and only same-origin `https:` urls are followed; anything else is refused and logged. Relative urls continue to work when the page is served over HTTPS.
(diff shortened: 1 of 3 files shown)


js/src/elements/form-behavior.ts
```diff
@@ -781,31 +781,62 @@ function canValidate(
 	return isValueElement(target, form) && (target as HTMLValueElement).type !== 'hidden';
 }
 
+const safeNavigationProtocols = ['http:', 'https:'];
```
wibjorn:
```suggestion
const safeNavigationProtocols = ['https:'];
```

js/src/elements/form-behavior.ts
```diff
@@ -781,31 +781,62 @@ function canValidate(
...
+	}
+
+	const trimmed = href.trim();
+	if (!trimmed) {
+		return null;
+	}
+
+	try {
+		const url = new URL(trimmed, window.location.href);
+		return safeNavigationProtocols.includes(url.protocol) ? trimmed : null;
```
wibjorn: `startsWith` - https://url.com/?http:
```suggestion
		return safeNavigationProtocols.startsWith(url.protocol) ? trimmed : null;
```

js/src/elements/form-behavior.ts
```diff
@@ -781,31 +781,62 @@ function canValidate(
 	return isValueElement(target, form) && (target as HTMLValueElement).type !== 'hidden';
 }
 
+const safeNavigationProtocols = ['http:', 'https:'];
+
+function getSafeNavigationHref(href: string | null): string | null {
```
wibjorn: Naming suggestion:
`hardenNavigationHref`

js/src/elements/form-behavior.ts
wibjorn: Doesn't really need to be an array

js/src/elements/form-behavior.ts
```diff
@@ -781,31 +781,62 @@ function canValidate(
...
+		return null;
+	}
+
+	const trimmed = href.trim();
+	if (!trimmed) {
+		return null;
+	}
+
+	try {
+		const url = new URL(trimmed, window.location.href);
```
wibjorn, line 795: Does this technically disallow any requests not to location.origin? Is it expected behavior? Are there downstream cases where this would break behavior? Hopefully not but seems possible.

js/src/elements/form-behavior.ts
```diff
@@ -781,31 +781,62 @@ function canValidate(
...
+		return safeNavigationProtocols.includes(url.protocol) ? trimmed : null;
+	} catch {
+		return null;
+	}
+}
+
 export function navigateAfterSubmit(href: string | null, navigationStep: NavigationSteps) {
+	// Never hand an unvalidated url to a navigation sink; `javascript:` urls execute script.
+	const safeHref = getSafeNavigationHref(href);
```
wibjorn: would suggest renaming the function param to requestedHref and then putting this back to `href`. Fewer diffs, and I don't love labelling things as `safe`.

js/src/elements/form-behavior.ts
sahiljohari: ah good catch! I believe not - let me look into this though.

sahiljohari: added an additional check for URL origin

.changeset/stale-baboons-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@microsoft/atlas-js': patch
```
wibjorn: I think this is technically a breaking change. We just need to change this and also indicate what the breaking change is: navigation to non-origin URLs is no longer supported. http and other url schemas are no longer supported.

wibjorn (approved):

bbalentine87 (approved):

PR #919, opened 2026-09-03
Improve adaptive grid interactive example

wibjorn:
## Summary

- Let the interactive workspace break out of the page reading width while keeping its controls aligned with normal prose.
- Separate the visual workspace from the width-query container so slider values match the queried content width.
- Clamp the slider to the workspace width as the viewport changes.
- Represent the component's default gap with the equivalent `.gap-xs` atomic.
- Simplify the example and add narrow-screen guidance.

## Testing

- `npm run build:site`
- `npm run lint`

### DIFF
--- a/site/src/components/adaptive-grid.md
+++ b/site/src/components/adaptive-grid.md
@@ -102,9 +102,12 @@ two-column class and threshold pairs with `$adaptive-grid-columns-2-variants`.
 ## Interactive container size
 
 Use the button groups to apply a column modifier and gap atomics, then use the slider to change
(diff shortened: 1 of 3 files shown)


sahiljohari (approved):

ollips (approved):

site/src/components/adaptive-grid.md
```diff
@@ -102,9 +102,12 @@ two-column class and threshold pairs with `$adaptive-grid-columns-2-variants`.
 ## Interactive container size
```
ollips, line 102: changeset needed?

